PR #155, opened 2017-01-10
Use absolute urls in README so images and links work on NPM

crookedneighbor:
Noticed that the image link was broken on NPM.

screen shot 2017-01-10 at 2 45 39 pm

The links were also broken, since they were pointing to relative pages on the npm page instead of to Github.


bluepnume:
Thanks for the catch! Merging

PR #184, opened 2017-02-09
Add files param to package.json to reduce file size of npm package

crookedneighbor:
On master, this module currently package up 169 files amounting to 4.4 MB. By specifying only the `dist` directory in the package.json, we can eliminate the test and example app code from `npm` and bring it down to 10 files amounting to 1.2 MB.

I wrote a small module to assist with showing what is getting packaged up called `npm-unpack`.

```bash
$ npm i -g npm-unpack
$ npm-unpack
Packing paypal-checkout@4.0.39...

Number of files: 10

package.json
README.md
index.js
LICENSE.txt
dist/checkout.js
dist/checkout.lib.js
dist/checkout.min.js
dist/checkout.js.map
dist/checkout.lib.js.map
dist/checkout.min.js.map
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,11 @@
     "test": "gulp test",
     "flow": "flow; test $? -eq 0 -o $? -eq 2"
   },
+  "files": [
+    "dist/checkout.lib.js",
+    "dist/checkout.lib.js.map",
+    "src/"
+  ],
   "repository": {
     "type": "git",
     "url": "git://github.paypal.com/paypal/paypal-checkout.git"


bluepnume:
I'd actually like to keep src/. If another module requiring paypal-checkout uses ES6 with webpack, they can just require us from `src/` directly and the resulting bundle stays deduped. We're using this pattern internally for our dependency graph between paypal-checkout, xcomponent and post-robot.

So I'm thinking we can keep:

```dist/checkout.lib.js
src/**```

We can dump the other stuff from dist/ since that's mainly geared towards paypalobjects -- `checkout.lib.js` is the UMD module designed for consumption through npm.

That work?

crookedneighbor:
Makes sense to me. That brought down the package size to `468K`.

bluepnume:
Perfect. Merging this once the build passes.

PR #195, opened 2017-02-16
Fixing repo in `git remote add upstream` call

ErikMHummel:
The repo still referenced `krakenjs` org instead of the `paypal` org.


bluepnume:
Looks good, thanks for the fix!

PR #208, opened 2017-03-01
prop validations for credit button

vishakha94:
- Error out if the user passes a custom color with the credit button
- Error out when the merchant passes ‘tiny’ in the size prop for credit

### DIFF
--- a/src/components/common.js
+++ b/src/components/common.js
@@ -26,6 +26,22 @@ export function validateProps(props : Object, required : boolean = true) {
         }
     }
 
+    // Prop Validations for the credit button
+
+    // Tiny size for credit button is not supported
+    if (props.style.label && props.style.size) {
+        if (props.style.label === 'credit' && props.style.size === 'tiny') {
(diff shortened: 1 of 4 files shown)


src/components/common.js
```diff
@@ -26,6 +26,22 @@ export function validateProps(props : Object, required : boolean = true) {
         }
     }
 
+    // Prop Validations for the credit button
+
+    // Tiny size for credit button is not supported
+    if (props.style.label && props.style.size) {
```
bluepnume: Can we make sure there's a null check for `props.style`?

We can probably make this whole line:

`if (props.style && props.style.label === 'credit' && props.style.size === 'tiny') {`

vishakha94: ok

bluepnume:
I reverted this for now, since the build was broken.

Also would you mind throwing a couple of tests in [https://github.com/paypal/paypal-checkout/blob/master/test/tests/button/validation.js](https://github.com/paypal/paypal-checkout/blob/master/test/tests/button/validation.js)? Thanks!

PR #223, opened 2017-03-06
Credit button validations

vishakha94:


bluepnume:
Tests look great :) Were you also going to add a check / test to see if `style.label` is one of `[ 'checkout', 'credit' ]`?

vishakha94:
sure will do!

bluepnume:
Looks like the CI job is failing -- did `gulp test` pass for you, locally?

src/components/common.js
```diff
@@ -27,6 +27,22 @@ export function validateProps(props : Object, required : boolean = true) {
         }
     }
 
+    if (props.style && props.style.label) {
+        if (props.style.label !== 'checkout' || props.style.label !== 'credit') {
```
bluepnume: I think this check will always be true? Guessing it should be `&&` rather than `||`.

vishakha94: Bummer

bluepnume:
Looks great! Merging

PR #229, opened 2017-03-07
(setup): Fix setup function from being called twice

trainerbill:
Setup function is being called more than once because currentScript is always defined and there are muliple exports of setup.js. not sure if this works or if it passes tests because I wasn't able to build the project.  Probably need to take a look at the contributing documentation and make sure its accurate.

image

image

### DIFF
--- a/src/setup.js
+++ b/src/setup.js
@@ -133,29 +133,35 @@ export function setup({ env, stage, apiStage, paypalUrl, state, ppobjects, light
     $logger.debug(`current_protocol_${currentProtocol}`);
 }
 
-if (currentScript) {
+let setupComplete = false;
+if (!setupComplete) {
+    
+    if (currentScript) {
 
-    setup({
-        env:       currentScript.getAttribute('data-env'),
-        stage:     currentScript.getAttribute('data-stage'),
-        apiStage:  currentScript.getAttribute('data-api-stage'),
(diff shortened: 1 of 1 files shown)


bluepnume:
This is happening because setup is being called in the button iframe too. So I don't think this fix is going to work.

On a side note, let me add some quick-start steps in the readme.

bluepnume:
FYI: https://github.com/paypal/paypal-checkout/blob/master/README.md#development

trainerbill:
@bluepnume So is checkout.js required on the iframe page as well, or is it there just to have the same version of xcomponent?  Couldn't you just have xcomponent?

Is setup required to be called on the iframe page?  If not can we make a param to does not call it if set?

If its required then can we get a logPrefix param?

```
logPrefix:  currentScript.getAttribute('data-log-prefix'),
```

Then on Hermes do "data-log-prefix="ppxo_hermes"

bluepnume:
> @bluepnume So is checkout.js required on the iframe page as well, or is it there just to have the same version of xcomponent? Couldn't you just have xcomponent?

No, the way xcomponent works, the component definition needs to live in both the parent and the child. This is also necessary so we can do things like rendering `paypal.Checkout` up from `paypal.Button`'s iframe.

> If its required then can we get a logPrefix param?

Yeah, that's a good idea. 👍

PR #230, opened 2017-03-07
(feat): Log prefix parameter on script tag

trainerbill:
I commented out one log line in getCurrentScript since $logger wouldn't be initialized yet... I can change it to console.log if you want.

There are probably better ways to accomplish this but this is the least intrusive. 

Other than that it seems to be working, but the build process is still not working quite right so I am not 100% sure.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -32,7 +32,8 @@
         "__MINOR_VERSION__": true,
         "__MAJOR_VERSION__": true,
         "__FILE_NAME__": true,
-        "__FILE_VERSION__": true
+        "__FILE_VERSION__": true,
+        "HTMLElement": true
     },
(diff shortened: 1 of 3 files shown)


bluepnume:
I like the idea of changing the prefix per-component, but I'm not sure about adding it as a script prop that anyone can change. Let me mull this over, see if there's a way to do this internally.

trainerbill:
Originally I wanted it because the logs are out of control.  After I had it done I realized that all logs with host: "localhost(orwhatever)" were generated from my app and anything from Hermes has sandbox.paypal.com...  So its probably not needed.  was fun anyway

bluepnume:
OK. Cool.

PR #271, opened 2017-03-17
Remove keys with undefined values from style prop

### LINKED ISSUE #253: feat(button): Buttons need to be able to work with undefined values
Now that the credit is available and we cant use color with credit we need to be able to support undefined values in the style properties.

```
var buttonConfig = [
            {
                id: '#paypal-button1',
                size: 'tiny',
                color: 'silver',
                shape: 'pill'
            },
            {
                id: '#paypal-button2',
                size: 'medium',
                label: 'credit'
            }
        ];
        
        buttonConfig.forEach(function (button) {
            paypal.Button.render({
                env: 'sandbox',
                client: {
                    sandbox:    'AQzwHJt9jZAyQbf-DnaqZefvbTCMammV7ZFsnW23thr9qUQX9RGMgSVbzRjxXTAODKjh2W0k6Ei4Q2m9'
                },
                onError: onError,
                onCancel: onCancel,
                payment: createPayment,
                onAuthorize: executePayment,
                commit: true,
                style: {
                    
                    size: button.size,
                    shape: button.shape,
                    label: button.label,
                    color: button.color
                }
            }, button.id);
        });
```

trainerbill:
Fixes #253

### DIFF
--- a/src/components/button/component.js
+++ b/src/components/button/component.js
@@ -267,6 +267,14 @@ export let Button = xcomponent.create({
 
             validate(style) {
 
+                Object
+                    .keys(style)
+                    .forEach(key => {
+                        if (!style[key] || typeof style[key] !== 'string') {
+                            delete style[key];
+                        }
+                    });
+
                 if (style.size && config.buttonStyles.size.indexOf(style.size) === -1) {
                     throw new Error(`Invalid button size: ${style.size}`);
                 }


bluepnume:
Added https://github.com/krakenjs/xcomponent/commit/4625e172e8022206ff36b4505604f38b6a0c4bcc to avoid mutating props in `validate()`. Published the fix. This fixes the issue in my localhost -- please let me know if you still see it.

trainerbill:
Nice.  That was actually my original fix, but I kept getting the popup undefined and thought it was a regression error that I couldn't figure out... so I went with this and got the same issue.

PR #277, opened 2017-03-22
IE Compatible Header Detection

vishakha94:


src/lib/errors.js
```diff
@@ -38,6 +38,10 @@ export function checkForCommonErrors() {
         $logger.warn(`ie_intranet_mode`);
     }
 
+    if (isIE() && !isIECompHeader()) {
+        $logger.warn(`Drop tag: <meta http-equiv="X-UA-Compatible" content="IE=edge">`);
```
bluepnume: Can we make this an alphanumeric event, since it ends up in our logging back-end?

Maybe `$logger.warn(`ie_meta_compatibility_header_missing`, { message: `Drop tag: <meta http-equiv="X-UA-Compatible" content="IE=edge">` });`?

bluepnume:
Looks good! Just one minor comment

src/lib/errors.js
vishakha94: Alphanumeric event?

bluepnume: Yeah, the first string you pass to `$logger` should not have any spaces or special characters, because it gets converted to a CAL event on our server side, and used to build reports of event counts.

szarouski:
Hi, we log warnings/errors from our site & `ie_meta_compatibility_header_missing` is one of the errors we started to see some time around end of July. We support IE11 & up. How can we get rid of the error? Thank you.

bluepnume:
Do you have the `<meta http-equiv="X-UA-Compatible" content="IE=edge">` header on your page?

szarouski:
Is it required for IE11 and up? https://stackoverflow.com/questions/6771258/what-does-meta-http-equiv-x-ua-compatible-content-ie-edge-do

bluepnume:
We've seen plenty of strange behavior from leaving this out. So it definitely doesn't hurt. I think there's still a chance IE11 can fall into quirks mode.

PR #279, opened 2017-03-22
Fix priorities of languages for each country 

vishakha94:


bluepnume:
Looks like tests failed :(

vishakha94:
Apparently the indentation. Will be careful next time.

PR #293, opened 2017-04-12
content and validations for pay with button

vishakha94:


src/components/button/templates/component/template.js
```diff
@@ -1,11 +1,11 @@
 /* @flow */
 
-import { config } from '../../../../config';
+// import { config } from '../../../../config';
 
 import { componentLogos } from './logos';
 import { componentStyle } from './style';
 import { componentScript } from './script';
-import componentContent from './content.json';
+import componentContent from './buttonContent.json';
```
bluepnume: Can we keep it as `content.json`?

bluepnume:
Looks perfect :) just need to delete buttonContent.json and I'll merge

PR #309, opened 2017-04-27
Simplified Braintree Integration

bluepnume:
Allows integrating Braintree with the following pattern:

```javascript
paypal.Button.render({
    
    env: 'sandbox',
    
    braintree: braintree,
    
    client: {
        sandbox: 'SANDBOX_AUTH',
        production: 'PRODUCTION_AUTH'
    },
    
    payment: function(actions) {
        return actions.braintree.create({
            flow:     'checkout',
            amount:   '1.00',
            currency: 'USD',
            intent:   'sale'
        });
    },

    onAuthorize: function(data, actions) {
        console.log('Nonce:', data.nonce);
    }

}, '#paypal-button-container');
```

The objective of this is:

- Less boilerplate
- Easier to upgrade from a REST integration to Braintree
- Normalizes the integration patterns even further
- Parallelizes the Braintree client creation and the paypal.Button rendering

### DIFF
--- a/src/components/button/braintree.js
+++ b/src/components/button/braintree.js
@@ -1,0 +1,23 @@
+/* @flow */
+
+import { SyncPromise } from 'sync-browser-mocks/src/promise';
+
+export type Braintree = {
+    client : {
+        create : Function
+    },
+    paypalCheckout : {
+        create : Function
+    }
+};
+
+export type BraintreePayPalClient = {
+    createPayment : Function,
+    tokenizePayment : Function
+};
+
+export function awaitBraintreeClient(braintree : Braintree, authorization : string) : SyncPromise<BraintreePayPalClient> {
+    return braintree.client.create({ authorization }).then(client => {
+        return braintree.paypalCheckout.create({ client });
+    });
+}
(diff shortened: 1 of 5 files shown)


src/components/button/braintree.js
```diff
@@ -0,0 +1,34 @@
...
+
+export type BraintreePayPalClient = {
+    createPayment : Function,
+    tokenizePayment : Function
+};
+
+export function awaitBraintreeClient(braintree : Braintree, authorization : string) : SyncPromise<BraintreePayPalClient> {
+    return new SyncPromise((resolve, reject) => {
+
+        braintree.client.create({ authorization }, (err, client) => {
```
crookedneighbor: Since v3.9.0, `braintree.client.create` and `braintree.paypalCheckout.create` will return a promise if you don't pass a callback, so this could just be:

```js
return braintree.client.create({ authorization })
  .then(client => braintree.paypalCheckout.create({ client }));
```

bluepnume: Oh perfect!

bluepnume: OK changed to use promises instead. Going to merge this soon if no further thoughts. I've added some demo docs to our staging env at http://devdocs-4307.ccg21.dev.paypalcorp.com:8888/docs/integration/direct/express-checkout/integration-jsv4/braintree-sdk/

Issue #315, opened 2017-05-03
Developer's console message buried in PayPal debugging messages.


thirtified:
checkout.js is very chatty and currently floods the console with PayPal debugging messages, making it hard for an integrating developer to spot their own console messages: ![checkout js console output](https://cloud.githubusercontent.com/assets/818580/25667711/9f840710-3024-11e7-8613-0888bd9ed58c.png)

It is already possible to reduce the checkout.js log level and thus suppress most of the PayPal messages by setting an attribute `data-log-level="warn"` on the script tag, but this option is not documented.

Since most of PayPal's log messages seem to be of internal nature and don't seem to be relevant for external developers, I'd suggest to set the default log level to `warn` and thus keep the internal debugging information by default from those not explicitly interested.

bluepnume:
Yeah, ok. I was reluctant to do this originally since we rely so much on merchants being able to send us these logs, to triage issues, but I think we're in a stable enough state at the moment that we can do this.

Would be good to think of an easy way to help people testing enable logs on a case-by-case basis though (maybe some local storage key or something?)

PR #316, opened 2017-05-03
Clean up console output.

thirtified:
Hide internal debugging messages to console by default, see #315.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,12 @@ gulp build
 ```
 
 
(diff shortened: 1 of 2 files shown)


bluepnume:
Yeah, ok. I was reluctant to do this originally since we rely so much on merchants being able to send us these logs, to triage issues, but I think we're in a stable enough state at the moment that we can do this.

Would be good to think of an easy way to help people testing enable logs on a case-by-case basis though (maybe some local storage key or something?)

Thanks for the PR!

PR #320, opened 2017-05-05
Show that response can be obtained after successful promise

XcrossD:


bluepnume:
Looks good -- thanks!

PR #326, opened 2017-05-09
[WIP] David-dm dependencies badges and dependency updates

ErikMHummel:
# WORK IN PROGRESS -- DO NOT MERGE

Adding the David-DM badges.
The release dependencies were green already. This PR also updates devDependencies to green except for these two:

1. flow-runtime is at 0.11.0 as upgrading to 0.12.0 fully breaks. Unsure if it is a bug from the upstream change or a breaking change we need to handle. The change seems like it shouldn't be breaking based on the [diff of the two versions](https://github.com/codemix/flow-runtime/compare/da7ffa7826fba85b767be92c0e8bc735ebaa4a7c...16ce6fee732ce0011d841b0b9d70e8091b510813), so I feel it is a bug but have not yet proven it.

1. mocha stays with 2.5.3 as the 3.x release introduced breaking changes that will require a decent amount of changes to the tests as described below which is copy/paste from the [mocha changelog.md for 3.0.0 release breaking changes section](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#boom-breaking-changes)
> To avoid common bugs, when a test injects a callback function (suggesting asynchronous execution), calls it, and returns a Promise, Mocha will now throw an exception:
```javascript
const assert = require('assert');

it('should complete this test', function (done) {
  return new Promise(function (resolve) {
    assert.ok(true);
    resolve();
  })
    .then(done);
});
```
> The above test will fail with Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both..


ErikMHummel:
Closing this PR and creating a new branch and PR as I somehow have screwed up the commits worse trying to rebase to clean it up. 😝

PR #328, opened 2017-05-09
David-dm dependencies badges and dependency updates

ErikMHummel:
# Adding the David-DM badges.

The release dependencies were green already. This PR also updates devDependencies to green except for `flow-runtime`:

1. `flow-runtime` is only updated to 0.11.` as upgrading to 0.12.0 fully breaks. Unsure if it is a bug from the upstream change or a breaking change we need to handle. The change seems like it shouldn't be breaking based on the [diff of the two versions](https://github.com/codemix/flow-runtime/compare/da7ffa7826fba85b767be92c0e8bc735ebaa4a7c...16ce6fee732ce0011d841b0b9d70e8091b510813), so I feel it is a bug but have not yet proven it.

**Note:** mocha devdependency upgrade had a major and breaking change requiring me to change some testing it broke. [CHANGELOG.md for 3.0.0 release breaking changes section](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#boom-breaking-changes)


bluepnume:
I think something weird is happening with this commit again. I'd recommend doing `git pull --rebase` to potentially avoid some of these issues, going forward.

ErikMHummel:
<grumble> I thought I did...

ErikMHummel:
@bluepnume I just made the changes regarding Mocha in the description to a single test that failed because of the breaking change. Can you review the diff and ensure it is cool to you before I work through all the other tests with this type of failure?
https://github.com/paypal/paypal-checkout/pull/328/files#diff-b9d93cb789e635716d5f99f27e06f9cb

ErikMHummel:
@bluepnume I just made the changes regarding Mocha in the description to a single test that failed because of the breaking change. Can you review the diff and ensure it is cool to you before I work through all the other tests with this type of failure? https://github.com/paypal/paypal-checkout/pull/328/files#diff-b9d93cb789e635716d5f99f27e06f9cb

bluepnume:
Hey @ErikMHummel I think that's not going to work as it removes the actual testing from the test :) Would it be possible for us to leave the `done` and just not return a promise in these tests? (so just remove the `return` statement)?

ErikMHummel:
@bluepnume, I cannot figure out the `flow-runtime` dependency upgrade. If you are cool, could we merge the branch now so it doesn't get out of sync? I have rebased and squashed my changes.

bluepnume:
You got it. Looks good!

PR #338, opened 2017-05-15
buy now button

vishakha94:
Finally we have:
- CSS for buy now button
-Relavent tests for the above
-RegEX to accomodate China C2 
-content.json formatted for buynow -- Replaced buynow_nowordmark with buynow


bluepnume:
Looks good!

PR #343, opened 2017-05-17
branded buy now button

vishakha94:


src/components/button/component.js
```diff
@@ -418,8 +418,9 @@ export let Button = xcomponent.create({
                 if (style.label === 'credit' && style.color) {
                     throw new Error(`Custom colors for ${style.label} button are not supported`);
                 }
-                if (style.label === 'credit' && style.fundingicons) {
-                    throw new Error(`Funding Instruments for ${style.label} button are not supported`);
+
+                if (style.label === 'credit' && style.fundingicons !== undefined) {
```
bluepnume: Any reason we moved from the truthy check here?

src/components/button/component.js
```diff
@@ -429,6 +430,26 @@ export let Button = xcomponent.create({
                 if (style.label === 'buynow' && style.size === 'tiny') {
                     throw new Error(`Invalid ${style.label} button size: ${style.size}`);
                 }
+
+                if (style.label === 'buynow' && style.fundingicons !== undefined && style.branded !== undefined) {
+                    throw new Error(`Must use either of fundingicons or branded option for ${style.label} button`);
+                }
+
+                if (style.label === 'buynow' && style.branded === false) {
```
bluepnume: Could we avoid having four separate conditions here and just do:

```if (style.hasOwnProperty('branded') && style.branded !== true) {
    throw new Error(`Invalid option: style.branded must be unspecified or true`);

Comment (bluepnume) [inline src/components/button/component.js]: Also can we making this `style.branding` rather than `style.branded`

File: src/components/button/templates/component/template.js [hunk 3]
```diff
@@ -108,8 +126,14 @@ export function componentTemplate({ props } : { props : Object }) : string {
...
     let labelTag = '';
 
+    // tag content below button allowed only if fundingicons is false
     if (!props.style.fundingicons) {
-        labelTag = conf.tagline && content[`${label}_tag`] ? content[`${label}_tag`] : '';
+        labelTag = conf.tagline && content[`${conf.tagcontent}`] ? content[`${conf.tagcontent}`] : '';
+    }
+
+    // Hack for buy now branded option to support custom css
+    if (props.style.branded && label === 'buynow') {
```
bluepnume: Could we not introduce a new label here, and instead just have a new class? Like `paypal-button-branded` and `paypal-button-unbranded`?

vishakha94: we do have a new class: 
paypal-style-buynow-branded

src/components/button/component.js
vishakha94: ok

src/components/button/templates/component/template.js
bluepnume: Sure -- I just feel it makes more sense to have an additional class, rather than overloading the `buynow` label class.

bluepnume: It also avoids having extra logic in this render function to overwrite the label.

src/components/button/component.js
vishakha94: shouldnt we throw error if the user is trying to use branding option with anything other than buynow?

src/components/button/component.js
```diff
@@ -430,6 +431,18 @@ export let Button = xcomponent.create({
                 if (style.label === 'buynow' && style.size === 'tiny') {
                     throw new Error(`Invalid ${style.label} button size: ${style.size}`);
                 }
+
+                if (style.label === 'buynow' && style.hasOwnProperty('fundingicons') && style.hasOwnProperty('branding')) {
```
bluepnume: I wanna chat internally about this -- my feeling is we should allow funding icons for any kind of button (except credit)

src/components/button/templates/component/template.js
```diff
@@ -92,8 +96,22 @@ export function componentTemplate({ props } : { props : Object }) : string {
         throw new Error(`Unexpected size for ${label} button: ${size}`);
     }
 
-    let logoColor = conf.logos ? conf.logos[color] : '';
+    let logoColor;
+
+    // button config override for branded buy now button
+    if (props.style.branding && label === 'buynow') {
+        contentText = `\$\{pp\}\$\{paypal\} ${contentText}`;
+        conf.tagline = true;
```
bluepnume: We shouldn't mutate the conf object as this will potentially affect other buttons. Can we use a local variable instead?

src/components/button/templates/component/template.js
```diff
@@ -108,8 +126,15 @@ export function componentTemplate({ props } : { props : Object }) : string {
...
     let labelTag = '';
 
+    // tag content below button allowed only for checkout, branded buynow, credit button;
+    // also when style.fundingicons is false
     if (!props.style.fundingicons) {
-        labelTag = conf.tagline && content[`${label}_tag`] ? content[`${label}_tag`] : '';
+        labelTag = conf.tagline && content[`${conf.tagcontent}`] ? content[`${conf.tagcontent}`] : '';
+    }
+
+    // Hack for buy now branded option to support custom css
```
bluepnume: Not sure if my last comment got lost, but just making a note here -- let's use a different class name for branded / unbranded

src/components/button/component.js
```diff
@@ -419,8 +419,9 @@ export let Button = xcomponent.create({
                 if (style.label === 'credit' && style.color) {
                     throw new Error(`Custom colors for ${style.label} button are not supported`);
                 }
-                if (style.label === 'credit' && style.fundingicons) {
-                    throw new Error(`Funding Instruments for ${style.label} button are not supported`);
+
+                if (style.label === 'credit' && style.hasOwnProperty('fundingicons')) {
+                    throw new Error(`Invalid option: style.fudningicons for ${style.label} button`);
```
bluepnume: style.fundingicons

bluepnume:
Added a few comments -- we should also think for future about how to break as much of these rules into configuration as possible, so we can easily see/change which buttons support which options.

PR #360, opened 2017-06-01
feat(http): Add XHR Timeout configuration

trainerbill:


trainerbill:
@bluepnume Not sure why the button size test is failing...  It did that for me locally as well.  The test doesn't use the http library at all so I am not sure whats up

bluepnume:
Think I found the problem. Let me restart the PR job...

PR #385, opened 2017-06-21
Create NOOP pptm.js script upon setup

brunops:
Create async tag manager script upon setup, works as a NOOP. The idea is that it can be used later on to power new paypal integrations *after* opt-in with no effort.

- Update `src/loader/util#loadScript` so it accepts custom attributes. Used for `id` and `async` when creating `pptm.js` script
- Serve empty `/tagmanager/pptm.js` file through Karma server so there are no 404s when running the tests
- Do not load pptm.js when inside paypal domain. This is necessary so it doesn't end up loaded twice, one in the third-party website, and another inside the iframe
- Add logger calls when creating pptm.js tag and make sure not to include it twice on the same page
- Log when trying to load and after loaded
- Also log if the there was an attempt to create the pptmjs tag twice


karma.conf.js
```diff
@@ -35,6 +35,12 @@ module.exports = function(config) {
             },
 
             {
+                pattern: 'test/lib/pptm.js',
```
bluepnume: Could you just include with a regular `export * from './pptm.js'` to match the other test files?

brunops: This is something else, the idea is to setup karma to serve this empty pptm.js file. So there's no 404 when the tests run, and the script is added to the DOM.

bluepnume: Oh my bad -- sounds good then

bluepnume:
Can't see a single problem -- let me know if you're ready for me to merge, then we can figure out a time this week to do a release!

PR #391, opened 2017-06-26
Buy Now Button bug fix

vishakha94:
Buy Now button wasn't displaying text for some locales since the text was overflowing on the button's content area.
Solution:
Make the size of the unbranded button fit the actual text. So, now size varies according to the text.


bluepnume:
Let's add a max-width: 100% too?

PR #456, opened 2017-09-07
(docs) Add missing parameter to callback functions

santiq:


bluepnume:
Thanks a mil for taking care of this!

PR #458, opened 2017-09-08
Vertical button arrangement (aka paywall)

bluepnume:


src/components/button/templates/style.js
```diff
@@ -0,0 +1,102 @@
+/* @flow */
```
vishakha94: Good job with this!

PR #461, opened 2017-09-12
Update content.json

quadrifolia:
All german translations are missing a word. Added the missing word "Hier".


bluepnume:
Thanks, really appreciate the fix!

PR #519, opened 2017-10-29
[fix] 100% width on iPhone 5 portrait

khanghoang:
WIP #515 

iphone 6 ios 10 3 14e8301 2017-10-29 16-02-41
iphone 5 ios 10 3 14e8301 2017-10-29 15-57-41

### DIFF
--- a/src/components/checkout/templates/container/style.js
+++ b/src/components/checkout/templates/container/style.js
@@ -376,7 +376,6 @@ export function getContainerStyle({ id, tag, CONTEXT, CLASS, ANIMATION } : { id 
         #${ id }.${ tag }-context-${ CONTEXT.IFRAME } .${ CLASS.OUTLET } iframe {
             width: 1px;
             min-width: 100%;
-            *width: 100%;
             height: 100%;
         }
 
--- a/src/components/checkout/templates/container/style.js
+++ b/src/components/checkout/templates/container/style.js
@@ -374,7 +374,9 @@ export function getContainerStyle({ id, tag, CONTEXT, CLASS, ANIMATION } : { id 
         }
 
         #${ id }.${ tag }-context-${ CONTEXT.IFRAME } .${ CLASS.OUTLET } iframe {
-            width: 100%;
+            width: 1px;
+            min-width: 100%;
+            *width: 100%;
             height: 100%;
         }
 
--- a/src/components/checkout/component.js
+++ b/src/components/checkout/component.js
@@ -579,7 +579,7 @@ if (Checkout.isChild()) {
             // $FlowFixMe
             Object.assign = extend;
         } catch (err) {
-            // pass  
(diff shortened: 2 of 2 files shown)


khanghoang:
@bluepnume is there any particular reason for this style that you can think of? https://github.com/paypal/paypal-checkout/blob/master/src/components/checkout/templates/container/style.js#L173
and this one https://github.com/paypal/paypal-checkout/blob/master/src/components/checkout/templates/container/style.js#L177

src/components/checkout/templates/container/style.js
```diff
@@ -374,7 +374,9 @@ export function getContainerStyle({ id, tag, CONTEXT, CLASS, ANIMATION } : { id
         }
 
         #${ id }.${ tag }-context-${ CONTEXT.IFRAME } .${ CLASS.OUTLET } iframe {
-            width: 100%;
+            width: 1px;
```
bluepnume: Not sure I understand the fix -- could you explain the 1px and the *width?

bluepnume:
The styles you linked -- I don't think they're 100% necessary, so long as the text stays centered. (BTW no need to work on this on the weekend, it's not hugely urgent!)

src/components/checkout/templates/container/style.js
khanghoang: The close button is clipped on iphone 5 portrait mode
iphone 5 ios 10 3 14e8301 2017-10-29 17-05-26
more details here https://stackoverflow.com/questions/23083462/how-to-get-an-iframe-to-be-responsive-in-ios-safari?answertab=votes#tab-top

khanghoang:
@bluepnume I do this just because just because I got nothing else to do. It's just for fun.

src/components/checkout/templates/container/style.js
bluepnume: Makes sense! I think we can drop the *width here though since we only need to support IE9+

khanghoang:
@bluepnume is there any particular reason for these styles that you can think of? https://github.com/paypal/paypal-checkout/blob/master/src/components/checkout/templates/container/style.js#L173
and this one https://github.com/paypal/paypal-checkout/blob/master/src/components/checkout/templates/container/style.js#L177

bluepnume:
I don't see any reason for them to be there, you can feel free to remove them so long as the text in that overlay stays centered.

bluepnume:
Merged! Thanks @khanghoang

khanghoang:
@bluepnume this is WIP PR, I wanted to confirm with you about my concerns about those styles before I make any change. 
This PR didn't fix the overflowed text, it only fixes the clipped close button. Therefore, I need to make another PR.

bluepnume:
Ah sorry, didn't realize it was WIP

PR #586, opened 2018-01-31
Correct PayerID mapping in return uri for NativeXO

andreabondi:
data.payerID was undefined due to a different capitalisation in the return uri between Hermes and Checkout.js via NativeXO. Adding an if statement to ensure any backwards compatibility if any other client is using this function.

Andrea and Donald

### DIFF
--- a/src/integrations/popupBridge.js
+++ b/src/integrations/popupBridge.js
@@ -84,13 +84,9 @@ function extractDataFromQuery(query : Object) : Object {
         paymentToken: query.token,
         billingToken: query.ba_token,
         paymentID:    query.paymentId,
-        payerID:      query.payerId,
+        payerID:      query.PayerID,
         intent:       query.intent
     };
-
-    if (!data.payerID) {
-        data.payerID = query.PayerID;
-    }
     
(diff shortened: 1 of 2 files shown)


bluepnume:
Thanks for the fix, good catch! Next release will most likely be wednesday of next week.

src/integrations/popupBridge.js
```diff
@@ -88,6 +88,10 @@ function extractDataFromQuery(query : Object) : Object {
         intent:       query.intent
     };
 
+    if (!data.payerID) {
```
bluepnume: Any chance you could make this change on line 87 in the original mapping?

bluepnume:
Could you make sure the tests are accounting for this fix as well? Probably would just need to change to `PayerID` here and make sure tests still pass: https://github.com/paypal/paypal-checkout/blob/master/test/tests/common.js#L620

dasx199:
@bluepnume Does this look OK now?

bluepnume:
Looks good! Merged

bluepnume:
This should be released now -- please clear your browser cache and let me know if you still see the problem.

PR #622, opened 2018-02-28
small typo on the Reactjs implementation

opmftw:
just enclosing the boolean in curly brackets, otherwise it throws an error, thanks for the project bro


bluepnume:
Good catch, thanks!

PR #629, opened 2018-03-07
SPB for LATAM

vishakha94:
Requires review


src/button/validate.js
```diff
@@ -22,7 +22,7 @@ export function validateButtonLocale(locale : string) {
     }
 }
 
-export function validateButtonStyle(style : Object = {}) {
+export function validateButtonStyle(style : Object = {}, props : Object ) {
```
bluepnume: Can we make this more specific?

```javascript
export function validateButtonStyle(style : Object = {}, { locale } : { locale : string }) {
```

src/constants/country.js
```diff
@@ -262,3 +262,8 @@ export const LANG_TO_DEFAULT_COUNTRY = {
     [LANG.TR]: COUNTRY.TR,
     [LANG.ZH]: COUNTRY.CN
 };
+
+export const INSTALLMENT_COUNTRY = {
```
bluepnume: Would it be more appropriate to make this an array?

src/constants/country.js
```diff
@@ -262,3 +262,8 @@ export const LANG_TO_DEFAULT_COUNTRY = {
...
+    BR: 'BR',
```
bluepnume: Can we use country constants here

src/button/validate.js
```diff
@@ -127,6 +128,31 @@ export function validateButtonStyle(style : Object = {}) {
...
+
+    validateRegionSpecificButton(style, props.locale);
+
+}
+
+export function validateRegionSpecificButton(style : Object = {}, locale : string) {
+
+    let country = locale.split('_')[1];
+
+    let isInstallmentCountry = Object.keys(INSTALLMENT_COUNTRY).find(elem => {
```
bluepnume: Probably if this was an array we could just do `(INSTALLMENT_COUNTRY.indexOf(country) !== -1)`

src/button/validate.js
```diff
@@ -127,6 +128,31 @@ export function validateButtonStyle(style : Object = {}) {
             throw new Error(`style.${ BUTTON_STYLE_OPTIONS.TAGLINE } is not allowed for ${ BUTTON_LAYOUT.VERTICAL } layout - got ${ tagline }`);
         }
     }
+
+    validateRegionSpecificButton(style, props.locale);
+
+}
+
+export function validateRegionSpecificButton(style : Object = {}, locale : string) {
```
bluepnume: Like the idea of this 👍

src/constants/country.js
vishakha94: Hve grouped this in a separate object for the purpose of validation. During validation, it checks whether the locale is found in this specific group.

src/button/validate.js
```diff
@@ -43,7 +43,8 @@ export function validateButtonStyle(style : Object = {}) {
         [ BUTTON_STYLE_OPTIONS.TAGLINE ]:      tagline,
         [ BUTTON_STYLE_OPTIONS.LAYOUT ]:       layout,
         [ BUTTON_STYLE_OPTIONS.MAXBUTTONS ]:   maxbuttons,
-        [ BUTTON_STYLE_OPTIONS.HEIGHT ]:       height
+        [ BUTTON_STYLE_OPTIONS.HEIGHT ]:       height,
+        [ BUTTON_STYLE_OPTIONS.INSTALLMENTNUM ]:  installmentnum
```
bluepnume: Can we call this `installmentperiod`?

src/button/validate.js
vishakha94: can be done. However, we may need other props ( like locale this time) for the purpose of button validation.

src/button/validate.js
vishakha94: :(

src/button/validate.js
vishakha94: 💯

src/constants/country.js
bluepnume: Also, thinking about it -- can we call this `ALLOWED_INSTALLMENT_COUNTRIES` just to be a little more explicit? Otherwise it's not clear whether installments are enabled by default for these countries or just allowed.

src/button/validate.js
vishakha94: ok.

src/button/template/componentTemplate.jsx
```diff
@@ -83,15 +83,18 @@ function renderFundingIcons({ cards, fundingicons } :
...
-function renderContent(text : string, { locale, color, branding, logoColor, funding, env, cards } :
-    { locale : LocaleType, color : string, branding? : boolean, logoColor? : string, funding? : FundingSelection, env : string, cards : Array<string> }) : JsxHTMLNode {
+function renderContent(text : string, { label, locale, color, branding, logoColor, funding, env, cards, dynamicContent } :
+    { contentLabel?: string, locale : LocaleType, color : string, branding? : boolean, logoColor? : string, funding? : FundingSelection, env : string, cards : Array<string>, dynamicContent : Object }) : JsxHTMLNode {
 
     let content = getLocaleContent(locale);
 
     return jsxRender(text, {
 
         text(value : string) : JsxHTMLNode {
```
bluepnume: Would it be possible to add a new key/function to `jsxRender`, so we don't overload `text`? Then we could have support for `foo {break} bar`, and the parser that's already there would take care of regexing the string and splitting it up. It would also mean we'd be able to use multiple breaks in a string if we ever need to.

src/constants/country.js
vishakha94: 👍

src/button/template/componentTemplate.jsx
vishakha94: I would need to check the feasibility for that. jsxRender function is tightly coupled with the current pattern of content.

src/button/config.js
```diff
@@ -144,6 +147,19 @@ export const BUTTON_CONFIG : ButtonConfig = {
...
+            label:     `{ content: installment }`,
+            logoLabel: `{ logo: ${ BUTTON_LOGO.PP } } { logo: ${ BUTTON_LOGO.PAYPAL } }`,
+
+            allowPrimary: true,
+            allowPrimaryVertical:   false,
+            allowPrimaryHorizontal: true,
+            allowSecondaryVertical:   false,
+            allowSecondaryHorizontal: false,
+            requiresTextWrap: true,
+            allowedDynamicContent: { 'installmentnum': 'installment_dynamic' }
```
bluepnume: This abstraction is cool, but I'm a little worried that it might be a bit specific to the current use case; i.e. "use content A or content B depending on another style key", and that might not scale to other use cases.

Would it be possible instead to make `label` a function? Something like:

```
label: (props.style) => `{ content: ${ props.style.installmentperiod ? 'installment' : 'installment_period' } }`
```

src/button/template/componentTemplate.jsx
```diff
@@ -157,20 +149,31 @@ function renderContent(text : string, { label, locale, color, branding, logoColo
 
         separator() : JsxHTMLNode {
             return <span class={ CLASS.SEPARATOR }></span>;
+        },
+
+        break(value : string) : JsxHTMLNode {
+            let className = `${ CLASS.TEXT }`;
+            return <span class={ className }>{ value.split('<br>')[0] }<br>{ value.split('<br>')[1] }</br></span>;
```
vishakha94: I would like to split it dynamically here but can't think of a proper way.

src/button/validate.js
bluepnume: :D

src/lib/jsx.js
```diff
@@ -117,7 +117,13 @@ export function jsxRender(template : string, renderers : { [string] : (string) =
                 return text;
             }
 
-            return renderers.text(text);
+            // handle break here
+            let regex = /\<br\>/;
```
bluepnume: Do we need this logic? Wouldn't line `113` already call the break function?

src/button/template/componentTemplate.jsx
bluepnume: I think we can just `return <br />` here and simplify this a lot. The jsxRender function should take care of placing the break in the right place in the content.

src/button/template/content.json
```diff
@@ -615,12 +615,13 @@
             "safer_tag": "A maneira f\u00e1cil e segura de enviar pagamentos.",
             "later_tag": "Compre agora e pague depois.",
             "pay": "Pague com {logo:paypal}",
-            "buynow": "{logo:pp} {logo:paypal} Comprar agora"
+            "buynow": "{logo:pp} {logo:paypal} Comprar agora",
+            "installment": "{logo:pp} {logo:paypal}Pagamentos <br>sem juros",
```
bluepnume: So, like, this would become "{logo:pp} {logo:paypal}Pagamentos{break}sem juros"

src/button/template/content.json
```diff
@@ -3342,15 +3343,15 @@
         "es": {
             "checkout": "Pagar con {logo:pp} {logo:paypal}",
             "safer_tag": "La forma r\u00e1pida y segura de pagar",
-
             "later_tag": "Compre ahora y pague m\u00e1s adelante.",
             "pay": "Pagar con {logo:paypal}",
-            "buynow": "{logo:pp} {logo:paypal} Comprar ahora"
+            "buynow": "{logo:pp} {logo:paypal} Comprar ahora",
+            "installment": "Pagos sin<br>inter\u00e9s",
+            "installment_period": "Pague hasta<br>[installmentperiod]x sin inter\u00e9s"
```
bluepnume: I wonder if instead of introducing a new style of placeholder here (`[foo]` vs `{foo}`) which might cause confusion, like "what's the difference between the two styles?", can we use `{installmentperiod}`? Then we can just define another function which we pass to `jsxRender` which handles the `installmentperiod` case.

src/button/validate.js
```diff
@@ -127,6 +128,29 @@ export function validateButtonStyle(style : Object = {}) {
             throw new Error(`style.${ BUTTON_STYLE_OPTIONS.TAGLINE } is not allowed for ${ BUTTON_LAYOUT.VERTICAL } layout - got ${ tagline }`);
         }
     }
+
+    validateRegionSpecificButton(style, props.locale);
```
bluepnume: Shall we also add some validation to make sure installmentperiod is a number, and one of the accepted values we expect?

src/button/validate.js
```diff
@@ -127,6 +128,29 @@ export function validateButtonStyle(style : Object = {}) {
...
+
+}
+
+export function validateRegionSpecificButton(style : Object = {}, locale : string) {
+
+    let country = locale.split('_')[1];
+
+    let isInstallmentAllowedCountry = ALLOWED_INSTALLMENT_COUNTRIES.indexOf(country) !== -1;
+
+    if ( !isInstallmentAllowedCountry && style.installmentperiod ) {
```
bluepnume: Let's use `style[BUTTON_STYLE_OPTIONS.INSTALLMENTPERIOD]` then any future name changes only need to happen in one place.

src/button/validate.js
```diff
@@ -127,6 +128,29 @@ export function validateButtonStyle(style : Object = {}) {
...
+        throw new Error(`style.${ BUTTON_STYLE_OPTIONS.INSTALLMENTPERIOD } is invalid for country: ${ country }`);
+    }
+
+    if ( !isInstallmentAllowedCountry && style.label === BUTTON_LABEL.INSTALLMENT ) {
+        throw new Error(`Unexpected label: style.${ style.label } for country: ${ country }`);
+    }
+
+    if ( isInstallmentAllowedCountry && style.installmentperiod && style.label !== BUTTON_LABEL.INSTALLMENT ) {
```
bluepnume: Same here with `style[BUTTON_STYLE_OPTIONS.INSTALLMENTPERIOD]`

src/button/template/componentStyle/responsive.js
```diff
@@ -35,6 +35,12 @@ export function buttonResponsiveStyle({ height, cardNumber = 4 } : { height? : n
                 .${ CLASS.BUTTON }.${ CLASS.BRANDING }-${ BUTTON_BRANDING.UNBRANDED } {
                     font-size: ${ max(perc(buttonHeight, 45), 10) }px;
                 }
+                
+                .${ CLASS.BUTTON } .${ CLASS.TEXT }.${ CLASS.WRAPTEXT } {
```
bluepnume: Do we actually use this class anywhere in a template?

src/button/template/content.json
vishakha94: The square brackets here denote the `dynamic content` which is being plugged in. Now, if we replace [] with {}, we will need an additional specific function to be passed to the jsxRender. Whereas, in the present scenario, we are handling the replacement of foo [foo] inside context function itself. It picks up the foo from dynamicContent object without having to explicitly call it. What if we have tens of different dynamic content? Then it might not be a good idea to feed different functions for the respective contents to jsxRender.
As for the scalability, all we got to take care of is feeding the dynamicCOntent obj with different value and the [foo] would be replaced automatically in the content func.

vishakha94: https://github.com/vishakha94/paypal-checkout/blob/latam/src/button/template/componentTemplate.jsx#L128-L133

src/button/template/componentTemplate.jsx
vishakha94: Resolved.

src/lib/jsx.js
vishakha94: Yes, with the current standpoint.

src/button/validate.js
vishakha94: yes

src/button/validate.js
vishakha94: 👍

src/button/validate.js
vishakha94: 👍

src/button/template/componentStyle/responsive.js
vishakha94: Will remove.

src/button/template/componentTemplate.jsx
vishakha94: Yes, added a separate break function.

src/button/template/componentTemplate.jsx
bluepnume: 👍

src/lib/jsx.js
bluepnume: 👍

test/screenshot/config.js
```diff
@@ -324,3 +328,216 @@ for (let color of getButtonConfig('paypal', 'colors')) {
         }
     });
 }
+
+
+/** INSTALLMENT BUTTON CONFIG **/
```
vishakha94: For screenshots, added the config for installment btn separately. My early plan was to hijack the buttonConfig object and add the locale prop to button. However, our new friend flow doe not like adding properties to `sealed objects`. SO, I have duplicated the tests. COuld have done a bunch of ifs and lese in the existing one but thought of keep it separate and clean.
Creating unsealed objects would have been too much explicit declarations which we dont want.

test/screenshot/config.js
```diff
@@ -324,3 +328,216 @@ for (let color of getButtonConfig('paypal', 'colors')) {
...
+                    height: 45,
+                    size:   BUTTON_SIZE.SMALL
+                }
+            }
+        });
+    }
+
+
+    let installmentnum = ALLOWED_INSTALLMENT_PERIOD[locale.split('_')[1]];
+    for (let installmentperiod of installmentnum) {
```
vishakha94: We are also reitertaing for all intallment periods and installment locales!

vishakha94:
Build passed! Pls review. @bluepnume

PR #638, opened 2018-03-20
Altpay

vishakha94:


test/screenshot/config.js
```diff
@@ -324,3 +328,216 @@ for (let color of getButtonConfig('paypal', 'colors')) {
         }
     });
 }
+
+
+/** INSTALLMENT BUTTON CONFIG **/
+
+for (let locale of INSTALLMENT_BTN_LOCALE) {
```
bluepnume: So, one thing about these tests is they're really slow. So maybe let's not test every locale and combination. Can we, for now, maybe just render two buttons -- one with installment period and one without?

Also can we get one button for each of the new logos?

test/tests/button/validation.js
```diff
@@ -124,6 +129,44 @@ let buttonConfigs = [
         }))
     },
 
+    {
```
bluepnume: Can we add some tests for the new logos?

test/screenshot/config.js
```diff
@@ -323,4 +320,274 @@ for (let color of getButtonConfig('paypal', 'colors')) {
...
+            style:  {
+                color,
+                layout:     'vertical',
+                maxbuttons: 4
+            },
+            funding: {
+                allowed: [ FUNDING.MYBANK ]
+            }
+        }
+    });
```
bluepnume: This all looks great 👍

test/screenshot/config.js
vishakha94: ok

test/tests/button/validation.js
vishakha94: they are checked in now

PR #652, opened 2018-04-09
alt pay button: url config

vishakha94:


src/integrations/checkout.js
```diff
@@ -31,7 +31,8 @@ export function determineUrl(env : string, fundingSource : ?string, payment : st
         return config.guestUrls[env];
     }
 
-    if (fundingSource === FUNDING.IDEAL) {
```
bluepnume: This'll work for now. In future we should probably define a map of funding method -> url though.

vishakha94: I agree. Was thinking about this.

PR #654, opened 2018-04-12
enable alt pay button only for commit=true

vishakha94:


bluepnume:
Perfect :) thanks for taking care of this!

PR #667, opened 2018-04-23
Shipping Options changes

abarco:
Requiring unique IDs for shipping_options.
Merging responses for get and execute payment details with shipping options selection from checkout.


bluepnume:
Looks good. Will merge once tests pass 👍

PR #696, opened 2018-05-23
feat(label): Label support for vertical PayPal button

mcntsh:
Buttons will now display labels in the vertical layout which have `allowPrimaryVertical` set to
`true` in the button config. Right now this is limited to the PayPal button labels
(PAYPAL/CHECKOUT/PAY/BUYNOW)


src/button/template/componentTemplate.jsx
```diff
@@ -38,14 +38,22 @@ function getLocaleContent(locale : LocaleType) : Object {
     return componentContent[country][lang];
 }
 
-function determineButtons({ label, color, sources, multiple } : { label : $Values<typeof BUTTON_LABEL>, color : string, sources : FundingList, multiple : boolean }) :
+function determineCanRenderLabel({ label, source, multiple, layout } : { label : $Values<typeof BUTTON_LABEL>, source : FundingSource, multiple : boolean,  layout : $Values<typeof BUTTON_LAYOUT> }) : boolean {
+    if (!multiple || layout === BUTTON_LAYOUT.VERTICAL) {
```
vishakha94: do we need to check for vertical layout here?

bluepnume: Totally understand the logic here but it's a little hard to wrap my head around this boolean logic :) (honestly, at some point we need to refactor all of this label<->funding logic because it's starting to get pretty hairy)

For now I wonder if we could do something like:

```javascript
function determineLabel({ label, source, multiple, layout }) {

    let defaultLabel = fundingToDefaultLabel(source);
    let labelMatchesFunding = (labelToFunding(label) === source);

    // If chosen label is not for this funding source, display the default label
    if (!labelMatchesFunding) {
        return defaultLabel;
    }

    // If there are multiple horizontal buttons, display the default label
    if (layout === BUTTON_LAYOUT.HORIZONTAL && multiple) {
        return defaultLabel;
    }

    // Otherwise display the chosen label
    return label;
}

let buttonLabel = determineLabel({ label, source, multiple, layout });
```

bluepnume:
LGTM, one suggested change just to make it clearer how we're deciding whether to show the selected label or not.

src/button/template/componentTemplate.jsx
mcntsh: Thanks for this! Agreed on the boolean logic, and @vishakha94 had similar comments. The reason I went with this type of a function rather than one that returns the label is because the same logic is used to [determine whether or not to use the `label` or the `logoLabel`](https://github.com/mcntsh/paypal-checkout/blob/b4ac674f723480fd12a9032f8211e5a494ec37dc/src/button/template/componentTemplate.jsx#L175).

I'm going to figure out how to consolidate better.

mcntsh: Yeah I definitely see what you mean. The reason why I did it this way was because I needed to re-use that logic to [determine the `buttonText` in `renderButton`](https://github.com/mcntsh/paypal-checkout/blob/b4ac674f723480fd12a9032f8211e5a494ec37dc/src/button/template/componentTemplate.jsx#L175), but let me take this back and see how I can clean this up using the code you provided.

Thanks for taking a look!

PR #697, opened 2018-05-24
feat(credit): PayPal Credit button will use DE logo for DE locales

mcntsh:
Adding in the DE PPC logo ("Ratenzahlung Powered by PayPal"). The `logoLabel` will use this logo in
the template if passed the DE locale.  Perhaps it should be separated out from credit into its own
funding option?


src/button/config.js
```diff
@@ -157,7 +157,11 @@ export const BUTTON_CONFIG : ButtonConfig = {
 
     [ BUTTON_LABEL.CREDIT ]: {
         label:     `{ logo: ${ BUTTON_LOGO.PP } } { logo: ${ BUTTON_LOGO.PAYPAL } } { logo: ${ BUTTON_LOGO.CREDIT } }`,
-        logoLabel: `{ logo: ${ BUTTON_LOGO.PP } } { logo: ${ BUTTON_LOGO.PAYPAL } } { logo: ${ BUTTON_LOGO.CREDIT } }`,
+        logoLabel: ({ locale }) => {
+            return locale.country === COUNTRY.DE
```
vishakha94: let's create an object and make this configurable instead of a condition?
Ex:
`const creditLogoForLocale = { 'DE': 'CREDIT_LOGO_DE }` and then read it off that?

src/resources/fundingLogos/de_credit_white.svg
```diff
@@ -0,0 +1,7 @@
+<svg width="135" height="32" viewBox="0 0 135 32" preserveAspectRatio="xMinYMin meet" xmlns="http://www.w3.org/2000/svg">
```
bluepnume: Not a review comment, just a heads up; at some point I'd like to turn these SVGs into jsx or something so we can programatically generate them and save shipping a bunch of redundant SVGs. Maybe that jsx-rendering function can take the locale as a parameter, and we can push the logic down into there.

bluepnume: Actually since it potentially makes this PR a whole bunch simpler, let me see how easy that would be...

bluepnume:
OK I fixed flow and merged in the jsx changes in the other pr. Feel free to use for locale-specific logos when you're ready.

mcntsh:
@bluepnume Awesome, I pushed some changes which combine all of the Credit variations into one JSX template. The `fetchLocalizedLabels` logic is still needed because the DE Credit logo has a different layout than the regular Credit logo.

CI is failing but tests pass locally.

bluepnume:
Could we encapsulate this logic into the config.js file for the credit button?

```
[ BUTTON_LABEL.CREDIT ]: {
    label: ({ locale }) => {
        if (locale.country === COUNTRY.DE) {
            return `{ logo: ${ BUTTON_LOGO.CREDIT } }`;
        }

        return `{ logo: ${ BUTTON_LOGO.PP } } { logo: ${ BUTTON_LOGO.PAYPAL } } { logo: ${ BUTTON_LOGO.CREDIT } }`;
    }
}
```

Figure this might be a bit more maintainable up until the point where we're sure localized label layouts are more of a common use case. Right now it feels like a corner case that we should solve with a simple function.

mcntsh:
@bluepnume That makes sense. This is similar to how I had it working originally, but Vishakha and I both agreed that it would make sense to extract it into something that can be covered through configuration for future cases. 

But since we are likely going to shift this logic into JSX templates, and this particular thing is more of an edge case, I reverted it back to how it was handled originally – through the `label` and `logoLabel` template function 👍

PR #726, opened 2018-06-27
adding sofort to countries AT, BE, ES, IT, NL and logo change

aditya27marri:
Adding sofort eligibility in AT, BE, ES, IT, NL other than  DE.
changes to sofort logo to use klarna logo.


bluepnume:
- Please remove any of the `-old.png` files from this commit. These should not be committed.

PR #748, opened 2018-07-16
APM blacklisting functionality

vishakha94:
- Add blacklisting capability for APM buttons
- Disables IDEAL button across the board 
- Skip IDEAL tests/screenshots
- Blacklisting of a few merchants from APM buttons


src/button/component.jsx
```diff
@@ -76,6 +76,24 @@ function isCreditDualEligible(props) : boolean {
     return true;
 }
 
+function determineAPMBlacklist (allowed, disallowed) : Object {
+    Object.getOwnPropertyNames(ALTERNATE_PAYMENT_METHOD).forEach(apm => {
```
abarco: FYI, there's for.. in: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

vishakha94: yes, but it takes in all enumerable properties, though it'll be just better to use ^ instead.

bluepnume: FYI best to use: `for (let key in Object.keys(obj)) {`

PR #753, opened 2018-07-17
fix(redirect): Fixing `actions.redirect` from within iOS webviews.

mcntsh:
REF: https://github.com/paypal/paypal-checkout/issues/736

In a native iOS webview, the merchant page cannot perform a redirect until the iframe has been
closed. So the logic is now procedural just for iOS (close -> redirect)

### DIFF
--- a/src/button/component.jsx
+++ b/src/button/component.jsx
@@ -14,7 +14,7 @@ import { SOURCE, ENV, FPTI, FUNDING, BUTTON_LABEL, BUTTON_COLOR,
 import { redirect as redir, checkRecognizedBrowser,
     getBrowserLocale, getSessionID, request, getScriptVersion,
     isIEIntranet, isEligible,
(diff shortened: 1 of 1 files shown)


src/button/component.jsx
```diff
@@ -551,15 +551,11 @@ export let Button : Component<ButtonOptions> = create({
...
-                        if (isIosWebview()) {
-                            return ZalgoPromise.try(() =>
-                                actions.close()
-                            ).then(() =>
-                                redir(win || window.top, url || data.returnUrl)
-                            );
-                        }
-
-                        return ZalgoPromise.all([ actions.close(), redir(win || window.top, url || data.returnUrl) ]);
+                        return ZalgoPromise.try(() => {
```
bluepnume: Probably need to do the same in `onCancel` too yeah?

PR #754, opened 2018-07-17
Reverting https://github.com/paypal/paypal-checkout/pull/689/ 

abarco:
PR https://github.com/paypal/paypal-checkout/pull/689/  is causing issues when logging with `x-app-name`. Reverting until we can determine the root cause

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -727,10 +727,7 @@ export let config = {
     },
 
     get loggerUrl() : string {
-        let isTestExperiment = Math.random() < config.loggerThrottlePercentage;
-        let loggerUrl = isTestExperiment ? config.loggerUri : config.hermesLoggerUri;
-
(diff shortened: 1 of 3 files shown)


bluepnume:
Merged. Fyi @mstuart 

- There's the issue with `x-app-name`. Probably need to make this a configurable CORS header in the beaver middleware.
- Looks like `/xoplatform/logger` is the old closed-source beaver url? that's gonna be a problem. stuff changed with the open-source beaver and pretty sure it's not backwards compatible.

mstuart:
Not sure on the `x-app-name` issue.  Looks like it's an allowed header [1] and hermes and xoon are able to send logs over from their paypalobjects.com script to paypal.com too.

screen shot 2018-07-19 at 14 10 09

PR #757, opened 2018-07-19
Point client-side logs at /xoplatform/logger, not /webapps/hermes

mstuart:
*Changes*
- Running a 5% test for ramping logs (src/lib/logger) to
  /xoplatform/logger
- Pointing all beacon logs (src/lib/beacon) to /xoplatform/logger

*NOTE* This is similar to https://github.com/paypal/paypal-checkout/pull/260 which never actually landed :-)  Now we're getting around to finally ramping up our logs onto another URL.

img

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -431,7 +431,7 @@ export let config = {
 
     hermesLoggerUri: `/webapps/hermes/api/logger`,
 
-    loggerUri: `/xoplatform/logger`,
+    loggerUri: `/xoplatform/logger/api/logger`,
 
     loggerThrottlePercentage: 0.05, // 5%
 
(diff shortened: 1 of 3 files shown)


mstuart:
@dbrain @abarco Still haven't been able to repro that `x-app-name` error.  I'm pretty confident it would never happen, otherwise Hermes and Xoon (who both have their scripts hosted on paypalobjects.com) wouldn't be able to log anything at the paypal.com domain.

mstuart:
The main fix here is at https://github.com/paypal/paypal-checkout/pull/757/files#diff-175e532f7fea7eb24a3931edf58f2a7eR434.  Previously, wasn't hitting the correct endpoint :-|

mstuart:
All good now.  As you can see the `x-app-name` header is already allowed w/ CORS.  Hosted this up on paypalobjects.com and ran some tests.  All 200s.  👍 

screen shot 2018-07-19 at 14 13 20
screen shot 2018-07-19 at 14 13 07
screen shot 2018-07-19 at 14 12 57
screen shot 2018-07-19 at 14 12 32

PR #766, opened 2018-07-27
feat(callback-api): Adding `onShippingChange` callback prop

mcntsh:


src/checkout/component.js
```diff
@@ -333,6 +334,19 @@ export let Checkout : Component<CheckoutPropsType> = create({
             }
         },
 
+        onShippingChange: {
+            type:     'function',
+            required: false,
+
+            decorate(original) : Function {
+                return function decorateOnShippingChange(data, actions = {}) : ZalgoPromise<void> {
+                    return ZalgoPromise.try(() => {
```
bluepnume: Looks like this decorator isn't doing much. Can we remove it?

src/integrations/popupBridge.js
```diff
@@ -163,6 +165,9 @@ function renderThroughPopupBridge(props : Object, popupBridge : PopupBridge) : Z
 
         } else if (opType === OPTYPE.CANCEL) {
             return onCancel(data, actions);
+        
+        } else if (opType === OPTYPE.SHIPPING_CHANGE) {
```
bluepnume: Have we opened a conversation with the teams that own the popup bridge so they can support this optype? (native sdk and braintree)?

Actually I wonder if it's even possible...

src/button/component.jsx
```diff
@@ -614,6 +614,46 @@ export let Button : Component<ButtonOptions> = create({
             }
         },
 
+        onShippingChange: {
+            type:     'function',
+            required: false,
+            timeout:  __TEST__ ? 500 : 10 * 1000,
```
bluepnume: Can probably remove this since you're timing out in the decorator.

src/button/component.jsx
```diff
@@ -614,6 +614,46 @@ export let Button : Component<ButtonOptions> = create({
...
+                        [ FPTI.KEY.STATE ]:              FPTI.STATE.CHECKOUT,
+                        [ FPTI.KEY.TRANSITION ]:         FPTI.TRANSITION.CHECKOUT_SHIPPING_CHANGE,
+                        [ FPTI.KEY.BUTTON_SESSION_UID ]: this.props.buttonSessionID
+                    });
+
+                    flushLogs();
+                    let timeout = __TEST__ ? 500 : 10 * 1000;
+
+                    let reject = (...args) => {
+                        if (actions.reject && typeof actions.reject === 'function') {
```
bluepnume: Won't `actions.reject` always be present?

src/button/component.jsx
```diff
@@ -614,6 +614,46 @@ export let Button : Component<ButtonOptions> = create({
...
+                            return ZalgoPromise.try(() => {
+                                return actions.reject.call(null, ...args);
+                            });
+                        }
+                    };
+
+                    return ZalgoPromise.try(() => {
+                        return original.call(this, data, { ...actions, reject });
+                    }).timeout(timeout, new Error(`Timed out waiting ${ timeout }ms for payment`)).catch(err => {
+                        if (this.props.onError) {
```
bluepnume: I'd say let's not call the `onError` prop here, since we're handling the error in hermes and showing an error page. Typically for those cases we don't propagate the failure back up to the merchant.

src/checkout/component.js
```diff
@@ -32,6 +32,7 @@ function addHeader(name, value) : void {
 type CheckoutPropsType = {
     payment? : () => ZalgoPromise<string>,
     onAuthorize : ({ returnUrl : string }, { redirect : (?CrossDomainWindowType, ?string) => ZalgoPromise<void> }) => ?ZalgoPromise<void>,
+    onShippingChange? : (string) => ?ZalgoPromise<void>,
```
bluepnume: Guessing `string` should be an object type instead

test/windows/checkout/index.js
```diff
@@ -55,6 +55,22 @@ if (action === 'checkout') {
...
+    window.xprops.payment().then(paymentToken => {
+
+        window.xprops.onShippingChange({
+            paymentID:    paymentToken,
+            city:         'XXXXX',
+            state:        'YY',
+            postal_code:  '11111',
+            country_code: 'YY'
+        }, {
+            reject: (callback) => {
```
bluepnume: I don't see anyone passing this callback?

mcntsh: Hmm perhaps there's a better way to test this, but I'm just trying to assert that the action can be passed up from the child and called in the parent.

test/tests/button/happy.js
```diff
@@ -870,6 +870,58 @@ for (let flow of [ 'popup', 'iframe' ]) {
...
+                payment() : string | ZalgoPromise<string> {
+                    return generateECToken();
+                },
+
+                onAuthorize() : void {
+                    return done(new Error('Expected onAuthorize to not be called'));
+                },
+
+                onShippingChange(data, actions) : void {
+                    return actions.reject(() => {
```
bluepnume: I think we can just do `actions.reject().then(() => done())` here right? Then it's a little closer to the actual usage of reject.

bluepnume:
One minor comment otherwise looks good to me!

PR #770, opened 2018-08-01
Update Documentations

rzhe97:
Added link to download tool.
Grammatical fixes.


bluepnume:
Hi, sorry, not sure your grammar fix is actually correct; also we're hoping anyone wanting to use this module with npm is already familiar with npm, otherwise ideally they should be using the CDN integration.

PR #771, opened 2018-08-02
fix(config): Removing ELV from horizontal layouts

mcntsh:


test/tests/button/funding.js
```diff
@@ -574,14 +574,14 @@ describe(`paypal button component funding mix`, () => {
         }, '#testContainer');
     });
 
-    it('should render elv in vertical layout', (done) => {
+    it('should not render elv in vertical layout', (done) => {
```
bluepnume: should this be "horizontal"?

mcntsh: Hah, my bad! This is what peer reviews are for.

jmoney999:
What is this message pertaining to

PR #809, opened 2018-08-21
Experimenting with ramping presentment of Venmo without a cookie

abarco:
To measure engagement, we'll be experimenting with the presentment of Pay With Venmo button for our mobile population without a Venmo cookie.

### DIFF
--- a/src/button/component.jsx
+++ b/src/button/component.jsx
@@ -77,6 +77,7 @@ function isCreditDualEligible(props) : boolean {
 }
 
 let creditThrottle;
+let venmoThrottle;
 
 let localeThrottle = getThrottle('locale_resolution_rule', 50);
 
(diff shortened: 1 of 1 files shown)


src/button/component.jsx
```diff
@@ -426,10 +427,22 @@ export let Button : Component<ButtonOptions> = create({
...
             decorate({ allowed = [], disallowed = [] } : Object = {}, props : ButtonOptions) : {} {
 
+                // remove Venmo from our allowed list if the rendering device is not a mobile one
                 if (allowed && allowed.indexOf(FUNDING.VENMO) !== -1 && !isDevice()) {
                     allowed = allowed.filter(source => (source !== FUNDING.VENMO));
                 }
 
+                // Uncookied venmo ramp. Even without 'pwv' cookie, we'll be rendering venmo button
+                // for a sample of the mobile population. 
+                if (allowed && allowed.indexOf(FUNDING.VENMO) === -1 && isDevice()) {
```
bluepnume: I would also check if it's in remembered group -- if so should be ineligible for the experiment.

abarco: ah, good call

PR #839, opened 2018-08-30
Point client-side logs at /xoplatform/logger, not /webapps/hermes

mstuart:
*Changes*
- Running a 5% test for ramping logs (src/lib/logger) to
  /xoplatform/logger
- Pointing all beacon logs (src/lib/beacon) to /xoplatform/logger

img

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -431,7 +431,7 @@ export let config = {
 
     hermesLoggerUri: `/webapps/hermes/api/logger`,
 
-    loggerUri: `/xoplatform/logger`,
+    loggerUri: `/xoplatform/logger/api/logger`,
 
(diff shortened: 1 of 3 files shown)


mstuart:
Btw, the last time we tried this, `/xoplatform/logger` didn't work w/ non .paypal.com domains and didn't accept an `x-app-name` header.  Now that has been fixed 👍 

screen shot 2018-08-30 at 16 35 15
screen shot 2018-08-30 at 16 35 07

Tested this out w/ `local:8000` pointing at `paypal.com/xoplatform/logger` and see that `x-app-name` and `local:8000` are allowed w/ CORS.  Should be good for the next release.

Happy to help w/ that too.

PR #850, opened 2018-09-07
Bml funding test

YSNIS:


src/lib/throttle.js
```diff
@@ -132,3 +136,55 @@ export function getReturnToken() : ?string {
...
     }
 }
+
+type CustomThrottle = Throttle & {
+    isActive : () => boolean
+};
+
+export function buildFundingLogoThrottle(props : Object) : CustomThrottle {
+
+    let emptyThrottle = {
```
bluepnume: Seems like rather than creating an entire mock throttle here along with a `CustomThrottle` type, could just `return`, right? Then do a null check before using the throttle to check if it's active? Might simplify this case a little.

src/lib/throttle.js
```diff
@@ -132,3 +136,55 @@ export function getReturnToken() : ?string {
...
+        isDisabled:   () => true,
+        getTreatment: () => '',
+        log:          () => emptyThrottle,
+        logStart:     () => emptyThrottle,
+        logComplete:  () => emptyThrottle
+    };
+
+    let { layout, label, locale, browserLocale } = props;
+
+    if (browserLocale.country !== COUNTRY.US) {
```
bluepnume: The `locale` prop is derived from browser locale. I don't think we need to check both separately. We can just rely on `locale`.

src/lib/throttle.js
```diff
@@ -132,3 +136,55 @@ export function getReturnToken() : ?string {
...
+        return emptyThrottle;
+    }
+
+    let localeString = locale;
+
+    if (typeof locale === 'object') {
```
bluepnume: Locale prop will always be a string. Should be no need for this check.

src/lib/throttle.js
```diff
@@ -132,3 +136,55 @@ export function getReturnToken() : ?string {
...
+        localeString = `${ locale.lang }_${ locale.country }`;
+    }
+
+    if (localeString !== 'en_US') {
+        return emptyThrottle;
+    }
+
+    if (label !== 'checkout' && label !== 'paypal' && label !== 'pay' && label !== 'buynow') {
```
bluepnume: Can you use the `BUTTON_LABEL` const here?

src/resources/fundingLogos/credit.jsx
```diff
@@ -42,7 +42,11 @@ export function creditLogo({ logoColor, locale } : { logoColor : $Values<typeof
             </svg>
         );
     default:
-        return (
+        return isFundingThrottleEnabled ? (
+            <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 53 11">
```
bluepnume: Can you please specify width and height? We've seen SVGs act funky on older browsers without this.

src/button/component.jsx
```diff
@@ -187,6 +192,14 @@ export let Button : Component<ButtonOptions> = create({
             queryParam: true
         },
 
+        isFundingThrottleEnabled: {
+            type:     'boolean',
+            required: false,
+            def() : boolean {
+                return false;
```
bluepnume: Can we set up the throttle here? `def()` will be passed props, so you should be able to do

```
def(props) {
    fundingLogoThrottle = buildFundingLogoThrottle(props);
    return fundingLogoThrottle && fundingLogoThrottle.isEnabled();
}
```

This is better than overwriting `props.isFundingThrottleEnabled = true;` and means that the correct value will be sent to the server, which we'll need to do for the server-side portion of the render to work correctly.

src/button/component.jsx
```diff
@@ -187,6 +192,14 @@ export let Button : Component<ButtonOptions> = create({
             queryParam: true
         },
 
+        isFundingThrottleEnabled: {
+            type:     'boolean',
```
bluepnume: Can we add `queryParam: true` -- since the server-side will need to know to render this experience too.

src/button/component.jsx
```diff
@@ -721,6 +751,13 @@ export let Button : Component<ButtonOptions> = create({
                             [FPTI.KEY.BUTTON_SESSION_UID]: this.props.buttonSessionID
                         });
                     }
+                    
+                    if (fundingLogoThrottle.isActive()) {
```
bluepnume: Would also recommend adding a `logComplete()` in onAuthorize.

src/button/template/componentTemplate.jsx
```diff
@@ -25,7 +25,11 @@ function getCommonButtonClasses({ layout, shape, branding, multiple, env }) : st
     ].join(' ');
 }
 
-function getButtonClasses({ label, color, logoColor }) : string {
+function getButtonClasses({ label, color, logoColor, isFundingThrottleEnabled }) : string {
+    if (isFundingThrottleEnabled && label === 'credit') {
+        color = 'silver';
```
bluepnume: Can we use `BUTTON_COLOR.SILVER`

src/button/template/componentTemplate.jsx
```diff
@@ -120,14 +124,26 @@ function renderContent(text : string, { label, locale, color, branding, logoColo
                 return;
             }
 
-            if (!logoColor) {
+            if (!logoColor && !isFundingThrottleEnabled) {
                 throw new Error(`Can not determine logo without logo color`);
             }
+
+            if (isFundingThrottleEnabled && layout !== 'vertical') {
+                if (label === 'credit' && (name === 'paypal' || name === 'pp')) {
```
bluepnume: Can we use `BUTTON_LABEL`

src/button/template/componentTemplate.jsx
```diff
@@ -120,14 +124,26 @@ function renderContent(text : string, { label, locale, color, branding, logoColo
...
+                    return;
+                }
+
+                if (name === 'credit') {
```
bluepnume: Same here -- would be good to use constants.

bluepnume:
Added a few comments otherwise looks good. We should also sync up on changes to xo-buttonjs needed to pass the prop for server-side rendering. That should be straightforward.

src/lib/throttle.js
YSNIS: I thought so too, but as I was poking around I saw that locale was passing in what 'locale' was set in the script by the user.  We need to check both what the user is setting the locale as, and what the browser is set at.

src/lib/throttle.js
YSNIS: Yes I can.

src/button/component.jsx
YSNIS: Done.

src/button/template/componentTemplate.jsx
YSNIS: Done.

src/button/template/componentTemplate.jsx
YSNIS: Done

src/button/template/componentTemplate.jsx
YSNIS: Done.

src/button/component.jsx
YSNIS: Looks like the funding decorate function is called first, so I'm setting up the throttle in that block, but then using it to define the isFundingThrottleEnabled in the def() above.

src/lib/throttle.js
YSNIS: Done.

src/lib/throttle.js
YSNIS: When you set `locale: 'en_US'` it actually shows up here as an object.  If `locale` is not set it defaults to the string `'en_US'`.

src/resources/fundingLogos/credit.jsx
YSNIS: Done, though the width is 964px and the height is 200px.  Just the FYI.

src/lib/throttle.js
bluepnume: Understood -- but if a locale is set by the merchant, checkout.js disregards the browser locale entirely, so there's a single source of truth. If we later enable this button universally for buyers in the US, that logic will preside. So it would be good to make the experiment match that.

src/lib/throttle.js
bluepnume: Ah hmm, weird, that sounds like a quirk we need to fix on our side. Thanks for the heads-up. For now let's leave this check in that case.

src/resources/fundingLogos/credit.jsx
bluepnume: OK. I guess the main thing is, if you can just make sure it renders without any problems in IE11 we should be ok.

src/button/component.jsx
bluepnume: Hmm this is going to be problematic... lemme circle back on this one.

src/config/config.js
```diff
@@ -228,6 +228,297 @@ export let config = {
...
+            'stage.crazyhotseeds.com',
+            'ca.bathmatedirect.com',
+            'fi.bathmatedirect.com',
+            'au.bathmatedirect.com',
+            'bathmatedirect.com',
+            'eu.bathmatedirect.com',
+            'nl.bathmatedirect.com',
+            'il.bathmatedirect.com',
+            'int.bathmatedirect.com',
+            'www.bathmatedirect.com',
```
bluepnume: If we're stripping out the `www.` in the url in `buildFundingLogoThrottle`, we should do the same here right? Otherwise many of these urls won't match.

src/lib/throttle.js
```diff
@@ -132,3 +136,40 @@ export function getReturnToken() : ?string {
...
+        return null;
+    }
+
+    let domain = getDomain().replace(/^https?:\/\//, '').replace(/^www\./, '');
+    if (config.bmlCreditTest.domains.indexOf(domain) === -1) {
+        return null;
+    }
+
+    if (layout === undefined || (layout && layout === BUTTON_LAYOUT.HORIZONTAL)) {
+        return {
```
bluepnume: Can probably just directly return `getThrottle('ppc_rebrand', 50)` here right?

YSNIS: Oh yup!  That was a leftover from adding in the isActive flag to the object.

src/button/component.jsx
```diff
@@ -187,6 +188,15 @@ export let Button : Component<ButtonOptions> = create({
             queryParam: true
         },
 
+        isFundingThrottleEnabled: {
+            type:       'boolean',
+            queryParam: true,
+            required:   false,
+            def() : boolean {
+                return fundingLogoThrottle && fundingLogoThrottle.isEnabled() ? true : false;
```
bluepnume: So -- given that the order that `def` and `decorate` is called is a little non-deterministic, let's do this. In both places let's do:

```javascript
fundingLogoThrottle = fundingLogoThrottle || buildFundingLogoThrottle(normalizeProps(props));
```

-- then we can be sure that the throttle will exist when we need it in both places.

src/config/config.js
YSNIS: Yup.  Taking out the `www.`.

src/button/component.jsx
```diff
@@ -425,6 +435,15 @@ export let Button : Component<ButtonOptions> = create({
                 return {};
             },
             decorate({ allowed = [], disallowed = [] } : Object = {}, props : ButtonOptions) : {} {
+                
+                fundingLogoThrottle = buildFundingLogoThrottle({ ...normalizeProps(props), browserLocale: getBrowserLocale() });
+                if (fundingLogoThrottle) {
+                    allowed = [ ...allowed, FUNDING.CREDIT ];
+                    const creditIndex = disallowed.indexOf(FUNDING.CREDIT);
```
bluepnume: Lot of iteration going on here. Can we just do `disallowed = disallowed.filter(source => (source !== FUNDING.CREDIT))`?

bluepnume: Actually, thinking about it -- if a merchant has already opted-out of credit, do we really want to disregard that opt-out?

src/lib/throttle.js
YSNIS: If locale is set by the merchant, for instance `locale: 'en_JP'`, browserLocale is still coming back as `{country: "US", lang: "en"}`.  The rest of checkout might disregard it, but the throttle seems to still be differentiating even when locale is set.

src/button/component.jsx
YSNIS: Sure thing.

src/lib/throttle.js
bluepnume: Right -- `getBrowserLocale()` will always return the browser locale. And it may be different from the locale set by the merchant.

But the intention of the merchant set locale is to override the browser locale. All of the funding eligibility, content determination, even the locale which is passed to the hermes checkout flow, will be set to the locale set by the merchant. In every other scenario we normalize to a single locale. This experiment probably should not be an exception.

src/button/component.jsx
YSNIS: Yup/

src/resources/fundingLogos/credit.jsx
YSNIS: Tested rendering the SVG on IE11 with no issues.

PR #852, opened 2018-09-07
feat(callback-api): moving shipping options patch into payment patch

mcntsh:


bluepnume:
Looks good to me; coupla things:

- Can we get a coupla tests in here?
- Can you document a few example actions.payment.patch() calls and run them by `Kushal Sheth` to make sure they'll work when he builds out support in the actual api?

Thanks!

PR #855, opened 2018-09-12
fix(button outline): adding button outline to individual card icons. removing from transparent

mcntsh:


mcntsh:
@bluepnume I'm also working on changes to only show outlines when tab is pressed. We can wait for those changes to merge if you think that's appropriate.

src/button/template/componentScript.js
```diff
@@ -131,7 +131,32 @@ export function getComponentScript() : () => void {
             }
         }
 
+        function setupTabOutlineEvent() {
+            const buttonsContainer = document.getElementsByClassName('paypal-button-container')[0];
```
bluepnume: Can you use the `CLASS` const for `'paypal-button-container'`

src/button/template/componentScript.js
```diff
@@ -131,7 +131,32 @@ export function getComponentScript() : () => void {
...
+            function handleFirstTab(e) {
+                if (e.keyCode === tabKeyCode) {
+                    buttonsContainer.classList.add('{ CLASS.SHOULD_FOCUS }');
+
+                    window.removeEventListener('keydown', handleFirstTab);
+                    window.addEventListener('mousedown', handleMouseDownOnce);
+                }
+            }
+
+            buttonsContainer.classList.add('{ CLASS.SHOULD_FOCUS }');
```
bluepnume: Can we do this when we build the actual jsx for this template?

PR #871, opened 2018-09-21
Feat: update card icons

khanghoang:
Update latest credit card icons and screenshot tests


khanghoang:
@bluepnume can we discuss this PR again?

PR #880, opened 2018-10-03
Updated package.json `repository.url`

dlukeomalley:
Updated package.json `repository.url` to point to github.com, rather than github.paypal.com. This should fix the broken link from the [paypal-checkout npmjs.com](https://www.npmjs.com/package/paypal-checkout) page.


bluepnume:
Thanks a mil!

PR #895, opened 2018-10-16
add min confidence score for vpns epm

vishakha94:


src/button/component.jsx
```diff
@@ -187,6 +187,15 @@ export let Button : Component<ButtonOptions> = create({
             queryParam: true
         },
 
+        minConfidenceScore: {
```
bluepnume: Can we just put this in xo-buttonjs?

PR #900, opened 2018-10-20
Exposing card interface

karthik972:


bluepnume:
Going to expose this as a separate module.

PR #919, opened 2018-10-25
fix(error-focus-closed): Catching the error from the promise

mcntsh:
`try { ... } catch ...` doesn't work for these cross-domain functions. Doh.


mcntsh:
In zoid, `actions.close` is explicitly defined as a `ZalgoPromise<void>`. I could do a check just in case, though.

mcntsh:
@bluepnume I ended up using `ZalgoPromise.try` since that catches thrown errors and rejected promises 👍

PR #943, opened 2018-11-09
Add commitizen packages for commit message convention

khanghoang:
From now on, run `npm run commit` will prompt you to make a commit message that follows [Angular's commit message convention](https://gist.github.com/stephenparish/9941e89d80e2bc58a153) 🎉


khanghoang:
@mcntsh but I added 2 deps. Do we **not** commit `package-lock.json`?

PR #950, opened 2018-11-16
chore(venmo-blacklist): Blacklisting barkbox from venmo

chanson:
Hello, we have recently started using this library to support our PayPal payments for BarkBox. Because we are a subscription-based service using recurring payments/billing agreements, Venmo is not an option for us. Could you please have barkbox.com added to the blacklist? Thank you!


mcntsh:
No problem! Thanks for making your own PR :)

PR #969, opened 2018-12-17
Electron: more robust way of detection

AleksMeshkov:
https://github.com/paypal/paypal-checkout/issues/323#issuecomment-447604784

### DIFF
--- a/src/lib/device.js
+++ b/src/lib/device.js
@@ -129,7 +129,7 @@ export function isIECompHeader() : boolean {
 export function isElectron() : boolean {
     let userAgent = getUserAgent();
     // here we want a case-insensitive full word boundary
-    return /\belectron\b/i.test(userAgent);
+    return (/\belectron\b/i).test(userAgent);
 }
 
 export function isIEIntranet() : boolean {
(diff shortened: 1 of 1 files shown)


src/lib/device.js
```diff
@@ -127,10 +127,9 @@ export function isIECompHeader() : boolean {
 }
 
 export function isElectron() : boolean {
-    if (typeof process !== 'undefined' && process.versions && process.versions.electron) {
-        return true;
-    }
-    return false;
+    let userAgent = getUserAgent();
```
bluepnume: Can we also keep the `process.versions.electron` check, for backwards-compat? We can do an `||` on that and the user-agent.

AleksMeshkov: Actually I think the previous check was not relevant at all so I'm not sure it can provide any backwards-compatibility. Regarding other versions, I've just quickly looked through the aggregated data grouped by different electron versions (https://developers.whatismybrowser.com/useragents/explore/software_name/electron-application/1) and came to conclusion that the current check will pass in every case.

PR #972, opened 2018-12-19
Feat/inline guest responsive size

khanghoang:


src/button/template/componentStyle/button.js
```diff
@@ -122,6 +122,18 @@ export let buttonStyle = `
         visibility: hidden;
     }
 
+    @media only screen and (max-width: 249px) {
```
bluepnume: We can remove this and just display it all the time.

src/card/index.js
```diff
@@ -40,6 +40,11 @@ export const Card : Component<CardOptions> = create({
         popup:  false
     },
 
+    dimensions: {
+        height: '100%',
```
bluepnume: Can we set height to a sensible default? Like 300px?

khanghoang:
@bluepnume I updated the PR

PR #1007, opened 2019-02-04
Guard against falsy render param value for 'funding.allowed'

dechov:
Fixes https://github.com/paypal/paypal-checkout/issues/1006

Guards against falsy values (e.g. `null`) passed as the `funding.allowed` parameter in the Button render code (similar to how `allowed` is checked for 16 lines above), to avoid breaking the render.

### DIFF
--- a/src/button/component.jsx
+++ b/src/button/component.jsx
@@ -471,7 +471,7 @@ export let Button : Component<ButtonOptions> = create({
 
                 let apmFunding = APM_FUNDING.filter(source => (isApmEligible(source, props)));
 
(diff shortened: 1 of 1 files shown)


mcntsh:
Thanks @dechov for your PR. This bug is a P1 impacting several merchants so we went ahead and filed our own which addresses a couple of other use-cases: 

1. All non-array values (not just falsy)
2. Disallowed list as well as Allowed.

We really appreciate your report and your quick PR. This fix will be out in the next hour.

PR #1015, opened 2019-02-08
fix(whitelist-apm): Fix apm domain name matching issue

amyng:


mcntsh:
Hey @amyng, my one nitpick is that `String.endsWith` isn't supported in IE 11 AFAIK. Could you add a little helper function to `src/lib/util`? 

It could look something like this:

```js
export function endsWith(str : string, substr : string) : boolean {
    return str.substring(str.length - substr.length, str.length) === substr;
}
```

bluepnume:
It would also be good to split out the domain check into a memoized function, so we're not running the list for every funding source.

bluepnume:
Also just noticed .find is going to be a problem too in IE11 -- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

`.some` should be ok though.

amyng:
1. endsWith may not be a sufficient domain check.  We don't want to allow happy.com when unhappy.com is the whitelisted domain for instance.
2. suppressed security/detect-non-literal-regexp eslint.  There's no security risk since apmTestDomains is controlled in config.

src/button/component.jsx
```diff
@@ -77,21 +77,29 @@ function isCreditDualEligible(props) : boolean {
...
 
-function isApmEligible(source, props) : boolean {
+let isDomainAllowed = memoize(() : boolean => {
 
-    let { locale } = normalizeProps(props, { locale: getBrowserLocale() });
+    let domain = getDomain().replace(/^https?:\/\//, '').replace(/^www\./, '');
 
-    if (getFundingConfig(source, 'allowedCountries', [ locale.country ]).indexOf(locale.country) === -1) {
+    if (!config.apmTestDomains.some(allowDomain => {
+        let regex = new RegExp(`[^a-zA-Z\\d\\-]*${ allowDomain }$`);  // eslint-disable-line security/detect-non-literal-regexp
```
bluepnume: This is problematic. The `.` in the domain will be interpreted as "any character" when inserted into the regex.

amyng: Thanks for pointing it out!  Replaced with escaped period.

src/button/component.jsx
```diff
@@ -77,21 +77,29 @@ function isCreditDualEligible(props) : boolean {
...
 
-function isApmEligible(source, props) : boolean {
+let isDomainAllowed = memoize(() : boolean => {
 
-    let { locale } = normalizeProps(props, { locale: getBrowserLocale() });
+    let domain = getDomain().replace(/^https?:\/\//, '').replace(/^www\./, '');
 
-    if (getFundingConfig(source, 'allowedCountries', [ locale.country ]).indexOf(locale.country) === -1) {
+    if (!config.apmTestDomains.some(allowDomain => {
+        let regex = new RegExp(`[^a-zA-Z\\d\\-]*${ allowDomain.replace(/\./, '\\.') }$`);  // eslint-disable-line security/detect-non-literal-regexp
```
bluepnume: you need `/\./g` otherwise it'll only replace the first `.`

PR #1020, opened 2019-02-11
Add localhostUrl and checkoutUri to enable easier internal testing

bryclee:
Adding `localhostUrl` and `checkoutUri` props to get more control testing `checkout.js` flows.

### DIFF
--- a/src/checkout/component.js
+++ b/src/checkout/component.js
@@ -65,6 +65,10 @@ export let Checkout : Component<CheckoutPropsType> = create({
                 throw new Error(`Expected payment id or token to be passed, got ${ token }`);
             }
 
+            if (isPayPalDomain() && window.xprops && window.xprops.checkoutUri) {
+                return window.xprops.checkoutUri;
+            }
+
             return determineUrl(env, props.fundingSource, token);
         });
     },
@@ -75,8 +79,7 @@ export let Checkout : Component<CheckoutPropsType> = create({
 
     get domain() : Object {
         return {
-            ...config.paypalDomains,
-            [ ENV.LOCAL ]: /^http:\/\/localhost.paypal.com:\d+$/
+            ...config.paypalDomains
         };
     },
 
--- a/src/checkout/component.js
+++ b/src/checkout/component.js
@@ -199,10 +199,10 @@ export let Checkout : Component<CheckoutPropsType> = create({
             required:   false,
             queryParam: true,
 
-            def(props) : ?string {
-                const env = props.env || config.env;
-
-                return config.checkoutUris[env];
+            def() : ?string {
+                if (config.checkoutUri) {
+                    return config.checkoutUri;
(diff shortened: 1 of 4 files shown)


src/checkout/component.js
```diff
@@ -65,6 +65,10 @@ export let Checkout : Component<CheckoutPropsType> = create({
                 throw new Error(`Expected payment id or token to be passed, got ${ token }`);
             }
 
+            if (isPayPalDomain() && window.xprops && window.xprops.checkoutUri) {
+                return window.xprops.checkoutUri;
```
mcntsh: Should we also be pre-pending the `config.checkoutUrls[props.env]` here?

bryclee: Hm... so my understanding is that `xprops` is passed from the button component props, so `window.xprops.checkoutUri` is the already built URL. Are there cases where that's not true?

src/config/config.js
```diff
@@ -424,7 +424,7 @@ export let config = {
 
     get paypalDomains() : Object {
         return {
-            [ ENV.LOCAL ]:      `http://localhost.paypal.com:${ config.ports.default }`,
+            [ ENV.LOCAL ]:      /^https?:\/\/.*\.paypal\.com:?\d*$/,
```
bryclee: Making this a looser check - sometimes while testing locally, you might need to switch between stage pages and localhost pages.

bluepnume:
So checkout-uri prop makes sense here. But I'm not sure about adding localhosturl as a prop for the buttons. What do you think about this plan:

- Add a `data-local-url` param to the script tag (to match `data-stage-url`)
- Then let's set up xo-buttonjs to be able to proxy to different local servers on different ports (like we do [here](https://github.paypal.com/Checkout/clientsdknodeweb/blob/master/package.json#L23-L31))

That way the button and the checkout experience can run on the same domain/port (which is required for a few use-cases like passing access token), and we're also closer to the production environment.

bryclee:
Sure, sounds good. Just to understand, what kind of concerns are there over using the data attribute vs passing it as a prop?

bluepnume:
data-attribute would mean it can be universally used for any component rendered by checkout.js, rather than being limited to just the button. In fact, I'm tempted to say we should do the same for the checkout-uri since technically checkout can be rendered without the button being present

bryclee:
Updated to use data-attributes to control.

Ran the tests locally to verify passing.

bryclee:
Can I get a re-review for this? Let me know if there are any concerns.

bluepnume:
Re-running travis. Will merge once it's green, looks like some lint errors getting in the way.

bluepnume:
Could you rebase this pr? made a bunch of test fixes, hoping things are stable by now 👍

tophernsc:
What is this?

Chris

PR #1022, opened 2019-02-12
feat(pptm-clientid): Add client_id to pptm.js call

joshbeam:
# Background
Currently checkout.js sends the URL or mrid to pull down pptm.js. However, as part of ramping for some new MUSE features, we want to increase coverage by using client_id if it's available.

# Problem
client_id is only available during button render.

# Solution
Retain current behavior of how pptm.js is pulled down, but if the pptm.js bundle is empty, reload it during button render if a client_id is available (for those merchants who do a client-side integration instead of serverside).

**Design**

We do this by setting up a window function (`window. __pptmLoadedWithNoContent`) in the button component that will be called when pptm.js is pulled down in its `onload` handler. The handler will check if pptm.js is empty (by checking if an internal pptm event called `snippetRun` was pushed to the `paypalDDL` event queue), and if so, it will remove the empty pptm.js bundle and try to pull it down again, but this time by client_id (only if the merchant supplied a client_id during button render). If pptm.js is pulled down, then all good. If it's empty again, then nothing more happens. Note: due to the async nature of button render vs when pptm.js is loaded, there was some complexity with how the window listener is setup, but most of the complexity is abstracted away into the `pptm.reloadPptmScript` method. This is to ensure that timing of button render and script loading behaves the same regardless of the order they're called.

### DIFF
--- a/src/external/pptm-factory.js
+++ b/src/external/pptm-factory.js
@@ -1,0 +1,164 @@
+/* @flow */
+
+import { track, info } from 'beaver-logger/client';
+
+import { config } from '../config';
+import { FPTI, PPTM_ID } from '../constants';
+import { stringifyError, extendUrl, loadScript, getElement, isPayPalDomain } from '../lib';
+
+
+function shouldCreateInitialPptmScript() : boolean {
+    const id = window.location.hostname;
+
+    if (!id) {
+        return false;
+    }
+
+    if (isPayPalDomain()) {
+        return false;
+    }
+
+    const existingScript = getElement(PPTM_ID);
+    const alreadyDownloaded = Boolean(existingScript);
+
+    if (alreadyDownloaded) {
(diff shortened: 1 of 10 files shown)


src/external/pptm-factory.js
```diff
@@ -0,0 +1,164 @@
...
+                delete params.client_id;
+            }
+        
+            const fullUrl = extendUrl(config.pptmUrl, params);
+        
+            loadScript(fullUrl, 0, {
+                async:  true,
+                id:     PPTM_ID,
+                onload: `(function() {
+                    if (window.paypalDDL && window.paypalDDL[0] && window.paypalDDL[0].event === 'snippetRun') {
```
bluepnume: Can we run this in the `.then()`? That should be invoked on script load.

bluepnume:
Looks good to me! Just one minor comment.

PR #1041, opened 2019-02-28
[DO NOT MERGE]chore(config): change the config for inline guest

khanghoang:


bluepnume:
Leeeet's not do it all at once given how many requests there are!! :D We should start with 1% traffic and go from there.

bluepnume:
Probably a `Math.random() < 0.01` style throttle.

khanghoang:
You're right about that. I created it PR just as an example of how we can get rid of Hermes on local development.

PR #1062, opened 2019-03-18
tests(popup-bridge): Increasing delay on button click for bridge tests.

mcntsh:
+ simplifying travis config, for consistency in how it runs karma


webpack-dev-server.config.js
```diff
@@ -1,11 +1,13 @@
 /* @flow */
 /* eslint unicorn/filename-case: 0 */
-import { BASE, getWebpackConfig } from './webpack.config';
```
bluepnume: Did you mean to change this file too?

karma.conf.js
```diff
@@ -42,6 +39,13 @@ export default function configKarma(karma : Object) {
         ...karmaConfig.files
     ];
 
+    karmaConfig.customLaunchers = {
+        Chrome_travis_ci: {
```
bluepnume: Was one of the flags passed to chrome in grumbler-scripts causing the problem?

.travis.yml
```diff
@@ -16,9 +16,4 @@ script:
 sudo: required
 dist: trusty
 addons:
-  firefox: "52.0"
-  apt:
-    sources:
-      - google-chrome
-    packages:
-      - google-chrome-stable
+  chrome: stable
```
bluepnume: 👍

karma.conf.js
```diff
@@ -4,9 +4,6 @@ import { getKarmaConfig } from 'grumbler-scripts/config/karma.conf';
 
 import { TEST } from './webpack.config';
 
-// eslint-disable-next-line no-process-env
-process.env.CHROME_BIN = require('puppeteer').executablePath();
```
bluepnume: Hmm weird, thought this was needed to fix something else. but if it passes then sounds good 🤷‍♂️

webpack-dev-server.config.js
mcntsh: Yeah, fixing linter errors to get the build to pass. Was originally in its own PR but, cant get the build to pass without these other changes :)

karma.conf.js
mcntsh: No, I think the problem was the version of Chrome installed by Aptitude. This launcher config just ensures we use Chrome headless for Travis for consistency sake.

karma.conf.js
mcntsh: I think it was intended to fix the original problem by using Puppeteer's version of Chromium rather than the system's version, but in this case we have more control over the system's version so it makes sense to remove this line.

webpack-dev-server.config.js
bluepnume: Aha I had some fixes locally for this -- I've pushed them up

karma.conf.js
bluepnume: ok gotcha. then I'd say let's stick to grumbler-script's chrome definition if it's not specifically causing any problems. then we get the benefit of any flags we want to pass through the command line (like `--dev-tools`)

karma.conf.js
bluepnume: 👍 works for me

webpack-dev-server.config.js
mcntsh: Got it, I just removed those changed from this branch

PR #1065, opened 2019-03-20
US1271427 add whitelist merchant check for rendering apm buttons

Shaaaaman:
This PR is basically enabling the APMs buttons for the merchants who is in the whitelist regardless commit true or false.

Rally ticket:
https://rally1.rallydev.com/#/11744705241d/detail/userstory/289801231332

### DIFF
--- a/src/funding/funding.js
+++ b/src/funding/funding.js
@@ -1,10 +1,7 @@
 /* @flow */
 
-import { getDomain } from 'cross-domain-utils/src';
-
 import { ENV, CARD_PRIORITY, FUNDING, BUTTON_LAYOUT, FUNDING_ELIGIBILITY_REASON } from '../constants';
 import type { LocaleType, FundingSource, FundingSelection, FundingList } from '../types';
-import { config } from '../config';
 
 import { getFundingConfig, getCardConfig, FUNDING_PRIORITY, FUNDING_CONFIG } from './config';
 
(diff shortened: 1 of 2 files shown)


src/funding/funding.js
```diff
@@ -29,7 +33,7 @@ export function isFundingIneligible(source : FundingSource, { locale, funding, l
         return FUNDING_ELIGIBILITY_REASON.DISALLOWED_COUNTRY;
     }
 
-    if (getFundingConfig(source, 'requireCommitAsTrue') && !commit) {
+    if ((getFundingConfig(source, 'requireCommitAsTrue') && !commit) || config.apmTestDomains.indexOf(apmDomain) !== -1) {
```
bluepnume: I'm not sure this boolean logic is correct -- it should be an `&&`. Can we do this:

```javascript
if ((getFundingConfig(source, 'requireCommitAsTrue') && !commit) {
    if (config.apmTestDomains.indexOf(apmDomain) !== -1)) {
        return FUNDING_ELIGIBILITY_REASON.COMMIT_NOT_SET;
    }
}
```

Can we also hoist the domain check outside of this function? Otherwise we're looping that array for each individual funding source.

Shaaaaman: So the current logic is if the merchant is whitelisted, the APMs button should be rendered regardless commit passed as true or false

bluepnume: In which case, this should be `&& !isWhitelistedMerchant`.

Shaaaaman:
retest

bluepnume:
Looks good! Will let it go green then merge.

PR #1078, opened 2019-04-10
Add disableFunding, disableCard, merchantID as button props

vishakha94:


src/buttons/component.jsx
```diff
@@ -506,6 +506,24 @@ export function getButtonsComponent() : ZoidComponent<ButtonProps> {
...
+                },
+                
+                disableCard: {
+                    type:       'array',
+                    queryParam: true,
+                    value:      () => getDisableCard()
+                },
+                
+                merchantID: {
+                    type:       'string',
```
bluepnume: We should make this an array too, right? For future proofing.

PR #1101, opened 2019-05-10
feat(muse-spb): send an event to paypalDDL when button is rendered

sy2566:
Muse component will trigger analytics-xo tag once SPB button is rendered.
This PR is for sending the `paypalButtonRender` event to muse data layer.

### DIFF
--- a/src/external/pptm-factory.js
+++ b/src/external/pptm-factory.js
@@ -78,7 +78,8 @@ export function pptmFactory() : Object {
         },
         listenForButtonRender() {
             window.paypalDDL = window.paypalDDL || [];
-            if (!window.paypalDDL.find(e => e.event === 'paypalButtonRender')) {
+            const buttonRenderEvent = window.paypalDDL.filter(e => e.event === 'paypalButtonRender');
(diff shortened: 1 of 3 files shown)


src/external/pptm-factory.js
```diff
@@ -76,6 +76,12 @@ export function pptmFactory() : Object {
                 }
             };
         },
+        listenForButtonRender() {
+            window.paypalDDL = window.paypalDDL || [];
+            if (!window.paypalDDL.find(e => e.event === 'paypalButtonRender')) {
+                window.paypalDDL.push({ event: 'paypalButtonRender' });
```
joshbeam: [Find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) isn't compatible with IE, and I'm not sure if webpack will take care of this and make it compatible. Maybe you can see if it will automatically transform it to `filter`, or maybe you can just use `filter` instead.

sy2566: @joshbeam Good point. I update to use filter method. Thanks!

PR #1136, opened 2019-06-26
Add log if the inline guest is rendered but outside of the viewport

khanghoang:
Add log if the inline guest is rendered but outside of the viewport

### DIFF
--- a/src/button/component.jsx
+++ b/src/button/component.jsx
@@ -5,10 +5,10 @@
 import { ZalgoPromise } from 'zalgo-promise/src';
 import { create } from 'zoid/src';
 import { type Component } from 'zoid/src/component/component';
-import { info, warn, track, error, flush as flushLogs } from 'beaver-logger/client';
+import { info, warn, track, error, flush as flushLogs, immediateFlush } from 'beaver-logger/client';
 import { getDomain } from 'cross-domain-utils/src';
 import { base64encode } from 'belter/src';
-import { debounce } from 'zoid/src/lib';
+import { debounce, once } from 'zoid/src/lib';
 
 import { pptm } from '../external';
 import { config } from '../config';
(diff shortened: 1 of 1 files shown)


khanghoang:
@mcntsh I put the code to add the null check

src/button/component.jsx
```diff
@@ -892,17 +892,59 @@ export const Button : Component<ButtonOptions> = create({
...
+                        }
+
+                        if (document.documentElement) {
+                            return document.documentElement.scrollTop;
+                        }
+
+                        return 0;
+                    };
+
+                    const checkIfInlineGuestInViewport = () => {
```
bluepnume: Why do we care about the viewport? Surely the main concern is if the buttons are overflowing outside the parent element, no?

src/button/component.jsx
```diff
@@ -892,17 +892,59 @@ export const Button : Component<ButtonOptions> = create({
             type:     'function',
             required: false,
             get value() : Function {
+                let initialHeight;
+                const logInlineGuestOutOfViewPortOnlyOnce = once((data) => {
+                    info(`inline_guest_outside_viewport`, data);
```
bluepnume: Can we make this event name generic? This is something which could be an issue for any integration with a variable vertical height.

src/button/component.jsx
khanghoang: Actually, I think they're 2 different problems. 

This is the log for https://www.onbuy.com use case, this is their checkout page with SPBs at the bottom of the page.

Checkout 2019-06-26 23-28-38

This is how it looks like **after** the buyers click on the credit card icons
Checkout 2019-06-26 23-31-41

The problem is if the buyers don't scroll, they may not know that there is the inline guest form, and that is the case we want to log.

src/button/component.jsx
khanghoang: good point, I updated the code to address your point.

src/button/component.jsx
```diff
@@ -892,17 +892,60 @@ export const Button : Component<ButtonOptions> = create({
...
+
+                        if (document.documentElement) {
+                            return document.documentElement.scrollTop;
+                        }
+
+                        return 0;
+                    };
+
+                    // explanation https://github.com/paypal/paypal-checkout-components/pull/1136#discussion_r298025574
+                    const checkIfExpansionInViewport = () => {
```
mcntsh: Could this, and `getScrollOffsetY` be abstracted, and put into `lib/dom.js`? That way we can keep this wrapper thin and we can re-use the logic for any other components that need it.

khanghoang: I will put the `getScrollOffsetY` into `lib/dom.js` and will refactor the `checkIfExpansionInViewport` later since the latter function needs some variables of the current scope and little hard to decouple those.

PR #1171, opened 2019-07-30
fix APMs commit true and false logic

Shaaaaman:


bluepnume:
Can you please add this logic the original way, before it was removed, where each funding source could be configured individually to allow or disallow commit.

Shaaaaman:
@bluepnume If the original way is putting back requireCommitTrue for each funding source then it may not the way we want to show APM buttons. Because here is one case where we still want to show button for domain whitelisted merchant even if commit is false. However if we are using requireCommitTrue, we are not able to show APM buttons if commit is false.

Shaaaaman:
retest

PR #1181, opened 2019-08-12
feat(*): add onKeyPress event to BasicButton and VaultedButton

mihoowk:
Add keyboardAccessibility property to ButtonProps and ButtonPropsInputs buttons. Add keyboardAccessibilityHandler to VaultedButton and BasicButton buttons which trigger clickHandler when keyboardAccessibility property include event.keyCode.

-


bluepnume:
I'm happy to include this change -- but can we avoid making this part of the external interface for the module? We should internally listen for enter and space keypresses and treat those as clicks automatically.

Can you also please avoid pushing changes to dist/ -- these will be autogenerated as part of the release. Thanks!

bluepnume:
Thanks!

PR #1206, opened 2019-09-16
fix commit true false issue for APMs

Shaaaaman:
JIRA: https://engineering.paypalcorp.com/jira/browse/PPPLMER-75560


bluepnume:
Hi -- I'd still like to use requireCommitAsTrue, or some other config based flag. Even if that can be overridden by the whitelist.

Shaaaaman:
@bluepnume Please review. thanks

PR #1261, opened 2019-11-21
suppressing loading dots for button and tagline, not consuming button…

vishakha94:
… text from mors

### DIFF
--- a/src/button/template/componentTemplate.jsx
+++ b/src/button/template/componentTemplate.jsx
@@ -3,21 +3,21 @@
 
 import { base64encode } from 'belter/src';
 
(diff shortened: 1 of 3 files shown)


src/button/template/componentTemplate.jsx
```diff
@@ -296,10 +299,13 @@ function renderTagline({ label, tagline, color, locale, multiple, env, cards, ch
         return;
     }
   
+    
     if (__WEB__ && layout !== BUTTON_LAYOUT.VERTICAL) {
```
bluepnume: let's just comment this entire block?

vishakha94: ok

vishakha94: no

PR #1268, opened 2019-12-10
button text personalization animation

vishakha94:


vishakha94:
screenshot update pending

PR #1270, opened 2019-12-12
show personalized button text only when tagline:false

vishakha94:


src/funding/paypal/template.jsx
```diff
@@ -49,8 +49,10 @@ function getButtonStyle(opts : LabelOptions) : ?ChildType {
     if (__TEST__) {
         return null;
     }
+    
+    const { tagline } = opts;
 
-    const personalizationText = getPersonalizationText(opts);
+    const personalizationText = !tagline && getPersonalizationText(opts);
```
bluepnume: Let's put this condition in `getPersonalizationText`

PR #1297, opened 2020-04-29
Add Vietnamese (vi) and Romanian (ro)

mimckenna:
Vietnamese and Romanian are now live in PayPal.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paypal/checkout-components",
-  "version": "5.0.86",
+  "version": "5.0.87",
```
bluepnume, line 3: Please back this change out. This will happen automatically at publish time. Thanks!

mimckenna: Version bump rolled back.

PR #1315, opened 2020-05-21
change the tag property for zoid buttons to paypal

vishakha94:


bluepnume:
Can't change the tag :) as you'll have noticed from the failed tests, this will break real integrations.

bluepnume:
Can we set it here instead: https://github.com/paypal/paypal-checkout-components/blob/master/src/zoid/buttons/component.jsx#L51-L56

vishakha94:
done

PR #1326, opened 2020-06-02
Fix links in changelog

gregjopa:
I noticed the links in the changelog point to the old private repository. This PR fixes the issue with two changes:
1. Updates package.json to use the public repository link. 
1. Uses the [conventional-changelog-cli](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-cli) tool to regenerate the changelog. This awesome tool uses the repository link defined in package.json for creating the commit hyperlinks in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,350 +9,350 @@
 
 ## <small>5.0.94 (2020-05-28)</small>
 
-* Add data-instrument-type to button ([3a2533a](http://github.paypal.com/paypal/paypal-checkout/commit/3a2533a))
-* Fix standalone tests ([8dc61e0](http://github.paypal.com/paypal/paypal-checkout/commit/8dc61e0))
-* Use preventClickFocus on buttons ([58e3f5c](http://github.paypal.com/paypal/paypal-checkout/commit/58e3f5c))
+* Add data-instrument-type to button ([3a2533a](https://github.com/paypal/paypal-checkout-components/commit/3a2533a))
+* Fix standalone tests ([8dc61e0](https://github.com/paypal/paypal-checkout-components/commit/8dc61e0))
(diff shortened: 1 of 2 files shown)


mstuart:
Awesome! 👍 

I'm not 100% familiar w/ the release script, but does conventional-changelog-cli need to be added to https://github.com/paypal/paypal-checkout-components/blob/4feda2dc411698f093dc0aecb5f384b580f49a2e/scripts/version.sh#L6 so the links will continue to be fixed w/ the right domain? Also, was conventional-changelog-cli responsible for fixing the links from github.paypal.com to github.com?

gregjopa:
Hey @mstuart. Great questions!

> Does conventional-changelog-cli need to be added to version.sh?

I don't think any changes are needed with the version.sh release script. That bash script already contains the following line which should take care of calling the conventional-changelog app: https://github.com/paypal/paypal-checkout-components/blob/4feda2dc411698f093dc0aecb5f384b580f49a2e/scripts/version.sh#L7 

The conventioal-changelog tool uses the package.json respository.url link to generate the hyperlinks in the changelog. When we do our next release it will use the correct url.

> Was conventional-changelog-cli responsible for fixing the links from github.paypal.com to github.com?

Yep I used it with the `--release-count=0` option to generate the whole changelog. Here's the full command I ran:
```bash
./node_modules/.bin/conventional-changelog -i CHANGELOG.md -s --release-count 0
```

The `chore(release)` commits did get included in the changelog when running the above script. I manually did a regex find/replace in vscode to get rid of them. I assume this was only a problem because I generated the whole changelog. I don't think it will be a problem when doing releases going forward but I'm not 100% sure.

bluepnume:
Can you rebase? Added a fix for tests to master.

gregjopa:
@bluepnume thanks for fixing the tests! I just rebased and the CI checks are passing 🎉

PR #1328, opened 2020-06-04
Add Flex/Pay Later button

Seavenly:


src/funding/funding.js
```diff
@@ -39,6 +39,10 @@ export function isFundingEligible(source : $Values<typeof FUNDING>, { layout, pl
         return false;
     }
 
+    if (source === FUNDING.CREDIT && fundingEligibility.paylater && fundingEligibility.paylater.eligible) {
```
bluepnume, line 42: This should be handled in graphql.

src/zoid/checkout/component.jsx
```diff
@@ -196,6 +196,18 @@ export function getCheckoutComponent() : ZoidComponent<CheckoutPropsType> {
                 test: {
                     type:    'object',
                     default: () => (window.__test__ || { action: 'checkout' })
+                },
+
+                creditProductIdentifier: {
```
bluepnume, line 201: Can you share some details for these two params?

bluepnume:
Bulk of changes look good. 👍 Few comments and questions

PR #1331, opened 2020-06-09
fix iframe title name, aria-label text

vishakha94:


src/lib/util.js
```diff
@@ -0,0 +1,5 @@
+/* @flow */
+
+export function capitalize(str : ?string) : ?string {
```
bluepnume, line 3: Please use https://github.com/krakenjs/belter/blob/master/src/util.js#L714

src/funding/credit/config.jsx
```diff
@@ -53,6 +56,6 @@ export function getCreditConfig() : FundingSourceConfig {
             [ BUTTON_COLOR.WHITE ]: LOGO_COLOR.BLUE
         },
         
-        labelText: `${ FUNDING.PAYPAL } ${ FUNDING.CREDIT }`
+        labelText: `${ FUNDING_BRAND_LABEL.PAYPAL } ${ capitalizedCreditName }`
```
bluepnume, line 59: Can we just put `PayPal Credit` in `FUNDING_BRAND_LABEL`?

src/zoid/buttons/component.jsx
```diff
@@ -52,7 +52,7 @@ export const getButtonsComponent = memoize(() : ZoidComponent<ButtonProps> => {
             iframe: {
                 allowpaymentrequest: 'allowpaymentrequest',
                 scrolling:           'no',
-                title:               'paypal'
+                title:               `${ FUNDING_BRAND_LABEL.PAYPAL }`
```
bluepnume, line 55: Don't need the ``${}``

vishakha94: ooops

PR #1366, opened 2020-07-15
[DO NOT MERGE] Button attribute for dropdown selector

amaclean2:


src/ui/buttons/menu.jsx
```diff
@@ -12,6 +12,9 @@ export function MenuButton({ color = TEXT_COLOR.BLACK } : {| color? : $Values<ty
     return (
         <Style css={ css }>
             <div
+                role="button"
+                // localize this, then it can be released
+                // aria-label="funding-options dropdown"
```
bluepnume, line 17: Looks good 👍 -- we should have this use `props.content` then pass in the content mapping from spb when we render this menu component.

amaclean2: Yeah, but I don't think I have time to add new content, so I just commented it out

bluepnume:
Is this PR still needed?

PR #1401, opened 2020-08-04
[DO NOT MERGE] Added enableFunding Logic

JianyiGao:


src/funding/funding.js
```diff
@@ -27,30 +27,30 @@ export function isFundingEligible(source : $Values<typeof FUNDING>, { layout, pl
...
+    if (layout && fundingConfig.layouts && fundingConfig.layouts.indexOf(layout) === -1 && !enableFunding.some(funding => funding === fundingSource)) {
         return false;
     }
 
     if (fundingConfig.platforms && fundingConfig.platforms.indexOf(platform) === -1) {
         return false;
     }
 
-    if (fundingConfig.shippingChange === false && onShippingChange) {
+    if (fundingConfig.shippingChange === false && onShippingChange && !enableFunding.some(funding => funding === fundingSource)) {
```
bluepnume, line 38: This should not be necessary -- even if merchant passes enable funding, shipping change will not be supported.

src/funding/funding.js
```diff
@@ -27,30 +27,30 @@ export function isFundingEligible(source : $Values<typeof FUNDING>, { layout, pl
         return false;
     }
 
-    if (layout && fundingConfig.layouts && fundingConfig.layouts.indexOf(layout) === -1) {
+    if (layout && fundingConfig.layouts && fundingConfig.layouts.indexOf(layout) === -1 && !enableFunding.some(funding => funding === fundingSource)) {
```
bluepnume, line 30: This should not be necessary -- the APMs will continue to be eligible in standalone and vertical modes right?

src/funding/funding.js
JianyiGao: Right, but we would like to show the buttons regardless even though apms will not support horizontal layout and shipping changes. This should be an integration error, since merchants called out they want this apm. We are planning to add some error message for these two use cases.

bluepnume: If it's an integration error, why even show the buttons?

src/zoid/checkout/component.jsx
```diff
@@ -154,6 +154,12 @@ export function getCheckoutComponent() : ZoidComponent<CheckoutPropsType> {
                     queryParam: true,
                     default:    () => FUNDING.PAYPAL
                 },
+
+                standaloneFundingSource: {
```
bluepnume, line 158: We already have `fundingSource` param which signifies a standalone button, we should rely on that.

src/funding/funding.js
bluepnume: Still don't think this is required. This logic will not apply to standalone.

bluepnume:
Hi, happy to merge in the changes to propagate `enableFunding` as a prop, but let's not add any more eligibility logic to this module unless absolutely necessary. Ideally this module should just respect the result from fundingEligibility unless it's a truly exceptional case.

src/zoid/checkout/component.jsx
JianyiGao: This is the flag to identify if we override some rules because of enablefunding param 
https://github.com/paypal/paypal-smart-payment-buttons/blob/897f0a7da2c59cf119bbc08c777fd77760cda956/src/button/props.js#L219

bluepnume: I don't think we need an additional prop for this, right? This can be derived any time it's needed.

src/zoid/checkout/component.jsx
```diff
@@ -154,6 +154,18 @@ export function getCheckoutComponent() : ZoidComponent<CheckoutPropsType> {
...
+                    type:       'string',
+                    queryParam: true,
+                    required:   false
+                },
+
+                enableFunding: {
```
bluepnume, line 164: I think we should set `value: getEnableFunding` here.

src/zoid/checkout/component.jsx
bluepnume: Agreed to remove this right?

src/funding/funding.js
```diff
@@ -10,9 +10,9 @@ import type { OnShippingChange } from '../ui/buttons/props';
 
 import { getFundingConfig } from './config';
 
-export function isFundingEligible(source : $Values<typeof FUNDING>, { layout, platform, fundingSource, fundingEligibility, components, onShippingChange, flow } :
+export function isFundingEligible(source : $Values<typeof FUNDING>, { layout, platform, fundingSource, fundingEligibility, components, onShippingChange, flow, enableFunding } :
```
bluepnume, line 13: There are a lot of points in this PR where we're passing through `enableFunding` but not using it anywhere. Can we clear those up?

PR #1403, opened 2020-08-07
update fields

andyfen:
updates to render paymentfields iframe

### DIFF
--- a/src/zoid/fields/component.jsx
+++ b/src/zoid/fields/component.jsx
@@ -22,7 +22,7 @@ export function getFieldsComponent() : ZoidComponent<FieldsProps> {
             autoResize: {
                 width:   false,
                 height:  true,
-                element: '#fields'
+                element: '#payment-fields'
             },
 
             dimensions: {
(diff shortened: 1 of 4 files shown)


src/zoid/fields/component.jsx
```diff
@@ -51,6 +51,30 @@ export function getFieldsComponent() : ZoidComponent<FieldsProps> {
...
+
+                values: {
+                    type:       'object',
+                    queryParam: true,
+                    required:   false,
+                    decorate:   ({ value }) => value,
+                    default: () => ({})
+                },
+
+                styles: {
```
bluepnume, line 70: Let's use `style` to match existing components.

src/zoid/fields/component.jsx
```diff
@@ -51,6 +51,30 @@ export function getFieldsComponent() : ZoidComponent<FieldsProps> {
...
             props: {
+
+                createOrder: {
+                    type:       'function',
+                    queryParam: 'token',
+                    alias:      'payment',
+                    queryValue: ({ value }) => ZalgoPromise.try(value)
+                },
+
+                values: {
```
bluepnume, line 62: Can we go with `fields`? `values` is a little generic.

src/zoid/fields/component.jsx
```diff
@@ -51,6 +51,30 @@ export function getFieldsComponent() : ZoidComponent<FieldsProps> {
             },
 
             props: {
+
+                createOrder: {
```
bluepnume, line 55: We agreed that SPB would be handling this right? 

- User clicks on SPB
- SPB communicates with fields (can pass in order id if needed)
- SPB then launches popup

bluepnume:
Looks good! Can you `npm run lint -- --fix` and I'll be happy to merge this in.

PR #1404, opened 2020-08-10
feature/pageType

mnicpt:
Adds `data-page-type` attribute for v4 checkout script and logs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "webpack-dev-server": "^3"
   },
   "dependencies": {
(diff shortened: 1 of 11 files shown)


mnicpt:
@gregjopa Yeah, was thinking about that but were no tests on scripts and this version isn't supposed to get new features so I tested using the enclosed dev runner by testing invalid and valid scenarios and inspecting the logger requests.  Those are working as expected.  Didn't want to put too much into this being v4.  I'm open to adding tests if Daniel and others from commonwealth team would prefer as well.

PR #1405, opened 2020-08-11
feature/dataVersion

mnicpt:
`data-paypal-smart-button-version` attribute for checkout.js to support Honey. This was tested using the dev runner and attribute shows up as expected.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,4 +1,4 @@
-registry=http://registry.npmjs.com/
+registry=https://registry.npmjs.org/
 package-lock=false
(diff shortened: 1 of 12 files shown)


.npmrc
```diff
@@ -1,4 +1,4 @@
-registry=http://registry.npmjs.com/
+registry=http://registry.npmjs.org/
```
gregjopa, line 1: Can we also make this https?

```suggestion
registry=https://registry.npmjs.org/
```

PR #1434, opened 2020-09-22
Vaulted installments middleware and component

chetanjk:
Adding component for installments

### DIFF
--- a/src/zoid/buttons/container.jsx
+++ b/src/zoid/buttons/container.jsx
@@ -11,15 +11,17 @@ import { BUTTON_SIZE_STYLE, MINIMUM_SIZE, MAXIMUM_SIZE } from '../../ui/buttons/
 import { type ButtonProps } from '../../ui/buttons/props';
 
 const CLASS = {
-    VISIBLE:         'visible',
-    INVISIBLE:       'invisible',
-    COMPONENT_FRAME: 'component-frame',
-    PRERENDER_FRAME: 'prerender-frame',
-    SMART_MENU:      'smart-menu'
+    VISIBLE:            'visible',
+    INVISIBLE:          'invisible',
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -78,7 +78,7 @@
         "sync-browser-mocks": "^2.0.8"
     },
     "dependencies": {
-        "@paypal/common-components": "^1.0.10",
+        "@paypal/common-components": "github:chetanjk/paypal-common-components#vaulted-installments",
```
bluepnume, line 81: Can you revert this line

src/zoid/buttons/container.jsx
```diff
@@ -141,6 +143,8 @@ export function containerTemplate({ uid, props, tag, context, frame, prerenderFr
             <node el={ prerenderFrame } />
 
             <div id={ ID.SMART_MENU } class={ CLASS.SMART_MENU } />
+
+            <div id={ ID.INSTALLMENTS_MODAL } class={ CLASS.INSTALLMENTS_MODAL } />
```
bluepnume, line 147: I think we can just render this modal up to `body` right? Without introducing a new container

bluepnume:
Looks good, few minor changes + will merge.

src/zoid/buttons/container.jsx
chetanjk: This actually helps in accessibility where keyboard tabbing moves immediately to installments popup after buttons. We can have a check before rendering if this container exists or else append to body

package.json
chetanjk: done

bluepnume:
Can you rebase against master to pick up test fix? Thanks!

chetanjk:
Rebased

PR #1461, opened 2020-10-16
Set button color to gold when style.color is an empty string

amyegan:
When button color is specified as an empty string, the default color was not being picked up. The button was rendered with no background color or border. 
``` javascript
paypal.Buttons({
    style: {
        color: ''
    },
    ...
```

This change detects the empty string and resets it to the default button color value before the button is rendered.


mstuart (approved):

gregjopa (approved):
Nice work @amyegan! It's awesome to see unit tests included 👍 💯 🥇

bluepnume:
Thanks for the PR @amyegan, this is awesome!

One quick thing -- can we do this defaulting in `normalizeButtonStyle`? That's where other defaults are set right now: https://github.com/paypal/paypal-checkout-components/blob/4e62f46cff2fa61f468b23b6a236ee463a86d116/src/ui/buttons/props.js#L292

amyegan:
Thanks @bluepnume . That makes sense. I moved the change into `normalizeButtonStyle`

bluepnume (approved):

mstuart:
lgtm!! 👍

PR #1471, opened 2020-11-11
Reorder SPB and move PayPal branding below SCF

allen-tong:
Changes:
* Move "Debit or Credit Card" button to the bottom of the SPB stack
* Move "Powered by PayPal" branding below inline guest card fields

Before:
Screen Shot 2020-11-11 at 1 35 52 PM

After:
Screen Shot 2020-11-11 at 1 36 36 PM

### DIFF
--- a/src/funding/funding.js
+++ b/src/funding/funding.js
@@ -1,8 +1,8 @@
 /* @flow */
 
+import { SUPPORTED_FUNDING_SOURCES } from '@paypal/funding-components/src';
 import type { FundingEligibilityType } from '@paypal/sdk-client/src';
 import { PLATFORM, FUNDING, COMPONENTS } from '@paypal/sdk-constants/src';
-import { values } from 'belter/src';
 
 import type { Wallet } from '../types';
 import { BUTTON_LAYOUT, BUTTON_FLOW } from '../constants';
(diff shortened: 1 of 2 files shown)


bluepnume:
Awesome, thanks!

allen-tong:
@bluepnume @mnicpt I've taken care of the ordering by using a different enum as suggested, but I didn't notice anywhere that specified or enforced that the order would stay this way. Would you suggest that I add additional logic here to enforce that?

mnicpt:
@allen-tong It looks like https://github.com/paypal/paypal-checkout-components/pull/1471/files#diff-3fba3feb650bae3f6bfc93d232bac1d8cfde682d5bfffbdd40d8cdfeda7fcd8a enforces it as far as the credit button.

allen-tong:
@mnicpt Ah that just moves the card fields above the "Powered by PayPal" branding, but I was concerned that for the buttons themselves, there's nothing explicitly enforcing that "Debit or Credit Card" should be at the bottom of the stack. (If it's not, then the buttons below it aren't hidden when you open inline guest, leading to the situation seen in the first screenshot where you can still see buttons for SEPA, giropay, Sofort)

mnicpt:
@allen-tong Gotcha. Will look in the morning.

mnicpt:
@allen-tong You're correct.  We still don't want to use the enum to dictate order so will need something else to enforce it.  When iterating over the fundingSources in `buttons.jsx`, we can check for CARDS and not render with the group initially.  Then we can conditionally render cards after that prior to the tagline to ensure that if the enum is updated and another funding source is added to the end, that it won't mess up this ordering.

PR #1473, opened 2020-11-16
Enforce ordering of card button

allen-tong:
Continuation of #1471 

@mnicpt It seems the button elements are labeled according to index in their class names, so I reordered the list itself. I also reverted the previous enum change.

Sorry for the multiple PRs!

### DIFF
--- a/src/ui/buttons/buttons.jsx
+++ b/src/ui/buttons/buttons.jsx
@@ -98,11 +98,15 @@ export function Buttons(props : ButtonsProps) : ElementNode {
         nonce, components, onShippingChange, personalization, userIDToken, content, flow, experiment } = normalizeButtonProps(props);
     const { layout, shape, tagline } = style;
 
-    const fundingSources = determineEligibleFunding({ fundingSource, layout, remembered, platform, fundingEligibility, components, onShippingChange, flow, wallet });
+    let fundingSources = determineEligibleFunding({ fundingSource, layout, remembered, platform, fundingEligibility, components, onShippingChange, flow, wallet });
     const multiple = fundingSources.length > 1;
 
     if (!fundingSources.length) {
         throw new Error(`No eligible funding fundingSources found to render buttons:\n\n${ JSON.stringify(fundingEligibility, null, 4) }`);
+    }
+
+    if (fundingSources.includes(FUNDING.CARD)) {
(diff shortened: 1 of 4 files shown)


allen-tong:
bump ICYMI @mnicpt @bluepnume

mnicpt (approved):

src/ui/buttons/buttons.jsx
```diff
@@ -98,13 +98,17 @@ export function Buttons(props : ButtonsProps) : ElementNode {
...
 
-    const fundingSources = determineEligibleFunding({ fundingSource, layout, remembered, platform, fundingEligibility, components, onShippingChange, flow, wallet });
+    let fundingSources = determineEligibleFunding({ fundingSource, layout, remembered, platform, fundingEligibility, components, onShippingChange, flow, wallet });
     const multiple = fundingSources.length > 1;
 
     if (!fundingSources.length) {
         throw new Error(`No eligible funding fundingSources found to render buttons:\n\n${ JSON.stringify(fundingEligibility, null, 4) }`);
     }
 
+    if (fundingSources.includes(FUNDING.CARD)) {
```
gregjopa, line 108: Hi @allen-tong, would you mind adding a test for this code? 

Ex:
```js
it.todo("if card is a funding source, then make sure it's the last one in the array")
```

allen-tong:
Thanks for the review @gregjopa! I've added a test, if you could please take a look when you get the chance :)

test/integration/tests/button/renderOrder.js
```diff
@@ -0,0 +1,62 @@
+/* @flow */
+/* eslint max-lines: 0 */
```
gregjopa, line 2: Is this esline max-lines override needed? I wouldn't expect it to be since this file only 62 lines long.
```suggestion
```

gregjopa (approved):
Thanks @allen-tong! Looks great 💯 

I left one minor comment about the eslint max-lines override. Please review that and then we will get this merged up.

src/ui/buttons/buttons.jsx
gregjopa: Thanks for adding the unit test @allen-tong! 😄

test/integration/tests/button/renderOrder.js
allen-tong: You're right, thanks for the catch!

PR #1489, opened 2020-12-07
Replace Travis with GitHub Actions

gregjopa:
This PR uses GitHub Actions to validate the build instead of Travis CI. Features include:
- running `npm run build` for Node 10, 12, and 14 in parallel. These jobs seem to take about 8 minutes to complete. This is about the same as Travis. The big win here is that GitHub Actions starts our jobs immediately where as Travis CI seems to queue jobs up for several minutes/hours before starting them.
- collecting code coverage and publishing to codecov.io.
- running for every pull request and on every push to the master branch

Screen Shot 2020-12-08 at 10 48 09 AM

The GitHub Actions workflow script is heavily inspired by kcd-scripts: https://github.com/kentcdodds/kcd-scripts/blob/master/.github/workflows/validate.yml


.github/workflows/main.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+
+      - name: 📥 Download deps
+        uses: bahmutov/npm-install@v1
+        with:
+          useLockFile: false
+
+      - name: ▶️ Run build script
```
gregjopa, line 34: I forgot about running `npm run flow-typed`. Will add that now.

gregjopa:
The GitHub Actions workflow is working! 😄 

The `npm run flow-typed` step is failing with the following error:
Screen Shot 2020-12-07 at 2 40 39 PM

I'm going to try downgrading flow-typed to attempt to resolve. cc: @mnicpt

gregjopa:
I was able to resolve the flow-bin errors by pinning to an exact version of flow-bin (`"flow-bin": "0.135.0"`)

.github/workflows/main.yml
```diff
@@ -0,0 +1,41 @@
...
+  # run for every pull request
+  pull_request: {}
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [10.13, 12, 14]
+    steps:
+      - name: 🛑 Cancel Previous Runs
```
bluepnume, line 16: Do we need this step? Sometimes it's weird if previous commits get a red dot, even if they would technically pass.

.github/workflows/main.yml
```diff
@@ -0,0 +1,41 @@
+name: build
+on:
+  # run on push but only for the master branch
+  push:
+    branches:
+      - master
```
bluepnume, line 6: Would it be useful to just run on every branch? Good to know if feature branches are passing tests too

bluepnume:
Looks awesome! Just a few minor Qs.

.github/workflows/main.yml
gregjopa: We can run it on every branch. It works well if everyone forks the main repo and sends PRs. But if we work off of upstream repos (like this PR is doing) you end up double running CI on PRs like so:

Screen Shot 2020-12-07 at 4 17 27 PM

If we limit it to just the main branch for "push" with the following code then we can continue working off upstream and not have this problem:

```
  # run on push but only for the master branch
  push:
    branches:
      - master
```

I'm also paranoid about kicking off too many jobs with GitHub Actions and causing things to queue up and take a long time. So I've been on the conservative side of only running CI when absolutely necessary.

.github/workflows/main.yml
gregjopa: I was researching this and found out that Travis CI does has this behavior by default (kill any previously running CI jobs for PRs when updates come in). I assumed we wanted it as well and added this step for it. 

Here's an example of this happening on this PR when I pushed a commit and then pushed another commit a couple minutes later: https://github.com/paypal/paypal-checkout-components/actions/runs/406899687

I agree that it's annoying to see red even though it's not a failure. The nice thing about canceling previous runs is we free up ci resources. 

I'll remove this and if it becomes a problem then we can add it back in.

karma.conf.js
```diff
@@ -57,6 +57,11 @@ export default function configKarma(karma : Object) {
 
             'src/index.js': [ 'webpack', 'sourcemap' ],
             'src/**/*.js':  [ 'sourcemap' ]
+        },
+
+        coverageReporter: {
```
gregjopa, line 62: I'm messing around with collecting coverage from karma and having github actions upload the report to codecov.io (ex: https://codecov.io/github/paypal/paypal-js). It's still not working. Still hacking away at it

gregjopa: I got code coverage working! https://codecov.io/gh/paypal/paypal-checkout-components/tree/e49924c26ed11539c3ae2bde68b7e31463e0cac8/src. Once this gets merged, the readme badge should start working too.

Screen Shot 2020-12-07 at 4 56 58 PM

.github/workflows/main.yml
gregjopa: @bluepnume  I'd like to leave in this "push but only for the master branch" setting. Let me know if you have any questions with the explanation above.

PR #1498, opened 2020-12-15
v4 - Fix aria label for Debit or Credit Card button

gregjopa:
This PR updates the aria-label for the Debit or Credit Card button.

### Steps to verify the fix

1. Clone this branch
2. `npm install`
3. Change the port number from 9000 to 9001 here: https://github.com/paypal/paypal-checkout-components/blob/v4/package.json#L32
4. `npm run dev`
5. Go to https://localhost.paypal.com:9001/webpack-dev-server/demo/dev/card.htm and verify the aria-label of the Card Section:

Screen Shot 2020-12-15 at 4 45 46 PM

### DIFF
--- a/src/button/config.js
+++ b/src/button/config.js
@@ -816,7 +816,9 @@ export const BUTTON_CONFIG : ButtonConfig = {
         allowPrimary: false,
 
         allowPrimaryVertical:   false,
-        allowPrimaryHorizontal: false
+        allowPrimaryHorizontal: false,
+
+        title: `${ FUNDING_BRAND_LABEL.CARD }`
     }
 };
 
--- a/demo/dev/card.htm
+++ b/demo/dev/card.htm
@@ -1,0 +1,59 @@
+<!DOCTYPE html>
+
+<head>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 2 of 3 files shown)


src/button/config.js
```diff
@@ -816,7 +816,9 @@ export const BUTTON_CONFIG : ButtonConfig = {
         allowPrimary: false,
 
         allowPrimaryVertical:   false,
-        allowPrimaryHorizontal: false
+        allowPrimaryHorizontal: false,
+
+        title: `${ FUNDING_BRAND_LABEL.CARD }`
```
mstuart, line 821: Is the button component kind of a default component that others override/extend? If so, wondering if setting a default title would cause issues. Just checkin'

mnicpt: @gregjopa Is this an issue for v5 as well?

gregjopa: @mstuart I noticed this is the same approach that the PayPal button and Credit button are using for aria-label values. Ex: https://github.com/paypal/paypal-checkout-components/blob/v4/src/button/config.js#L212. Unsure if it's causing any issues yet. Still trying to work around getting a successful running local dev environment.

@mnicpt no, this is not an issue in v5. Only applies to v4.

gregjopa: @mstuart I got CI passing and added a local dev demo to help with verifying the fix. I think this config change is isolated to just the Debit/Credit Card button container.

bluepnume (approved):

PR #1499, opened 2020-12-15
GitHub actions v4

gregjopa:
This PR updates the v4 branch to use GitHub Actions.


gregjopa:
CI is failing with flow errors as expected due to type errors. Ex: https://github.com/paypal/paypal-checkout-components/runs/1559555800?check_suite_focus=true#step:6:136.

.flowconfig
```diff
@@ -1,4 +1,5 @@
 [ignore]
+.*/node_modules/@babel/core
```
gregjopa, line 2: This prevents the following flow error: https://github.com/paypal/paypal-checkout-components/runs/1559689002?check_suite_focus=true#step:6:93

src/legacy/button.js
```diff
@@ -64,7 +64,7 @@ function renderButton(id, { container, locale, type, color, shape, size }) : Zal
 }
 
 // eslint-disable-next-line flowtype/no-mutable-array
-export function renderButtons(id : string, buttons : Array<Object>) : ZalgoPromise<Array<Object>> {
+export function renderButtons(id : string, buttons : Array<Object>) : ZalgoPromise<$ReadOnlyArray<Object>> {
```
gregjopa, line 67: This resolves the following flow error:

```
Error --------------------------------------------------------------------------------------- src/legacy/button.js:69:12

Cannot return `ZalgoPromise.map(...)` because read-only array type [1] is incompatible with array type [2] in type
argument `R` [3].

   src/legacy/button.js:69:12
                   v------------------------------------
    69|     return ZalgoPromise.map(buttons, button => {
```

Ex: https://github.com/paypal/paypal-checkout-components/runs/1559689002?check_suite_focus=true#step:6:101

gregjopa:
CI should be fixed! This was my first fight with flow and I think I've won 😄 

I had to pin flow-bin to use the exact `0.70.0` version and make another couple tweaks.

nbierdeman (approved):
Looks good! Great work!

mnicpt (commented):
LGTM!

Issue #1248, opened 2019-11-07
Paypal v5 sdk checkout button still have problems with -unsafe-inline- unsafe-eval script [csp]


Deam94:
### Description

EDIT
--------------------------
I found data-csp-nonce is now available to add:
https://developer.paypal.com/docs/checkout/reference/customize-sdk/#csp-nonce
But this does not solve unsafe-eval errors. 
-------------

Im using the latest paypal sdk v5, i got told csp inline-script problems should be resolved, but i still got errors refused to evaluate and execute inline script:

```
[Report Only] Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'nonce-xxx' *.google.com *.paypal.com *.paypalobjects.com *.braintreegateway.com https://www.gstatic.com".
Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'nonce-xxx' *.google.com *.paypal.com *.paypalobjects.com *.braintreegateway.com https://www.gstatic.com".
Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'nonce-xxx' *.paypal.com *.paypalobjects.com *.google.com *.braintreegateway.com https://www.gstatic.com". Either the 'unsafe-inline' keyword, a hash ('sha256-xxx'), or a nonce ('nonce-...') is required to enable inline execution.
[Report Only] Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' 'nonce-xxx' *.google.com *.paypal.com *.paypalobjects.com *.braintreegateway.com https://www.gstatic.com". Either the 'unsafe-inline' keyword, a hash ('sha256-xxx'), or a nonce ('nonce-...') is required to enable inline execution.
```

Now even a new ressource popped up to bloat my site "tracking.ga.paypal.com".
I added all needed resources to my csp and nonces to my script. The button is working tough csp problems.

Code:
```
braintree.client.create({
  authorization: 'xxx'
}).then(function (clientInstance) {
  return braintree.paypalCheckout.create({
	client: clientInstance
  });
}).then(function (paypalCheckoutInstance) {
  return paypal.Buttons({
	  style: {
		layout:  'horizontal',
		color:   'blue',
		shape:   'rect',
		tagline: 'true'
	  },
	createBillingAgreement: function () {
[...]

PR #1513, opened 2021-01-06
Use nonce with container styles

BriceShatzer:
Fixes CSP issue raised in this [issue](https://github.com/paypal/paypal-checkout-components/issues/1248) and adds the `NoncedStyleElement` to help prevent regression involving nonce-less style tags in the future.

**Testing locally:**  
- clone & run this repo this repo locally: `npm install && npm run dev`  
- clone a local version of [clientsdknodeweb](https://github.paypal.com/Checkout-R/clientsdknodeweb) and [alias](https://github.paypal.com/Checkout-R/clientsdknodeweb#mounting-local-client-modules) it to your local version of `paypal-checkout-components` before [running it locally ](https://github.paypal.com/Checkout-R/clientsdknodeweb#quick-start)
- stick the 2 test files found [here](https://git.io/Jtvp4) in a directory and `npm install && node app.js`. It's set up to run a test page at `http://localhost:3000/` which loads a paypal button using the version of `clientsdknodeweb` that you have running locally, which is pulling from the local copy of `paypal-checkout-components`
- if you run into an SSL/HSTS issue when trying to load the test page do the `thisisunsafe` [thing](https://dev.to/brettimus/this-is-unsafe-and-a-bad-idea-5ej4)


package.json
```diff
@@ -5,7 +5,7 @@
     "main": "index.js",
     "scripts": {
         "dev": "babel-node $(npm bin)/webpack-dev-server --config webpack.config.dev.js --port 9001 --host localhost.paypal.com --open-page demo/dev/index.htm --https --hot=false --inline=false",
-        "lint": "eslint --ext .js --ext .jsx src/ test/ *.js",
+        "lint": "eslint --ext .js --ext .jsx src/ test/ *.js --fix",
```
mnicpt, line 8: I'm not so sure we should do this by default.  I've seen this wreak havoc before where linting was changed and broke the code.  Since there shouldn't be many linting errors, it seems safer to just be alerted to them and manually fixing them.

BriceShatzer: Fair. I was just getting annoyed with the linter constantly yelling at me about my sloppy whitespace lol

mnicpt:
@BriceShatzer I'm seeing that the only difference with the new `NoncedStyleElement` compared to the `Style` component is that `nonce` is required.  I would recommend that we just make `nonce` on the `Style` component required and default to an empty string if not passed in by the merchant.  So it's optional for the merchant and required for our components.

gregjopa:
>  I would recommend that we just make nonce on the Style component required and default to an empty string if not passed in by the merchant. So it's optional for the merchant and required for our components.

@mnicpt this is a great point. I agree. @BriceShatzer brought this up earlier too and shared this commit: https://github.com/krakenjs/jsx-pragmatic/commit/05d75e02db7ba5a66cec0349299f45a484381e2a. Let's update jsx-pragmatic with making nonce required and default to empty string. 

@BriceShatzer can you send a PR to jsx-pragmatic for this?

mnicpt:
@gregjopa That's great!  Will look for PR on jsx-pragmatic.

mnicpt:
@BriceShatzer What I meant above was to just alter the current Style component and don't have `NoncedStyleElement` at all.  Just change the `nonce` property within the Style component to be required with a default value for merchants.

gregjopa:
Hi @mnicpt, @BriceShatzer and @bluepnume synced up on Slack about the path forward with this PR. Here's a screenshot of the conversation. The TLDR is that jsx-pragmatic is meant to be generic so the Style nonce prop should be optional. And in our JS SDK we need nonce to be required and can accomplish this using the new `NoncedStyleElement` wrapper.

Screen Shot 2021-01-12 at 9 21 16 AM

mnicpt:
@gregjopa Thanks for the further context.  Taking another look at this.

src/lib/NoncedStyleElement.jsx
```diff
@@ -0,0 +1,19 @@
+/* @flow */
+/** @jsx node */
+
+import { node, type ChildType, type NullableChildrenType } from 'jsx-pragmatic/src/node';
+import { Style } from 'jsx-pragmatic/src/component';
+
+type StyleProps = {|
+    css : string | {| _getCss : () => string |},
+    nonce : ?string,
```
mnicpt, line 9: Is this what we really want here?  Is it ok if the `nonce` value is null or undefined?  Or do we want to enforce a value other than null or undefined with `nonce : string,`?

src/ui/text/text.jsx
```diff
@@ -25,13 +26,14 @@ export function Space() : ChildType {
 
 type PlaceHolderProps = {|
     chars : number,
-    color? : $Values<typeof TEXT_COLOR>
+    color? : $Values<typeof TEXT_COLOR>,
+    nonce : ?string
```
mnicpt, line 30: Same question here as above.

BriceShatzer:
This should be good for review if anybody wants to take a look at it @gregjopa @mnicpt

src/funding/paylater/config.jsx
```diff
@@ -41,11 +42,11 @@ export function getPaylaterConfig() : FundingSourceConfig {
             // }
 
             return (
-                <Style css={ css } nonce={ nonce }>
+                <NoncedStyleElement css={ css } nonce={ '' }>
```
bluepnume, line 45: Why not pass the nonce in here?

src/funding/paypal/template.jsx
```diff
@@ -158,6 +165,7 @@ export function Label(opts : LabelOptions) : ChildType {
 
 export function WalletLabelOld(opts : WalletLabelOptions) : ?ChildType {
     const { logoColor, instrument, locale, content, commit } = opts;
+    const nonce = opts.nonce || '';
```
bluepnume, line 168: Let's have a single place where we default the nonce to `''`, and make it a mandatory string everywhere else. Want to avoid many places having to default the value to `''`

src/zoid/modal/component.jsx
```diff
@@ -36,6 +39,9 @@ export function getModalComponent() : ModalComponent {
                 }
 
                 const { cspNonce } = props;
+                // const nonce = cspNonce || getCSPNonce() || '';
```
bluepnume, line 42: Can we keep the default as `|| getCSPNonce() || ''`? This means if a merchant passes `data-csp-nonce` to the sdk (which is the recommended pattern) we will use that default value.

(Honestly, maybe we should think about making `getCSPNonce` return `string` not `?string` to avoid having to do `getCSPNonce() || ''`, but happy to punt on that for now)

test/ssr/ssr.test.js
```diff
@@ -41,6 +46,7 @@ async function getButtonScript() : Promise<{| Buttons : (Object) => typeof Eleme
 }
 
 test(`Button should render with ssr, with minimal options`, async () => {
+    // const { page } = await setupBrowserPage;
```
mnicpt, line 49: Why are we leaving the comment?

test/ssr/ssr.test.js
```diff
@@ -52,6 +58,7 @@ test(`Button should render with ssr, with minimal options`, async () => {
         fundingEligibility
     }).render(html());
 
+
```
mnicpt, line 61: Why the empty line?

mnicpt (approved):
Left a few non-blocking comments.

src/funding/paypal/template.jsx
gregjopa: @bluepnume What do you think about defaulting nonce to an empty string here in sdk-client? https://github.com/paypal/paypal-sdk-client/blob/e2074894e776aa55881833c07158d1f66f70dc0d/src/script.js#L242. Currently it returns `undefined` if no nonce is set with data-csp-nonce attribute. Wondering if that should just default to empty string.

src/zoid/modal/component.jsx
BriceShatzer: The current implementation of this is to have `cspNonc` passed in as a prop, which I think is right?   
The commented out line of was  `const nonce = cspNonce || getCSPNonce() || '';` was me messing around with trying to set a fall back so certain tests wouldn't fail.

src/funding/paypal/template.jsx
bluepnume: Yup, that sounds good to me!

src/zoid/modal/component.jsx
bluepnume: `getCSPNonce()` will pull the nonce from the global `data-csp-nonce` prop passed to the SDK script tag. So we should probably respect/default to that whenever it's not passed as a prop, so any merchant using that prop can be confident it will be used by any child components. Then if a prop is passed it can override that value.

gregjopa:
Closing this one out since we are solving in other smaller PRs.

PR #1516, opened 2021-01-11
Remove usage of Array.prototype.includes to fix IE11

gregjopa:
I tested the JS SDK in IE11 over the weekend and noticed that it's breaking due to this usage of `includes()`. [Array.prototype.includes](https://caniuse.com/array-includes) is not available in older browsers so it requires a polyfill if we are going to use it. 

To resolve the issue, I replaced `includes` with the old school ES3 `indexOf('card') !== -1` check. This way we don't have to worry about bringing in a polyfill.

cc: @allen-tong


allen-tong (approved):
thanks for the catch @gregjopa!

PR #1523, opened 2021-01-14
Fix tagline

mnicpt:
Condition was `if (layout !== VERTICAL) return;` and needed to be either `=== VERTICAL` or `!== HORIZONTAL`. Also, there was an extra `<span>` wrapped around the tagline, the parent, which result in a 0 width, therefore, hiding it from the `toggleOptionals()`.


demo/dev/index.htm
```diff
@@ -3,7 +3,7 @@
 <head>
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <script src="/checkout.js" data-paypal-checkout></script>
+    <script src="../../dist/checkout.js?debug=true" data-paypal-checkout></script>
```
bluepnume, line 6: This was intentional, it's being served up by webpack-dev-server when you run `npm run dev`. Was that not working?

src/button/template/componentTemplate.jsx
```diff
@@ -403,8 +403,8 @@ function renderTagline({ label, tagline, color, locale, multiple, env, cards, ch
         return;
     }
   
-    
-    if (__WEB__ && layout !== BUTTON_LAYOUT.VERTICAL) {
+    // tagline is only supported in horizontal layout
+    if (layout === BUTTON_LAYOUT.VERTICAL) {
```
bluepnume, line 407: We also need to `return;` if `__WEB__` is true. We only show tagline on second render.

src/button/props.js
```diff
@@ -86,7 +86,7 @@ export const normalizeProps = memoize((props : Object, defs? : { locale? : Local
         [ BUTTON_STYLE_OPTIONS.SHAPE ]:        shape        = getButtonConfig(label, 'defaultShape'),
         [ BUTTON_STYLE_OPTIONS.BRANDING ]:     branding     = getButtonConfig(label, (layout === BUTTON_LAYOUT.VERTICAL) ? 'defaultVerticalBranding' : 'defaultBranding'),
         [ BUTTON_STYLE_OPTIONS.FUNDINGICONS ]: fundingicons = getButtonConfig(label, 'defaultFundingIcons'),
-        [ BUTTON_STYLE_OPTIONS.TAGLINE ]:      tagline      = getButtonConfig(label, 'defaultTagline'),
+        [ BUTTON_STYLE_OPTIONS.TAGLINE ]:      tagline      = style[BUTTON_STYLE_OPTIONS.TAGLINE] ? getButtonConfig(label, 'defaultTagline') : false,
```
bluepnume, line 89: Why change this? Doesn't `getButtonConfig(label, 'defaultTagline')` already give the correct default value?

demo/dev/index.htm
mnicpt: It was not.

src/button/template/componentTemplate.jsx
mnicpt: Updating...

src/button/props.js
mnicpt: It does but then if the user doesn't set the `tagline: false`, it defaults to true.

demo/dev/index.htm
mnicpt: No local overrides on this one.

src/button/props.js
gregjopa: For checkout.js do we want the tagline to default to being true? 

I was looking through the example code snippets and I think we do want the tagline on by default. This is the only code snippet I found in the v4 demo using tagline and it's explicitly setting it to false: https://developer.paypal.com/demo/checkout-v4/#/pattern/generic

gregjopa (approved):
Looks good! There's just the one open question about the default value for tagline.

src/button/props.js
bluepnume: Yep defaulting to true is the expected behavior. `getButtonConfig(label, 'defaultTagline')` should return `true`.

mnicpt: @gregjopa @bluepnume This was done as I was going off of the tests assuming they were the source of truth.  But no guessing that the tests were changed to match up with the incorrect code.  I've changed back and updated tests.

demo/dev/index.htm
gregjopa, line 6: There was some confusion around `npm run dev` vs `npm run demo`. Let's change this one back since it works fine with `npm run dev`:

```suggestion
    <script src="/checkout.js" data-paypal-checkout></script>
```

mnicpt: Yep.  Already on it from our conversation last week.   Thanks, @gregjopa

src/button/template/miscComponent.jsx
```diff
@@ -94,7 +94,7 @@ export function Tagline(tagColor : string, impression : ?string, text : string |
         ` } />);
     nodes[1] = (
         <div class={ `${ CLASS.TAGLINE } ${ CLASS.TAGLINE_COLOR }-${ tagColor }` }>
-            <span optional>{ text }</span>
```
bluepnume, line 97: Why remove `optional`? We still do need to hide the tagline where there isn't space for it right?

mnicpt: The text that gets past is `<span optional>{the_tagline}</span`

PR #1530, opened 2021-01-27
Adding nonce and branded params

shrutikapoor08:


src/zoid/checkout/component.jsx
```diff
@@ -75,7 +75,14 @@ export function getCheckoutComponent() : CheckoutComponent {
                     value:      getSessionID,
                     queryParam: true
                 },
-        
+
+                branded: {
+                    type:       'boolean',
+                    queryParam: true,
+                    required:   false,
+                    default:    true
```
bluepnume, line 83: Please leave out `default`. We can't change this for existing merchants.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@paypal/checkout-components",
-    "version": "5.0.135",
+    "version": "5.0.136",
```
bluepnume, line 3: Please unstage.

bluepnume:
Few comments, and an `npm run lint -- --fix` and this should be good. Thanks!

PR #1535, opened 2021-02-01
Update venmo eligibility

gregjopa:
This PR updates the client-side eligibility logic for venmo related to the user agent string. 

The venmo button should only display on mobile when using a valid native browser (we do not support 3rd party browsers like Chrome on iOS). It will make the venmo ineligible for the following use cases:
- WebViews (handled by `supportsPopups`)
- 3rd party mobile browsers (handled by `supportedNativeBrowser`)
- when `onShippingChange` is used

This new `requiresPopupSupport` eligibility setting is tied to this function in belter: https://github.com/krakenjs/belter/blob/821f3e135a055c1607313f8b604eda0cbc685ad3/src/device.js#L136. 

### Test Cases

I've been running the apps locally and testing with different user agent strings. Here are some screenshots:

Smart buttons - valid Android UA
pixel-2-valid-user-agent-string

Smart buttons  - invalid WebView UA
facebook-webview-invalid-user-agent-string

- Standalone button - valid iPhone UA
standalone-valid-user-agent-string

### DIFF
--- a/src/funding/funding.js
+++ b/src/funding/funding.js
@@ -2,7 +2,7 @@
 
 import type { FundingEligibilityType } from '@paypal/sdk-client/src';
 import { PLATFORM, FUNDING, COMPONENTS } from '@paypal/sdk-constants/src';
-import { values } from 'belter/src';
+import { values, supportsPopups } from 'belter/src';
 
 import type { Wallet } from '../types';
 import { BUTTON_LAYOUT, BUTTON_FLOW } from '../constants';
@@ -59,6 +59,10 @@ export function isFundingEligible(source : $Values<typeof FUNDING>,
     }
 
     if (fundingConfig.shippingChange === false && onShippingChange) {
+        return false;
+    }
+
+    if (fundingConfig.requiresPopupSupport === true && supportsPopups() === false) {
         return false;
     }
 
--- a/src/funding/funding.js
+++ b/src/funding/funding.js
@@ -2,7 +2,7 @@
 
 import type { FundingEligibilityType } from '@paypal/sdk-client/src';
 import { PLATFORM, FUNDING, COMPONENTS } from '@paypal/sdk-constants/src';
-import { values, supportsPopups, isAndroid, isChrome, isIos, isSafari } from 'belter/src';
+import { values } from 'belter/src';
 
 import type { Wallet } from '../types';
 import { BUTTON_LAYOUT, BUTTON_FLOW } from '../constants';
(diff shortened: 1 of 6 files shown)


src/funding/funding.js
```diff
@@ -62,6 +62,10 @@ export function isFundingEligible(source : $Values<typeof FUNDING>,
         return false;
     }
 
+    if (fundingConfig.requiresPopupSupport === true && supportsPopups() === false) {
```
bluepnume, line 65: This logic needs to be able to run on the server, so `supportsPopups()` will error. Instead we should pass down `supportsPopups` as a boolean flag into the renderer.

bluepnume (approved):
LGTM! Thanks Greg

gregjopa:
> `isSupportedNativeBrowser` function sounds fine, but let's keep it in this repo -- since the definition of a supported native browser is very paypal specific.

@bluepnume I just pushed up this change and I think this PR is ready to go. Please give it another review.

mnicpt (approved):

BriceShatzer (approved):

src/zoid/buttons/component.jsx
```diff
@@ -471,6 +472,28 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
...
+            supportsPopups: {
+                type:       'boolean',
+                value:      () => userAgentSupportsPopups(),
+                queryParam: true
+            },
+
+            supportedNativeBrowser: {
+                type:       'boolean',
+                value:      () => {
+                    if (isIos() && isSafari()) {
```
bluepnume, line 486: Mind re-adding `!supportsPopups() -> return false` here? Just for safety.

bluepnume (approved):
Looks good! just one small comment

src/zoid/buttons/component.jsx
gregjopa: Sure thing. I just pushed up this change. 

Thanks for reviewing again! 😄

wenderjean:
@gregjopa @bluepnume I'm trying to figure out why Venmo is not eligible to use the `onShippingChange` callback would you guys help me on this? I mean, it makes sense to think that the Venmo native app does not have a clear path to notice my application regarding address changes like the PayPal pop-up has but I couldn't find further documentation around the flow itself.

What I'm trying to do is support Venmo with my current implementation.

```js
<Buttons
  createOrder={createOrder}
  // onShippingChange={onShippingChange}
  onApprove={onApprove}
  onCancel={onCancel}
  style={{ height: 40 }}
  onInit={onInit}
/>
```

mnicpt:
@wenderjean The Venmo app doesn't support patching orders through smart-payment-buttons and manages it instead in the app. This is why we make it ineligible; so it doesn't render.

wenderjean:
@mnicpt Thanks for the answer, it helped me a lot. Just two more, in this case there is another option to calculate taxes and shipping costs in order to show them to the user? I mean a correct flow to use with Venmo? Should I use it for digital purchases only or for those customers I have collected addresses already?

PR #1538, opened 2021-02-08
Add 'supportedNativeBrowser' as a valid query param

gregjopa:
This fixes a bug with the new venmo eligibility logic. Previously `supportedNativeBrowser` wasn't being passed to the server as a query param so it broke eligibility on the server render.

Here's a screenshot of the fix showing these parameters being passed to the server locally:
venmo-eligibility-query-params


mnicpt (approved):

bluepnume (approved):
LGTM!

PR #1548, opened 2021-02-24
adds useNewVenmoPath query param

BriceShatzer:
Adds the query param `useNewVenmoPath` for use when venmo is used as the funding source.

**Testing locally:**  
- clone & run this repo this repo locally: `npm install && npm run dev`  
- clone a local version of [clientsdknodeweb](https://github.paypal.com/Checkout-R/clientsdknodeweb) and [alias](https://github.paypal.com/Checkout-R/clientsdknodeweb#mounting-local-client-modules) it to your local version of `paypal-checkout-components` before [running it locally ](https://github.paypal.com/Checkout-R/clientsdknodeweb#quick-start)
- stick the 2 test files found [here](https://git.io/Jt5ZM) in a directory and `npm install && node app.js`. It's set up to run a test page at `http://localhost:3000/` which loads a Venmo button using the version of `clientsdknodeweb` that you have running locally, which is pulling from the local copy of `paypal-checkout-components`
- if you run into an SSL/HSTS issue when trying to load the test page do the `thisisunsafe` [thing](https://dev.to/brettimus/this-is-unsafe-and-a-bad-idea-5ej4)

Screen Shot 2021-02-24 at 1 21 01 PM

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -497,7 +497,7 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
             useNewVenmoPath: {
                 type:       'boolean',
                 required:   false,
-                value:      ({ props }) => props && props.fundingSource && props.fundingSource === FUNDING.VENMO,
+                // value:      ({ props }) => props && props.fundingSource && props.fundingSource === FUNDING.VENMO,
                 queryParam: true
             }
         }
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -492,6 +492,13 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 type:       'boolean',
                 value:      isSupportedNativeBrowser,
                 queryParam: true
+            },
+
+            useNewVenmoPath: {
+                type:       'boolean',
+                required:   false,
+                value:      ({ props }) => props && props.fundingSource && props.fundingSource === FUNDING.VENMO,
+                queryParam: true
             }
         }
     });


gregjopa:
Hi @BriceShatzer, what is the purpose of this new `useVenmoNewPath` param? Can you explain the newPath vs oldPath? Currently it looks like `useVenmoNewPath` will always be set to true whenever venmo is eligible.

BriceShatzer:
I think you're right. This should be available & defined here, but set by the button init code & potentially moved over in controlled way via ELMO/etc. I've updated it.

bluepnume:
Hi, sorry not clear on what this is for? We already have an `enable-funding=venmo` param which covers this right?

BriceShatzer:
@bluepnume so we need the ability to pass in the additional `useNewVenmoPath` query param to help differentiate between the new & old venmo implementations for xo-landing-router which will be happening [here](https://github.paypal.com/Checkout/xo-landing-router/pull/568).  

from @mstuart :

> yeah, there’s this old experience (Pay w/ Venmo) that’s based on buyers opting in w/ a cookie that’s dropped by the Venmo app (in the Settings page). it still generates revenue. we don’t want to ramp that down until the new experience (Venmo Pay) generates more revenue. won’t take long. we’ll need to support both for a month or two.
> the query params passed to the universal link are controlled by JS SDK, so it’d be a pretty simple change on our end to add a new param. the Native SDK won’t need to make a change.

bluepnume:
Getting some mixed messages. Thought this only applied to non-spb flows. Started a slack thread.

bluepnume:
See slack thread -- we shouldn't need this change .You should only need to change xo-landing-router

- If fundingSource=venmo in url -> go to new venmo experience
- Else if venmo cookie present -> fall back to old venmo experience

PR #1549, opened 2021-02-24
Add client-side experiment for venmo

gregjopa:
This PR adds a new experiment for displaying Venmo.

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -5,7 +5,7 @@
 import { getLogger, getLocale, getClientID, getEnv, getIntent, getCommit, getVault, getDisableFunding, getDisableCard,
     getMerchantID, getPayPalDomainRegex, getCurrency, getSDKMeta, getCSPNonce, getBuyerCountry, getClientAccessToken, getPlatform,
     getPartnerAttributionID, getCorrelationID, getEnableThreeDomainSecure, getDebug, getComponents, getStageHost, getAPIStageHost, getPayPalDomain,
-    getUserIDToken, getClientMetadataID, getAmount, getEnableFunding, getStorageID, createExperiment } from '@paypal/sdk-client/src';
+    getUserIDToken, getClientMetadataID, getAmount, getEnableFunding, getStorageID } from '@paypal/sdk-client/src';
 import { rememberFunding, getRememberedFunding, getRefinedFundingEligibility } from '@paypal/funding-components/src';
 import { ZalgoPromise } from 'zalgo-promise/src';
 import { create, type ZoidComponent } from 'zoid/src';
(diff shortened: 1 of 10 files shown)


src/zoid/buttons/component.jsx
```diff
@@ -344,9 +346,7 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
             experiment: {
                 type:       'object',
                 queryParam: true,
-                value:      () => {
-                    return {};
-                }
+                value:      () => getVenmoExperiment(enableVenmoExperiment)
```
gregjopa, line 349: The thinking here is there will be two ways to see the Venmo button:
1. Using the `enable-funding=venmo` query param
2. Being on the allowed list and getting bucketed into the treatment.

bluepnume (approved):

src/zoid/buttons/util.js
```diff
@@ -34,3 +37,17 @@ export function isSupportedNativeBrowser() : boolean {
 
     return false;
 }
+
+export function createVenmoExperiment() : Experiment {
+    return createExperiment('enable_venmo', 0);
```
gregjopa, line 42: @bluepnume do these params look okay? 

1. name - My assumption is the name parameter value ("enable_venmo") is not tied to any elmo experiment. Instead, this name is used for the FTPI logs and we will need a custom dashboard to report on this experiment.
2. sample - I think a zero value for sample disables the experiment. And when we want to turn this on we will need to make a code change here.

Please let me know if I've overlooked anything.

test/paypal.js
```diff
@@ -6,7 +6,8 @@ import * as paypalCheckout from '../src/interface/button'; // eslint-disable-lin
 import * as paypalWallet from '../src/interface/wallet'; // eslint-disable-line import/no-namespace
 import * as paypalMarks from '../src/interface/marks'; // eslint-disable-line import/no-namespace
 
-insertMockSDKScript();
+// the enable-funding=venmo flag is temporarily needed for the venmo experiment
+insertMockSDKScript({ query: { 'client-id': 'test', 'enable-funding': 'venmo' } });
```
gregjopa, line 10: Let me know if there is a better way to mock this. I tried mocking the experiment state with `window.localStorage.setItem('enable_venmo', true)` but it would not work when being added to the multiple.js test file. Mocking the enable-funding value here seemed like the easiest way to get the integration tests passing.

src/zoid/buttons/component.jsx
```diff
@@ -235,6 +238,19 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 }
             },
 
+            onInit: {
+                type:     'function',
+                required: false,
+                default:  () => noop,
+                decorate: ({ props, value = noop }) => {
+                    return (...args) => {
+                        enableVenmoExperiment.logStart({ [ FPTI_KEY.BUTTON_SESSION_UID ]: props.buttonSessionID });
```
gregjopa, line 247: I noticed an older experiment using onInit this way to emit the logStart event. Is this needed?

src/zoid/buttons/util.js
bluepnume: Totally works!

test/paypal.js
bluepnume: 👍

src/zoid/buttons/component.jsx
bluepnume: Yep, this is needed

PR #1558, opened 2021-03-02
Create separate venmo experiments based on mobile platform

gregjopa:
This PR updates the existing Venmo experiment code so we have separate ramp controls for Android and iOS.


mstuart:
👍  @bluepnume, is this cool w/ you?

bluepnume (approved):

PR #1570, opened 2021-03-15
Add data-user-experience-flow

amyegan:
This change supports CCO requirements for pay with nonce flow

Dependent on...
* [paypal-sdk-constants #48](https://github.com/paypal/paypal-sdk-constants/pull/48)
* [paypal-sdk-client #62](https://github.com/paypal/paypal-sdk-client/pull/62)

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -27,7 +27,6 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
     const enableVenmoExperiment = createVenmoExperiment();
 
     const queriedEligibleFunding = [];
-    console.log(' in local buttons component');
     return create({
         tag:  'paypal-buttons',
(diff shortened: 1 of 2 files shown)


src/zoid/buttons/component.jsx
```diff
@@ -508,6 +507,12 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 type:       'boolean',
                 value:      isSupportedNativeBrowser,
                 queryParam: true
+            },
+            
+            userExperienceFlow: {
+                type:       'string',
+                required:   false,
+                value:      getUserExperienceFlow
             }
```
gregjopa, line 516: Does this need `queryParam: true` so the param is persisted in the querystring?

bluepnume: I think it's only needed on the client right now

bluepnume (approved):

src/zoid/buttons/component.jsx
gregjopa: Ok cool. Never mind then.

PR #1573, opened 2021-03-16
Update style tests

amyegan:
A recent pull request from the community revealed flaws in the style tests. This change modifies the test for empty string based on the issue brought up in that PR and adds a new test to check for style mutation

Related PR - #1568  

Update: This includes changes to all `style.color` tests and removes any mutation of the `style` object while still handling the empty string case for `style.color`

### DIFF
--- a/src/ui/buttons/props.js
+++ b/src/ui/buttons/props.js
@@ -319,7 +319,7 @@ export function normalizeButtonStyle(props : ?ButtonPropsInputs, style : ButtonS
         period
     } = style;
 
-    // if color is a empty or whitespace, reset it to undefined
+    // if color is a falsy value, set it to the default color from the funding config
     const color = style.color ? style.color : fundingConfig.colors[0];
 
     if (values(BUTTON_LAYOUT).indexOf(layout) === -1) {
--- a/src/ui/buttons/props.js
+++ b/src/ui/buttons/props.js
@@ -310,16 +310,17 @@ export function normalizeButtonStyle(props : ?ButtonPropsInputs, style : ButtonS
         throw new Error(`Expected ${ fundingSource || FUNDING.PAYPAL } to be eligible`);
     }
 
-    style.color = style.color ? style.color : fundingConfig.colors[0];
     const {
         label,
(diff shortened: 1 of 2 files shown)


test/integration/tests/button/style.js
```diff
@@ -86,16 +86,43 @@ describe('paypal button color', () => {
         }).render('#testContainer');
     });
 
-    it('should render a button with gold background when passed empty string', (done) => {
+    it('should set style.color to undefined when given an empty string', (done) => {
+        const style = {
+            color: ' '
```
bluepnume, line 91: I don't think we should accept `' '` and treat it as an empty string

bluepnume (approved):

src/ui/buttons/props.js
```diff
@@ -310,17 +310,18 @@ export function normalizeButtonStyle(props : ?ButtonPropsInputs, style : ButtonS
...
         label,
         layout = fundingSource ? BUTTON_LAYOUT.HORIZONTAL : fundingConfig.layouts[0],
-        color = fundingConfig.colors[0],
         shape = fundingConfig.shapes[0],
         tagline = (layout === BUTTON_LAYOUT.HORIZONTAL && !fundingSource),
         height,
         period
     } = style;
 
+    // if color is a empty or whitespace, reset it to undefined
```
gregjopa, line 322:
```suggestion
    // if color is a falsy value, set it to the default color from the funding config
```

test/integration/tests/button/style.js
```diff
@@ -61,41 +61,48 @@ describe('paypal button color', () => {
...
-
-        }).render('#testContainer');
+        }).render('#testContainer').then(() => {
+            assert.ok(getElementRecursive('.paypal-button-color-gold'));
+        });
     });
 
     it('should render a button with black background when passed "black"', () => {
-        window.paypal.Buttons({
+        return window.paypal.Buttons({
```
gregjopa, line 74: Thanks for adding tests and cleaning these up! 💯

gregjopa (approved):

PR #1575, opened 2021-03-16
Added branded standalone button query param

JianyiGao:


src/zoid/checkout/component.jsx
```diff
@@ -163,6 +163,12 @@ export function getCheckoutComponent() : CheckoutComponent {
                     required:   false
                 },
 
+                apmBrandedStandaloneButton: {
```
bluepnume, line 166: Can we just go with `branded`?

JianyiGao: `branded` seems to be an existing param https://github.com/paypal/paypal-smart-payment-buttons/pull/137/files#diff-e3a4efeee1834a0c5a8750c378aef4deb5d1a4d1638b72565c8b91759ab23211R167

bluepnume: Yes. We can use this param for the same purpose. It's designed to differentiate between branded and unbranded buttons.

PR #1577, opened 2021-03-18
Disallow more Disney URLs for Venmo

mstuart:
Disallow more Disney URLs for Venmo

### DIFF
--- a/src/config/config.js
+++ b/src/config/config.js
@@ -587,6 +587,94 @@ export const config = {
 
         'disneyaulani.com': {
             disable_venmo: true
+        },
+
+        '6abc.com': {
+            disable_venmo: true
+        },
+        
+        'abc30.com': {
+            disable_venmo: true
+        },
+        
+        'abc7.com': {
+            disable_venmo: true
+        },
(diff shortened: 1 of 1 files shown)


src/config/config.js
```diff
@@ -587,6 +587,94 @@ export const config = {
...
+        
+        'espnwsummit.com': {
+            disable_venmo: true
+        },
+
+        'plus.espn.com': {
+            disable_venmo: true
+        },
+
+        'subscriptioncard.disneyplus.com': {
```
mnicpt, line 676: @mstuart We really could just do the main domain as we just search for this in the domain.  Maybe just have `disneyplus.com`?

mnicpt: Also,  I would have a test where the domain is not in the disallowed list, the Venmo cookie is present, and they don't have Venmo in allowed or disallowed props to ensure that the button only shows if in the allowed prop.

mstuart: Sounds good. Fixed!

mnicpt (approved):

PR #1580, opened 2021-03-19
Block SFVC context for Venmo

mnicpt:
Don't show Venmo option when falling back to web because of being inside the context of Safari View Controller.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
         "jest-ssr": "jest test/ssr --env=node --no-cache",
         "jest-e2e": "rm -f ./test/e2e/screenshots/*.png && jest test/e2e",
         "get-size": "gzip -9 < dist/size.min.js | wc -c",
-        "check-size": "npm run webpack-size && if [ $(npm run -s get-size) -ge 85000 ]; then echo 'Bundle size greater than 65k'; exit 1; else npm run delete-size; fi;",
+        "check-size": "npm run webpack-size && if [ $(npm run -s get-size) -ge 70000 ]; then echo 'Bundle size greater than 65k'; exit 1; else npm run delete-size; fi;",
         "delete-size": "rm dist/size.* dist/report.html"
     },
     "files": [
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
         "jest-ssr": "jest test/ssr --env=node --no-cache",
         "jest-e2e": "rm -f ./test/e2e/screenshots/*.png && jest test/e2e",
         "get-size": "gzip -9 < dist/size.min.js | wc -c",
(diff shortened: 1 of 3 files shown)


src/funding/venmo/config.jsx
```diff
@@ -46,6 +47,10 @@ export function getVenmoConfig() : FundingSourceConfig {
             [ BUTTON_COLOR.GOLD ]:   BUTTON_COLOR.BLUE,
             [ BUTTON_COLOR.BLUE ]:   BUTTON_COLOR.SILVER,
             [ BUTTON_COLOR.SILVER ]: BUTTON_COLOR.BLUE
+        },
+
+        eligible: () => {
```
gregjopa, line 52: Could we incorporate this check in the `supportedNativeBrowser` prop? 

- https://github.com/paypal/paypal-checkout-components/blob/master/src/zoid/buttons/component.jsx#L506-L510
- https://github.com/paypal/paypal-checkout-components/blob/master/src/zoid/buttons/util.js#L22-L40

Venmo is already using the `supportedNativeBrowser` prop to help determine eligibility.

mnicpt: @gregjopa This is a great point!  Taking a look at it.

package.json
```diff
@@ -29,7 +29,7 @@
         "jest-ssr": "jest test/ssr --env=node --no-cache",
         "jest-e2e": "rm -f ./test/e2e/screenshots/*.png && jest test/e2e",
         "get-size": "gzip -9 < dist/size.min.js | wc -c",
-        "check-size": "npm run webpack-size && if [ $(npm run -s get-size) -ge 70000 ]; then echo 'Bundle size greater than 65k'; exit 1; else npm run delete-size; fi;",
+        "check-size": "npm run webpack-size && if [ $(npm run -s get-size) -ge 85000 ]; then echo 'Bundle size greater than 65k'; exit 1; else npm run delete-size; fi;",
```
bluepnume, line 32: This is concerning. The script size jumped 15k?

src/zoid/buttons/util.js
```diff
@@ -28,6 +28,10 @@ export function isSupportedNativeBrowser() : boolean {
         return false;
     }
 
+    if (isSFVC()) {
```
bluepnume, line 31: LGTM!

package.json
gregjopa: I tested this change locally and ran `npm run check-size` and it seemed to pass fine without changing this threshold. I got the following output:
```
Final bundle size: 66362
```

gregjopa (approved):
The change looks good! I left a comment about backing out the `check-size` change since it shouldn't be needed.

mnicpt:
@gregjopa So odd. Mine showed 75k. Thought that seemed high.  Will put back to 70k.

mstuart:
:+1: lgtm

PR #1582, opened 2021-03-22
Apple Pay

mnicpt:


src/funding/applepay/config.jsx
```diff
@@ -0,0 +1,59 @@
...
+        ],
+
+        logoColors:  {
+            [ BUTTON_COLOR.BLACK ]:  LOGO_COLOR.WHITE,
+            [ BUTTON_COLOR.WHITE ]:  LOGO_COLOR.BLACK
+        },
+
+        eligible: ({ fundingEligibility }) => {
+            const eligibility = fundingEligibility.card;
+            const branded = Boolean(eligibility && eligibility.branded);
```
mnicpt, line 43: Should I check for `branded` here or do this from props in Apple Pay flow?

src/funding/applepay/config.jsx
```diff
@@ -0,0 +1,59 @@
...
+        requiresPopupSupport:             false,
+        requiresSupportedNativeBrowser:   false,
+        shippingChange:                   true,
+
+        platforms: [
+            PLATFORM.DESKTOP,
+            PLATFORM.MOBILE
+        ],
+
+        layouts: [
```
mnicpt, line 24: Are we supporting both layouts for Apple Pay?

src/funding/applepay/config.jsx
```diff
@@ -0,0 +1,59 @@
...
+            [ BUTTON_COLOR.BLACK ]:  LOGO_COLOR.WHITE,
+            [ BUTTON_COLOR.WHITE ]:  LOGO_COLOR.BLACK
+        },
+
+        eligible: ({ fundingEligibility }) => {
+            const eligibility = fundingEligibility.card;
+            const branded = Boolean(eligibility && eligibility.branded);
+
+            const ua = getUserAgent();
+            const isIosSafari = isIOS() && isSafari();
```
bluepnume, line 46: Can't use these functions here, since this `eligible()` function will also be run on the server side without user agent. Anything we need, needs to be passed in as a prop. We can add an `applePaySupported` prop.

bluepnume: Thinking we can abstract out an `isApplePaySupported` function to belter/device.js?

src/funding/applepay/config.jsx
bluepnume: No harm in it.

src/funding/applepay/config.jsx
```diff
@@ -0,0 +1,59 @@
...
+            const isMacOS = ua ? Boolean(ua.match(/Macintosh.*AppleWebKit/i)) : false;
+            const isMacSafari = isMacOS && isSafari();
+            const isValidMobileVersion = ua ? Boolean(ua.match(/.*iPhone.*1[0-9]_[0-9]/i)) : false; // iOS 10+
+            const isValidMacVersion = ua ? Boolean(ua.match(/.*Macintosh.*OS.*?(1[0-9]_1[2-9])|.*?(1[1-9]_[0-9])/i)) : false; // macOS 10.12+
+
+            if (!branded && ((isIosSafari && isValidMobileVersion) || (isMacSafari && isValidMacVersion)) && window.ApplePaySession && window.ApplePaySession.canMakePayments()) {
```
bluepnume, line 52: Actually do we need all of the user agent checks if we're checking `window.ApplePaySession`?

mnicpt: @bluepnume I was wondering the same.  Would assume that if supported would be the presence of the `ApplePaySession`.  Versioning and platform must be more informational.

src/funding/applepay/config.jsx
```diff
@@ -0,0 +1,48 @@
...
+            BUTTON_COLOR.WHITE
+        ],
+
+        logoColors:  {
+            [ BUTTON_COLOR.BLACK ]:  LOGO_COLOR.WHITE,
+            [ BUTTON_COLOR.WHITE ]:  LOGO_COLOR.BLACK
+        },
+
+        eligible: () => {
+            if (window.ApplePaySession && window.ApplePaySession.canMakePayments()) {
```
bluepnume, line 41: Can't do this here either :) gotta pass it as a prop. Will be no `window` on the server.

mnicpt: @bluepnume  Shoot! Right!  I keep looking at this as being on the client. Wondering about creating another product specific prop though so we don't have to keep creating these for each new button.  Maybe just an 'isSupported' for all?

src/ui/buttons/button.jsx
```diff
@@ -58,7 +58,13 @@ export function Button({ fundingSource, style, multiple, locale, env, fundingEli
     } = style;
 
     if (multiple && i > 0) {
-        color = secondaryColors[color] || secondaryColors[BUTTON_COLOR.DEFAULT] || colors[0];
+        if (secondaryColors[color]) {
```
bluepnume, line 61: We should also check if `colors.indexOf(secondaryColors[color]) !== -1` right?

mnicpt: Yes. Thank you

mnicpt: Yes. Thank you.

bluepnume (approved):
LGTM!

gregjopa (approved):
Nice work! I like the new refactor with the `requires` object.

PR #1584, opened 2021-03-23
Persist correlationID prop in the query string

gregjopa:
This PR updates the existing correlationID prop to persist in the query string to make it easier to debug.

This is the correlationID that gets embedded into the JS SDK script (as the `getCorrelationID()` function). It can be used to look up the logs and see the initial funding eligibility request details.

When running this PR branch locally you will see the correlationID get passed in the query string of the smart buttons call:
local-buttons-query-string-screenshot


amyegan (approved):
lgtm

shrutikapoor08 (approved):

PR #1586, opened 2021-03-24
Apple Pay Marks

mnicpt:


src/funding/applepay/config.jsx
```diff
@@ -25,7 +25,8 @@ export function getApplePayConfig() : FundingSourceConfig {
             BUTTON_LAYOUT.VERTICAL
         ],
 
-        Logo: ({ logoColor, optional }) => ApplePayLogo({ logoColor, optional }),
+        Logo:  ({ logoColor, optional }) => ApplePayLogo({ logoColor, optional }),
+        Marks: ({ ...props }) => ApplePayMark({ ...props }),
```
bluepnume, line 29: Can we just call this `Mark` since there's only one?

src/marks/template.jsx
```diff
@@ -88,6 +97,10 @@ export function MarksElement({ fundingEligibility, fundingSources, height, exper
                     .paypal-button-card:last-child {
                         margin-right: 0px;
                     }
+
+                    .paypal-mark .paypal-logo-applepay {
```
bluepnume, line 101: Can we keep this in the applepay.js file, maybe in its own style tag?

src/ui/buttons/props.js
```diff
@@ -190,6 +190,7 @@ export type RenderButtonProps = {|
     experiment : Experiment,
     vault : boolean,
     userIDToken : ?string,
+    applePay : Object,
```
bluepnume, line 193: Can we use a full type for applePay? `Object` is very similar to using `any` and doesn't provide much type safety for inner keys.

src/zoid/buttons/util.js
```diff
@@ -74,3 +74,35 @@ export function getVenmoExperiment(experiment : ?Experiment) : VenmoExperiment {
...
+                        'shippingmethodselected',
+                        'shippingcontactselected',
+                        'paymentauthorized',
+                        'cancel'
+                    ];
+                    if (validNames.indexOf(name) === -1) {
+                        // eslint-disable-next-line no-console
+                        console.error(`Invalid ApplePaySession event ${ name }`);
+                    }
+                    session[`on${ name }`] = handler;
```
bluepnume, line 101: Does applepay not support `addEventListener`?

bluepnume:
Looks great to me! Just a few minor comments.

src/funding/applepay/config.jsx
mstuart: No, there's two.

<---

bluepnume: onlyone

src/zoid/buttons/util.js
mnicpt: Just thought it was a bit long.

src/zoid/buttons/util.js
```diff
@@ -74,3 +74,35 @@ export function getVenmoExperiment(experiment : ?Experiment) : VenmoExperiment {
         enableVenmo: Boolean(isExperimentEnabled || isEnableFundingVenmo)
     };
 }
+
+export function applePaySession() : Function {
+    try {
+        if (!window.ApplePaySession) {
+            return undefined;
```
gregjopa, line 81: I think we can just use `return;` and that will return undefined.

```suggestion
            return;
```

gregjopa:
PR looks good to me. CI is failing though with the following error:
```
Cannot create `MarkLogo` element because property `env` is missing in object type [1] but exists in props [2].
[incompatible-use]

   src/marks/template.jsx:34:20
   34|                 ? <MarkLogo
                          ^^^^^^^^

References:
   src/funding/common.jsx:84:14
   84|     Mark? : ({||}) => ChildType,
                    ^^^^ [1]
   src/marks/template.jsx:34:19
                         v--------
   34|                 ? <MarkLogo
   35|                     fundingEligibility={ fundingEligibility }
   36|                     locale={ getLocale() }
   37|                     experiment={ experiment }
   38|                     env={ env }
   39|                 />
                       -^ [2]
```

src/zoid/buttons/util.js
```diff
@@ -74,3 +74,35 @@ export function getVenmoExperiment(experiment : ?Experiment) : VenmoExperiment {
...
+                    if (validNames.indexOf(name) === -1) {
+                        // eslint-disable-next-line no-console
+                        console.error(`Invalid ApplePaySession event ${ name }`);
+                    }
+                    session[`on${ name }`] = handler;
+                }
+            };
+        };
+    } catch (e) {
+        return undefined;
```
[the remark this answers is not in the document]
gregjopa: One more `undefined` to remove.

gregjopa (approved):
Looks good to me! 

I'm pretty new to Apple Pay to it would be good to get @bluepnume's 👍  as well.

mnicpt:
Thanks @gregjopa

src/zoid/buttons/util.js
```diff
@@ -74,3 +74,38 @@ export function getVenmoExperiment(experiment : ?Experiment) : VenmoExperiment {
...
+        }
+
+        return (version, request) => {
+            const session = new window.ApplePaySession(version, request);
+            return {
+                addEventListener: (name, handler) => {
+                    session.addEventListener(name, handler);
+                },
+                completeMerchantValidation: (validatedSession) => {
+                    session.completeMerchantValidation(validatedSession);
```
bluepnume, line 91: Do these return promises? Should we return those so the caller can await them?

mnicpt: @bluepnume No.  They are fire and forget and return void: https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession/1778012-completepayment .  We do have 30s to fire from getting the `onpaymentauthorized` event or they will invalidate the payment which is good if something happens when calling onApprove.

bluepnume (approved):

PR #1597, opened 2021-03-30
Remove correlationID from buttons

nbierdeman:
### Purpose

This PR is step 3, "Remove the correlationID prop from checkout-components," of [this PR](https://github.com/paypal/paypal-checkout-components/pull/1585). It removes `correlationID` from buttons.

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -284,12 +284,6 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 value: getEnableThreeDomainSecure
             },
 
-            correlationID: {
-                type:       'string',
-                required:   false,
-                value:      getCorrelationID
-            },
-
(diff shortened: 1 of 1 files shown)


mnicpt:
@nbierdeman What is the history around removing this?  Not required as a prop and only for logging?

nbierdeman:
@mnicpt, @gregjopa is working on using more descriptive names for props to help clarify what each is used for. I believe he and @shrutikapoor08 are going to eventually create a master doc to share around.

gregjopa:
> What is the history around removing this? Not required as a prop and only for logging?

@mnicpt this PR is just finishing up with renaming the `correlationID` prop to `sdkCorrelationID`. We also include `sdkCorrelationID` in the query string (this was not the case with the old `correlationID` prop) to make it easier to debug our logs and associate button renders to sdk script builds.

PR #1609, opened 2021-04-16
Only show venmo on supported mobile browsers

gregjopa:
This PR adds additional user agent string checks for Venmo. The venmo button should only show on Safari on iOS and Chrome on Android.

This PR also refactors device.js to take advantage of belter and also to support the new `isSupportedNativeBrowser` function which is the same approach v5 takes with determining to show venmo.

#### Safari on iOS

safari-ios-venmo-screenshot

#### Chrome on iOS

Chrome on iOS is not supported. Note how the code sets funding.disallowed to prevent venmo from rendering in the second render.
chrome-ios-no-venmo-screenshot

### DIFF
--- a/src/button/component.jsx
+++ b/src/button/component.jsx
@@ -1028,13 +1028,6 @@ export const Button : Component<ButtonOptions> = create({
             def() : Object {
                 return { action: 'checkout' };
             }
-        },
-
-        // its unclear if this prop is actually needed for second render
-        supportedNativeBrowser: {
-            type:       'boolean',
-            value:      isSupportedNativeBrowser,
-            queryParam: true
         }
     }
 });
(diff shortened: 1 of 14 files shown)


src/button/component.jsx
```diff
@@ -1028,6 +1028,13 @@ export const Button : Component<ButtonOptions> = create({
             def() : Object {
                 return { action: 'checkout' };
             }
+        },
+
+        // its unclear if this prop is actually needed for second render
```
gregjopa, line 1033: Best I can tell, this prop is not needed since v4 uses funding.disallowed to prevent venmo from rendering on second render. https://github.com/paypal/paypal-checkout-components/pull/1609/files#diff-91faf8fd6bd34f82cce8f38fed64acd81ca263e9de4a00d6c2727fa4ac129b89R516

mnicpt (approved):
This is great refactoring this and getting this synced up with v5!

PR #1613, opened 2021-04-20
Update buttons container to use nonce

gregjopa:
This PR fixes an issue with the Buttons container not respecting the optional `data-csp-nonce` value passed into the JS SDK script as an attribute. With this change the buttons will successfully render using the nonce without throwing csp errors. 

### Before

In prod the container inline styles fail to render (https://github.com/paypal/paypal-checkout-components/blob/master/src/zoid/buttons/container.jsx#L94).

csp-test-app-prod-error-screenshot
csp-test-app-prod-markup-screenshot

### After

This change resolves the container inline style error. Note that it still shows a csp error for the local logger which is expected when running locally.

csp-test-app-local-screenshot

## Next Steps

This PR does not resolve all the CSP errors. On button click the `<Overlay>` component fails to render because it doesn't accept a nonce yet. That will be the next issue to resolve.


src/lib/cspNonce.js
```diff
@@ -0,0 +1,18 @@
+/* @flow */
+
+export type NormalizeNonceInputs = {|
```
gregjopa, line 3: The `csp` types will improve after this PR is merged: https://github.com/paypal/paypal-sdk-client/pull/67

src/zoid/buttons/container.jsx
```diff
@@ -53,12 +54,13 @@ export function containerTemplate({ uid, props, tag, context, frame, prerenderFr
     });
 
     // $FlowFixMe
-    const { style } = props;
+    const { style, csp, nonce: nonceProp } = props;
```
bluepnume, line 57: My memory is fuzzy on this -- but is it actually possible to pass `nonce` as a raw prop right now?

If so, would it make sense to just define this as a prop in buttons/component.jsx and default it there?

```
nonce: {
    type: 'string',
    default: getCSPNonce
}
```

gregjopa: Thanks @bluepnume that's a great suggestion. We do have `nonce` defined as a prop right now (https://github.com/paypal/paypal-checkout-components/blob/master/src/ui/buttons/props.js#L320). I made this change and got rid of the normalizeNonce helper function.

Do you think we should remove the `csp` prop? https://github.com/paypal/paypal-checkout-components/blob/master/src/zoid/buttons/component.jsx#L438-L446. With this approach I don't think it's needed anymore. We can always leverage `getCSPNonce()` from sdk-client when we need to know if its a global sdk script nonce vs a nonce passed in to a component as a prop. 🤔

gregjopa: I did not remove the `csp` prop since it seems valuable to know the source of the nonce.

mnicpt (approved):

BriceShatzer (approved):
Looks good 👍

PR #1616, opened 2021-04-20
add oninit event to fields

andyfen:
- add onInit event to fields 

- remove background color for iframe to allow transparency

### DIFF
--- a/src/zoid/fields/component.jsx
+++ b/src/zoid/fields/component.jsx
@@ -117,6 +117,11 @@ export function getFieldsComponent() : FieldsComponent {
                     value:      getEnv
                 },
 
+                onInit: {
+                    type:     'function',
+                    required: false
+                },
+
                 buyerCountry: {
                     type:       'string',
                     queryParam: true,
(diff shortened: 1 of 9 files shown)


src/zoid/fields/component.jsx
```diff
@@ -117,6 +117,11 @@ export function getFieldsComponent() : FieldsComponent {
                     value:      getEnv
                 },
 
+                onMount: {
```
bluepnume, line 120: In other components we use `onInit` for this. Can we follow the same convention here?

test/integration/tests/fields/standalone.js
```diff
@@ -0,0 +1,54 @@
+/* @flow */
```
bluepnume, line 1: I think this file is not being imported right now. Is the test actually running?

test/integration/windows/checkout/index.js
```diff
@@ -21,6 +21,10 @@ if (action === 'checkout') {
 
         return ZalgoPromise.try(() => {
 
+            if (window.xprops.onInit) {
```
bluepnume, line 24: Would probably need to create a `windows/fields` -- this file is only used for the checkout component which has a different url and different props.

PR #1621, opened 2021-04-23
Update types

mnicpt:
Update types for Apple Pay and ordering.
image

### DIFF
--- a/src/funding/funding.js
+++ b/src/funding/funding.js
@@ -2,7 +2,7 @@
 
 import type { FundingEligibilityType } from '@paypal/sdk-client/src';
 import { PLATFORM, FUNDING, COMPONENTS } from '@paypal/sdk-constants/src';
-import { values } from 'belter/src';
+import { SUPPORTED_FUNDING_SOURCES } from '@paypal/funding-components/src';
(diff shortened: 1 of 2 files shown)


mstuart (approved):

src/funding/funding.js
```diff
@@ -100,7 +100,7 @@ export function determineEligibleFunding({ fundingSource, layout, platform, fund
         return [ fundingSource ];
     }
 
-    let eligibleFunding = values(FUNDING).filter(source =>
+    let eligibleFunding = SUPPORTED_FUNDING_SOURCES.filter(source =>
```
gregjopa, line 103: This looks good to me. `SUPPORTED_FUNDING_SOURCES` is already an array so we should use it and respect that order:
- https://github.com/paypal/paypal-funding-components/blob/c6df70ecf6c97cf31ececd254772974b747a658f/src/config.js#L7
- https://github.com/paypal/paypal-sdk-constants/blob/master/src/funding.js#L3

gregjopa (approved):

shrutikapoor08 (approved):

PR #1628, opened 2021-05-07
Update ApplePaySession

mnicpt:
Event was not being passed through post-robot messages.  Setting up all listeners in checkout-components and registering listeners.

### DIFF
--- a/src/zoid/buttons/util.js
+++ b/src/zoid/buttons/util.js
@@ -82,9 +82,35 @@ export function applePaySession() : ?ApplePaySessionConfigRequest {
 
         return (version, request) => {
             const session = new window.ApplePaySession(version, request);
+            const listeners = {};
+
+            session.onvalidatemerchant = (e) => {
+                listeners.validatemerchant(e.validationURL);
+            };
+
+            session.onpaymentmethodselected = () => {
+                listeners.paymentmethodselected();
+            };
+
+            session.onshippingmethodselected = () => {
+                listeners.shippingmethodselected();
+            }
+
+            session.onshippingcontactselected = () => {
+                listeners.shippingcontactselected();
+            }
+
+            session.onpaymentauthorized = (e) => {
+                listeners.paymentauthorized(e.payment);
+            }
+
+            session.oncancel = () => {
+                listeners.cancel();
+            }
+                            
             return {
-                addEventListener: (name, handler) => {
-                    session.addEventListener(name, handler);
+                addListener: (name, handler) => {
+                    listeners[name] = handler;
                 },
                 completeMerchantValidation: (validatedSession) => {
                     session.completeMerchantValidation(validatedSession);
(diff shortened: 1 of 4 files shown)


src/zoid/buttons/util.js
```diff
@@ -82,9 +82,35 @@ export function applePaySession() : ?ApplePaySessionConfigRequest {
 
         return (version, request) => {
             const session = new window.ApplePaySession(version, request);
+            const listeners = {};
+
+            session.onvalidatemerchant = (e) => {
+                listeners.validatemerchant(e.validationURL);
```
bluepnume, line 88: Can we keep the same structure of the event object? e.g. `listeners.validatemerchant({ validationURL: e.validationURL })`

src/zoid/buttons/util.js
```diff
@@ -82,9 +82,35 @@ export function applePaySession() : ?ApplePaySessionConfigRequest {
...
+            };
+
+            session.oncancel = () => {
+                listeners.cancel();
+            };
+                            
             return {
-                addEventListener: (name, handler) => {
-                    session.addEventListener(name, handler);
+                addListener: (name, handler) => {
```
bluepnume, line 112: Hey Steve do you think maybe it was changing the above `listeners` that fixed this, rather than changing `addEventListener` to `addListener`? 

I'm wondering if previously we had problems because the `Event` object from the listener was being directly passed through post-robot and maybe failed to serialize, back when you were directly doing `session.addEventListener(name, handler)`

Can you try changing this to `addEventListener` but keeping the above `listeners` mapping, and see if it works?

bluepnume: (don't waste too much time on it if it doesn't work, but I do think there are advantages to someone looking at the code in the iframe and seeing calls that look like a normal apple-pay session. thinking people may be confused if they see calls like `addEvent` which aren't in the regular apple pay spec. but not a huge deal if it's still a blocker)

mnicpt: Possibl.  Will check.

mnicpt: `addEventListener` is fine with `listeners` mapping.

bluepnume: sweeeet, must have been the event thing. thanks for testing!

src/zoid/buttons/util.js
```diff
@@ -82,9 +82,35 @@ export function applePaySession() : ?ApplePaySessionConfigRequest {
 
         return (version, request) => {
             const session = new window.ApplePaySession(version, request);
+            const listeners = {};
+
+            session.onvalidatemerchant = (e) => {
+                listeners.validatemerchant(e.validationURL);
+            };
+
+            session.onpaymentmethodselected = () => {
```
bluepnume, line 91: This approach works for me, I guess this also means we can decide exactly which parts of the `event` to map and pass back. I think it's probably rare we add new events in future.

bluepnume (approved):

PR #1640, opened 2021-05-18
Apple Pay Interface Fix

mnicpt:
Getting undefined on `event` when passing as `e.<data>`.  Pulling out and setting into constant fixes this.

### DIFF
--- a/src/funding/applepay/config.jsx
+++ b/src/funding/applepay/config.jsx
@@ -38,6 +38,8 @@ export function getApplePayConfig() : FundingSourceConfig {
         logoColors:  {
             [ BUTTON_COLOR.BLACK ]:  LOGO_COLOR.WHITE,
             [ BUTTON_COLOR.WHITE ]:  LOGO_COLOR.BLACK
-        }
+        },
+
+        shippingChange: true
     };
 }
--- a/src/zoid/buttons/util.js
+++ b/src/zoid/buttons/util.js
@@ -89,19 +89,23 @@ export function applePaySession() : ?ApplePaySessionConfigRequest {
             };
 
             session.onpaymentmethodselected = (e) => {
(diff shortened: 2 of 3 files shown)


amyegan (approved):
lgtm

src/zoid/buttons/util.js
```diff
@@ -89,19 +89,23 @@ export function applePaySession() : ?ApplePaySessionConfigRequest {
             };
 
             session.onpaymentmethodselected = (e) => {
```
gregjopa, line 91: Have you tried destructuring like this? Seems strange that the previous solution didn't work. Wondering if we need a fasly check on the args of these callbacks.
```js
session.onpaymentmethodselected = ({ paymentMethod }) => {
  listeners.paymentmethodselected({ paymentMethod });
};

Comment (mnicpt) [reply src/zoid/buttons/util.js]: That works!  Thanks @gregjopa

File: src/funding/applepay/config.jsx [hunk 2]
```diff
@@ -38,6 +38,8 @@ export function getApplePayConfig() : FundingSourceConfig {
         logoColors:  {
             [ BUTTON_COLOR.BLACK ]:  LOGO_COLOR.WHITE,
             [ BUTTON_COLOR.WHITE ]:  LOGO_COLOR.BLACK
-        }
+        },
+
+        shippingChange: true
```
gregjopa, line 43: That's cool that ApplePay supports the shippingCallback! 💯

gregjopa (approved):

src/funding/applepay/config.jsx
mnicpt: We patch in off of Apple's Event listeners.  Now just have to get it working by upgrading LSAT.

PR #1645, opened 2021-05-24
Fix/659/overlay nonce

alan-bendjuya:
- Read nonce from props for Overlay component
- Set default value for nonce
- Added test


bluepnume (approved):

mstuart:
lgtm 👍

src/zoid/installments/component.jsx
```diff
@@ -33,14 +33,16 @@ export function getInstallmentsComponent() : InstallmentsComponent {
             logger: getLogger(),
 
             prerenderTemplate: ({ doc, props }) => {
+                const nonce = props.nonce || getCSPNonce();
```
gregjopa, line 36: Why is the or condition needed?

```suggestion
                const { nonce } = props;
```

PR #1646, opened 2021-05-24
Vaulted button for returning Venmo Pay users

gabrielo91:
# Description

- Similar to the "Pay Now" button for returning Pay Later users, we need to create a button for returning Venmo Pay buyers that has Venmo branding and has the buyer's Venmo username in there.

## Screenshots

<a href="https://gyazo.com/5af7fbcf3c4515c9085b0b3a1b58d0c6">Image from Gyazo</a>


angels7777:
Please payment

PR #1650, opened 2021-05-26
Fix Vue example

leogedler:
Modified the Vue example adding missing returning functions on the `onCreateOrder` and `onApprove` props.
Now this new simple Vue example is working as expected.

Reference: https://github.com/krakenjs/zoid/tree/master/demo/frameworks/vue

### DIFF
--- a/demo/vue.htm
+++ b/demo/vue.htm
@@ -1,59 +1,59 @@
 <!DOCTYPE html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <title>Vue Demo - PayPal JS</title>
+  <head>
+    <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


demo/vue.htm
```diff
@@ -1,58 +1,59 @@
...
+        template: `
+          <paypal-buttons :on-approve="onApprove" :create-order="createOrder" />
+        `,
+        components: {
+          "paypal-buttons": PayPalButton,
         },
-        onApprove: function (data, actions) {
-          return actions.order.capture();
+    
+        computed: {
```
Artem-Schander, line 29: This needs to be `methods`

leogedler: @Artem-Schander thanks for the feedback. In this case we used `computed` because we don't pass in any param  with this methods and we need to Vue to keep track of them when down stream the JS SDK call them (This is just a wrapper for the JS SDK).

PR #1658, opened 2021-06-02
Add err.code = 'validation_error' for specific error in Buttons render

agnjunio:
This PR is a pre-requisite for https://github.com/paypal/paypal-smart-payment-buttons/pull/161

After some discussion with @gregjopa and @bluepnume, it was suggested that Buttons should throw an error with err.code = 'VALIDATION_ERROR', so it could be catched in smart-payment-buttons as a clientError, thus throwing a 4xx code instead of 5xx.

### DIFF
--- a/src/lib/errors.js
+++ b/src/lib/errors.js
@@ -1,0 +1,13 @@
+/* @flow */
+
+import { ERROR_CODE } from '@paypal/sdk-constants/src';
+
+export class ValidationError extends Error {
+  code : string;
+
+  constructor(message : string) {
+      super(message);
+      this.name = 'ValidationError';
+      this.code = ERROR_CODE.VALIDATION_ERROR;
+  }
+}
--- a/src/ui/buttons/buttons.jsx
+++ b/src/ui/buttons/buttons.jsx
@@ -8,6 +8,7 @@ import { noop } from 'belter/src';
 import type { Wallet, WalletInstrument } from '../../types';
 import { CLASS, BUTTON_NUMBER, BUTTON_LAYOUT, BUTTON_FLOW } from '../../constants';
 import { determineEligibleFunding, isWalletFundingEligible } from '../../funding';
+import { ValidationError } from '../../lib';
 
 import { normalizeButtonProps, type ButtonPropsInputs, type OnShippingChange } from './props';
 import { Style } from './style';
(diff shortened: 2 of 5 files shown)


src/ui/buttons/buttons.jsx
```diff
@@ -102,7 +102,9 @@ export function Buttons(props : ButtonsProps) : ElementNode {
     const multiple = fundingSources.length > 1;
 
     if (!fundingSources.length) {
-        throw new Error(`No eligible funding fundingSources found to render buttons:\n\n${ JSON.stringify(fundingEligibility, null, 4) }`);
+        const err = new Error(`No eligible funding fundingSources found to render buttons:\n\n${ JSON.stringify(fundingEligibility, null, 4) }`);
+        err.code = 'validation_error';
```
agnjunio, line 106: I didn't find any way to use `ERROR_CODES.VALIDATION_ERROR` constant from https://github.com/paypal/paypal-smart-payment-buttons/blob/master/server/config/constants.js#L40, so I used the hard-coded value. Anyone has a suggestion on how to do this?

gregjopa: That's a good point. Ideally we should share constants using the sdk-constants repo. Can you send a PR to paypal-sdk-constants that adds an error.js file here: https://github.com/paypal/paypal-sdk-constants/tree/master/src. And then both paypal-checkout-components and smart-payment-buttons can get this from the constants:

```js
export const ERROR_CODE = {
  VALIDATION_ERROR: 'validation_error'
}
```

agnjunio: PR is open: https://github.com/paypal/paypal-sdk-constants/pull/59

src/lib/errors.js
```diff
@@ -0,0 +1,13 @@
+/* @flow */
+
+import { ERROR_CODE } from '@paypal/sdk-constants';
```
gregjopa, line 3: I noticed the other sdk-constants import directly from the src folder:
```suggestion
import { ERROR_CODE } from '@paypal/sdk-constants/src';
```

gregjopa:
@agnjunio this looks good. Could you add a unit test to cover this new `ValidationError` class?

gregjopa (approved):

test/ssr/ssr.test.js
```diff
@@ -81,6 +81,29 @@ test(`Button should fail to render with ssr, with invalid style option`, async (
...
+    try {
+        Buttons({
+            locale:             { country: 'US', lang: 'en' },
+            platform:           'desktop',
+            sessionID:          'xyz',
+            buttonSessionID:    'abc',
+            fundingEligibility: {}
+        }).render(html());
+    } catch (err) {
+        expectedErr = err;
```
gregjopa, line 99: Can we assert on `err.code` here?

PR #1659, opened 2021-06-02
Update Pay Later label for DE and FR

elizabethmv:
**Note:** Pending some releases/PRs before this is ready for merge.

This PR adds a check against variants returned for paylater products to determine the branding of the button label for DE and FR buyer country.

### DIFF
--- a/src/funding/paylater/config.jsx
+++ b/src/funding/paylater/config.jsx
@@ -27,8 +27,16 @@ export function getPaylaterConfig() : FundingSourceConfig {
 
             let label = <Text>Pay Later</Text>;
 
+            if (paylater?.products?.paylater?.eligible && paylater?.products?.paylater?.variant === 'DE') {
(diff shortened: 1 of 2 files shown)


mnicpt (approved):

src/funding/paylater/config.jsx
```diff
@@ -27,10 +27,18 @@ export function getPaylaterConfig() : FundingSourceConfig {
 
             let label = <Text>Pay Later</Text>;
 
+            if (paylater?.products?.paylater?.eligible && paylater?.products?.paylater?.variant === 'DE') {
+                label = <Text>Später Bezahlen</Text>;
+            }
+
             if (paylater?.products?.payIn4?.eligible) {
```
mnicpt, line 34: I would put this above the previous one.  Use it as the default and override based on country.

PR #1666, opened 2021-06-18
QR Code component updates

BriceShatzer:
- Updates the QRCode component to it's most current props
- Fixes a [bug](https://engineering.paypalcorp.com/jira/browse/DTCHKINT-860) where clicks on prerendered Venmo buttons would launch native flow in a desktop context.


src/zoid/buttons/prerender.jsx
```diff
@@ -31,8 +32,21 @@ export function PrerenderedButtons({ nonce, onRenderCheckout, props } : Prerende
...
+            }
+            if (label) {
+                label.setAttribute('style', 'display:none;');
+            }
+
+            onRenderCheckout({ fundingSource, card });
+
+        } else if (supportsPopups()) {
             // remember the popup window to prevent showing a new popup window on every click in the prerender state
             if (!win || win.closed) {
```
gregjopa, line 51: I believe this `window` check needs to happen for all use cases. Otherwise this bug will be introduced: https://github.com/paypal/paypal-checkout-components/pull/1635

src/zoid/buttons/util.js
```diff
@@ -18,6 +18,13 @@ export function determineFlow(props : ButtonProps) : $Values<typeof BUTTON_FLOW>
     }
 }
 
+export function canUseQRPay(funding : $Values<typeof FUNDING>) : boolean {
```
gregjopa, line 21: Could we simplify this condition? What do you think of this?
```js
export function canUseQRPay(funding : $Values<typeof FUNDING>) : boolean {
    if (funding === FUNDING.VENMO && !isIos() && !isAndroid()) {
        return true;
    }

    return false;
}
```

src/zoid/buttons/prerender.jsx
```diff
@@ -31,8 +34,21 @@ export function PrerenderedButtons({ nonce, onRenderCheckout, props } : Prerende
             [ FPTI_KEY.CONTEXT_ID ]:         props.buttonSessionID,
             [ FPTI_KEY.TRANSITION ]:         'process_button_prerender_click'
         }).flush();
-        
-        if (supportsPopups()) {
+
+        if (canUseQRPay) {
+            const target : HTMLElement = event.currentTarget;
+            const spinner = target.querySelector(`.${ CLASS.SPINNER }`);
```
gregjopa, line 40: What is the ideal behavior for the prerender click for QR Pay? My initial thought is it should launch a modal dialog with a loading spinner and second render replaces the content with the QR code. 

If we want to stick with this button loading spinner approach then can we make a helper function in button utils for adding this spinner? Something like `addButtonSpinner(element)`. That way we don't have class specific details in pre-render.

src/zoid/buttons/util.js
BriceShatzer: yep that seems easier to read

src/zoid/buttons/prerender.jsx
BriceShatzer: You're right in wanting to throw up the modal right away, but I think that could potentially create a situation where we launch the modal but then we do a native app switch. Putting the venmo button in a "wait & see" mode and then calling `onRenderCheckout` and let the rendered components figure out how to handle that button click makes the most sense to me, otherwise I feel like we'd end up duplicating the quasi-eligibility "context checking" code that exists in for this in SPB. 

That being said, moving the functionality to add the spinner out of this file makes sense. I've moved it into `zoid/buttons/utils.js => showButtonLoading()`

gregjopa: > You're right in wanting to throw up the modal right away, but I think that could potentially create a situation where we launch the modal but then we do a native app switch.

Good point. I didn't think about this use case.

> That being said, moving the functionality to add the spinner out of this file makes sense. I've moved it into zoid/buttons/utils.js => showButtonLoading()

Sounds good. Let's go with that

src/zoid/buttons/util.js
```diff
@@ -18,6 +18,30 @@ export function determineFlow(props : ButtonProps) : $Values<typeof BUTTON_FLOW>
     }
 }
 
+export function mightUseQRPay(funding : $Values<typeof FUNDING>) : boolean {
```
gregjopa, line 21: Can we use the prefix `supports`? We do that in a few other spots in this repo:
```suggestion
export function supportsQRPay(funding : $Values<typeof FUNDING>) : boolean {
```

src/zoid/buttons/util.js
```diff
@@ -18,6 +18,30 @@ export function determineFlow(props : ButtonProps) : $Values<typeof BUTTON_FLOW>
...
+        return true;
+    }
+
+    return false;
+}
+
+export function showButtonLoading (targetElement : HTMLElement) : void {
+    const buttonElement = (targetElement.getAttribute('role') === 'button') ?
+        targetElement :
+        targetElement.closest('[role="button"]');
```
gregjopa, line 32: We cannot use `closest()` since it doesn't work in all the browsers we support (IE11 😢 ) https://caniuse.com/?search=closest.

BriceShatzer: doesn't webpack fix that? We use [arrow functions](https://caniuse.com/arrow-functions) everywhere?

gregjopa: Things like arrow functions and const/let are easy to transpile and are safe to use. DOM apis like closest() would need to be polyfilled and that is not done automatically. We should figure out another selector to use instead of trying to use `closest()`

BriceShatzer:
since this is included in https://github.com/paypal/paypal-checkout-components/pull/1667/, moving to that PR

PR #1667, opened 2021-06-18
QRCode component eligibility updates 

BriceShatzer:
- Removes platform and feature requirements for Venmo
we must implement an eligibility solution prior to merging this. Most via an [ELMO experiment](https://github.com/paypal/paypal-checkout-components/blob/60e38603d1fd91f3f4faa5dcbcd24a2072caa0ea/src/zoid/buttons/component.jsx#L27).   

- Includes changes from https://github.com/paypal/paypal-checkout-components/pull/1666


src/zoid/buttons/util.js
```diff
@@ -18,6 +18,38 @@ export function determineFlow(props : ButtonProps) : $Values<typeof BUTTON_FLOW>
...
 
+export function supportsQRPay(funding : $Values<typeof FUNDING>) : boolean {
+    if (funding === FUNDING.VENMO && !isIos() && !isAndroid()) {
+        return true;
+    }
+
+    return false;
+}
+
+/* eslint-disable flowtype/require-exact-type */
```
gregjopa, line 29: Nice work with using `ownerDocument`. I wonder if there's a built-in Flow DOM type that would play nice with this approach. What error did you run into when trying the `HTMLElement` type?

src/zoid/buttons/util.js
```diff
@@ -18,6 +18,38 @@ export function determineFlow(props : ButtonProps) : $Values<typeof BUTTON_FLOW>
     }
 }
 
+export function supportsQRPay(funding : $Values<typeof FUNDING>) : boolean {
+    if (funding === FUNDING.VENMO && !isIos() && !isAndroid()) {
```
mstuart, line 22: isDevice() may be a better check, just to make sure that no other mobile OS's leak in here -- https://github.com/krakenjs/belter/blob/8cf35443d53cb81fcb7bce0bdb385aeaffc7423c/src/device.js#L8

src/zoid/buttons/util.js
BriceShatzer: Since the `EventTarget` isn't explicitly an `HTMLElement`[ in general built-in DOM typing](https://stackoverflow.com/questions/28900077/why-is-event-target-not-element-in-typescript), it's always going to throw an error when I try to call `ownerDocument` regardless if I try to pass in `event.target` typed as `HTMLElement` or event as `Event` and derive it as `event.target.ownerDocument`. Since the context of event is needed to drill into the frame, I figured it made the most sense to just pass the event rather then the target.

src/zoid/buttons/util.js
gregjopa: Good point Mark. I was wondering about the requirements for when to show the venmo button. Plus one for using `!isDevice()` for determining eligibility for showing venmo for the QR Pay experience.

BriceShatzer: yeah i use `isDevice()` later 🤔 

idk why I did that like that here...

src/zoid/buttons/util.js
gregjopa: Thanks for explaining. I'm okay with this workaround to please flow.

src/zoid/buttons/util.js
```diff
@@ -50,8 +82,8 @@ export function createVenmoExperiment() : Experiment | void {
     const isEligibleForVenmo = fundingEligibility && fundingEligibility[FUNDING.VENMO] && fundingEligibility[FUNDING.VENMO].eligible;
 
     // exclude buyers who are not eligible
-    // exclude integrations using enable-funding=venmo
-    if (!isEligibleForVenmo || isEnableFundingVenmo) {
+    // exclude non-desktop integrations using enable-funding=venmo
+    if (!isEligibleForVenmo || (isEnableFundingVenmo && isDevice())) {
```
gregjopa, line 86: What does `enable-funding=venmo` mean for Venmo Desktop? It appears that these integrations using would `enable-funding=venmo` would only get venmo desktop if they are bucketed into the experiment.

src/zoid/buttons/util.js
```diff
@@ -70,7 +102,7 @@ export function createVenmoExperiment() : Experiment | void {
 
 export function getVenmoExperiment(experiment : ?Experiment) : VenmoExperiment {
     const enableFunding = getEnableFunding();
-    const isEnableFundingVenmo = enableFunding && enableFunding.indexOf(FUNDING.VENMO) !== -1;
+    const isEnableFundingVenmo = enableFunding && enableFunding.indexOf(FUNDING.VENMO) !== -1 && isDevice();
```
gregjopa, line 105: One idea is to leverage `supportsQRPay()` so it's more clear on what this `isDevice()` check does. Just thinking about the future if you need to change out what browsers support QRPay then you only have to change that logic in one place `supportsQRPay()`.
```suggestion
    const isEnableFundingVenmo = enableFunding && enableFunding.indexOf(FUNDING.VENMO) !== -1 && supportsQRPay(FUNDING.VENMO);
```

src/zoid/buttons/util.js
gregjopa, line 86: Same here with leveraging `supportsQRPay`
```suggestion
    if (!isEligibleForVenmo || (isEnableFundingVenmo && supportsQRPay(FUNDING.VENMO) === false)) {
```

test/integration/tests/funding/venmo/index.js
```diff
@@ -43,3 +62,28 @@ describe(`venmo button label test`, () => {
         }).render('#testContainer');
     });
 });
+
+
+describe(`venmo desktop web button test `, () => {
+
+    beforeEach(() => {
+        createTestContainer();
+        window.navigator.mockUserAgent = WEBVIEW_USER_AGENT;
```
gregjopa, line 71: Should we use a desktop user agent string for this test?

gregjopa:
@BriceShatzer this is looking good! I left a few minor comments. There's also a test failing in CI:

```
venmo desktop web button test 
    ✗ should not display button when it's passed in as instrument on desktop web
	Error: Expected "[object Boolean]" to equal "[object Undefined]"
	    at Object.equal (test/integration/index.js:12011:40)
	    at Context.<anonymous> (test/integration/index.js:11460:31)
```

Once those are fixed, I'm happy to approve/merge.

src/zoid/buttons/util.js
BriceShatzer: Yep that is correct. Even though Venmo Desktop & Venmo App-Switch share literally everything (funding source, payment flow, query params, etc.) we're treating them as two separate things in regards to the roll out. 

My assumption is that there are merchants/implementations that use `enable-funding=venmo` to surface the Venmo App-Switch button/flow on mobile, but  we don't want to make Venmo Desktop available to initially. So until we change the experiment value to something other than 0, the only way Venmo Desktop will be available is by manually adding a page to experiment  via `window.localStorage.setItem('enable_venmo_desktop', true);`. 

Ideally, I think how this ultimately will/should work is that if a merchant or implementation wants to show venmo as funding source (either through natural eligibility or explicitly setting `enable-funding=venmo`), we do and then make a determination on whether to do initialize the QRCode or App-Switch flow based on their device.

src/funding/venmo/config.jsx
```diff
@@ -13,17 +12,8 @@ export function getVenmoConfig() : FundingSourceConfig {
     return {
         ...DEFAULT_FUNDING_CONFIG,
 
-        requires: {
-            popup:          true,
-            native:         true
```
gregjopa, line 18: This `requires.native` setting is what prevents Venmo from showing on unsupported mobile user agent string. I understand we need to remove this to support Venmo Desktop. But we do need to migrate this logic somewhere else so venmo does not show on unsupported mobile browsers.

src/zoid/buttons/util.js
gregjopa, line 105: We need `isSupportedNativeBrowser()` in this check as well.

gregjopa (approved):
🚀

PR #1671, opened 2021-06-22
Fix problem with window.popupBridge.onComplete and queryParams

leogedler:
Added logic to check if `window.popupBridge.onComplete` has or not queryItems before trying to accessing it.

This solves the problem when the flow gets cancelled and there is an error on the popupBridge when the JS SDK tries to get the queryItems from an undefined object.

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -228,7 +228,8 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                             start:     (url) => {
                                 return new ZalgoPromise((resolve, reject) => {
                                     window.popupBridge.onComplete = (err, result) => {
-                                        return err ? reject(err) : resolve(result.queryItems);
+                                        const queryItems = result && result.queryItems ? result.queryItems : {};
+                                        return err ? reject(err) : resolve(queryItems);
                                     };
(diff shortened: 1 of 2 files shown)


src/zoid/buttons/component.jsx
```diff
@@ -228,7 +228,8 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                             start:     (url) => {
                                 return new ZalgoPromise((resolve, reject) => {
                                     window.popupBridge.onComplete = (err, result) => {
-                                        return err ? reject(err) : resolve(result.queryItems);
+                                        const queryItems = result && result.queryItems ? result.queryItems : {};
```
gregjopa, line 231: When testing this locally, what is the value of `result`? I'm wondering what is causing `result.queryItems` to be invalid and require this default value.

leogedler: @gregjopa this is the log that I got:

```
06-23 17:05:55.448 I/chromium( 5672): [INFO:CONSOLE(26178)] "result ---> null", source: http://localhost.paypal.com:8000/sdk/js?client-id=alc_client1&currency=USD&debug=true (26178)
```

image

gregjopa (approved):

PR #1674, opened 2021-06-23
ApplePay Performance

mnicpt:


src/zoid/buttons/component.jsx
```diff
@@ -67,13 +67,13 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
...
                 style = {},
-                applePaySupport = isApplePaySupported(),
                 fundingEligibility = getRefinedFundingEligibility(),
                 supportsPopups = userAgentSupportsPopups(),
                 supportedNativeBrowser = isSupportedNativeBrowser(),
                 experiment = getVenmoExperiment(enableVenmoExperiment)
             } = props;
-
+            
+            const applePaySupport = fundingEligibility && fundingEligibility[FUNDING.APPLEPAY]?.eligible ? isApplePaySupported() : false;
```
gregjopa, line 76: I think you can drop the falsy check. Optional chaining does not blow up when doing `undefined.?`. Ex:

Screen Shot 2021-06-23 at 12 14 30 PM


```suggestion
            const applePaySupport = fundingEligibility[FUNDING.APPLEPAY]?.eligible ? isApplePaySupported() : false;
```

mnicpt: When I do that I get lint error.  That's why I added it.

package.json
```diff
@@ -4,7 +4,7 @@
     "description": "PayPal Checkout components, for integrating checkout products.",
     "main": "index.js",
     "scripts": {
-        "dev": "babel-node $(npm bin)/webpack-dev-server --config webpack.config.dev.js --port 9001 --host localhost.paypal.com --open-page demo/dev/index.htm --https --hot=false --inline=false",
+        "dev": "NODE_ENV=test babel-node $(npm bin)/webpack-dev-server --config webpack.config.dev.js --port 9001 --host localhost.paypal.com --open-page demo/dev/index.htm --https --hot=false --inline=false",
```
bluepnume, line 7: Can we remove this? I don't think we should run dev env in test mode.

src/zoid/buttons/component.jsx
gregjopa: ahh nevermind then.

package.json
gregjopa: @mnicpt I took a look at the webpack config and I think we can remove `NODE_ENV=test` from here. And to fix the error, we can add `__PAYPAL_DOMAIN__` and give it a proper value for local development: https://github.com/paypal/paypal-checkout-components/blob/master/webpack.config.dev.js#L30
```
__PAYPAL_DOMAIN__:   'https://www.sandbox.paypal.com',
```

src/zoid/buttons/component.jsx
mnicpt: Using now in funding config.

test/integration/tests/button/props.js
```diff
@@ -17,8 +17,7 @@ describe(`paypal button component props`, () => {
         destroyTestContainer();
     });
 
-    it('should render an Apple Pay button if applePaySupport is true', () => {
-        // setup applePaySupport
+    it.only('should render an Apple Pay button if eligible', () => {
```
gregjopa, line 20:
```suggestion
    it('should render an Apple Pay button if eligible', () => {
```

src/funding/applepay/config.jsx
```diff
@@ -3,6 +3,7 @@
 
 import { PLATFORM } from '@paypal/sdk-constants/src';
 import { ApplePayLogo, LOGO_COLOR } from '@paypal/sdk-logos/src';
+import { isApplePaySupported } from 'belter';
```
bluepnume, line 6: Gotta use `belter/src` to get tree shaking

PR #1678, opened 2021-06-28
FEATURE: Integrate Donate as a label

arielxin:
Introduce `Donate` label to SPB.
Epic Story: https://engineering.paypalcorp.com/jira/browse/DTEVEREST-48
Story for Release: https://engineering.paypalcorp.com/jira/browse/DTEVEREST-2629

Screen Shot 2021-06-30 at 11 41 42 AM


gregjopa:
Please do not check in the dist/button.js changes. That dist folder automatically gets updated when we release

arielxin:
@gregjopa sure removed it from the PR.

mstuart (approved):

src/funding/paypal/config.jsx
```diff
@@ -43,7 +45,50 @@ export function getPayPalConfig() : FundingSourceConfig {
         labelText: `${ FUNDING_BRAND_LABEL.PAYPAL }`,
 
         Logo,
-        Label,
+
+        Label: ({ ...opts }) => {
+            const BasicLabel = ({ logo, label, period, locale: { lang } }) => {
```
bluepnume, line 50: Can we just add to the existing `BasicLabel` rather than inlining everything here?

https://github.com/paypal/paypal-checkout-components/blob/4885982a8eae958d0a6f75bd4e8362a3eb65effc/src/funding/common.jsx#L98

arielxin: hmm so our requirement is to show `Donate` label only for the first button, if we add it to the existing BasicLabel as what we've done before in the first commit https://github.com/paypal/paypal-checkout-components/pull/1678/commits/05d060d8d7ea8dea5eb68d952e2d0dacffd0021e#diff-190ff8cc9dfcc7a3918ddf7b447584333f214924025c4be15aa81f2a68962320R103, we will be showing `donate` for other funding source too, like PayPal Credit.

gregjopa: Thanks for explaining the new requirement @arielxin. Could we try and re-use `BasicLabel` for the common logic? I haven't tested it, but here's what I'm thinking:

```js
Label: ({ ...opts }) => {
    const BasicLabelForPayPal = ({ logo, label, period, locale: { lang } } = options) => {
        const { Donate } = componentContent[lang];

        // only the paypal button is allowed to show the "Donate" label
        if (label === BUTTON_LABEL.DONATE && Donate) {
            return <Donate logo={ logo } />;
        }

        // use the shared template for all other label logic
        return BasicLabel(options);
    }

    return (
        <Fragment>
            <BasicLabelForPayPal { ...opts } />
            <ButtonPersonalization { ...opts } />
        </Fragment>
    );
}
```

arielxin: Thanks for the suggestion, have updated the PR accordingly.

bluepnume: If it's only available for PayPal button, let's make this config based rather than code based please. e.g. let's add a `supportedLabels: [ LABEL.PAY, LABEL.XYZ ]` value to each button config. Want to avoid having too many hard-coded if/else statements here that are very hard to grok vs just reading a config.

bluepnume: Also we would need to hook this into eligibility right? If someone passes `label: donate` we should fully suppress any buttons which do not support donation flow. So just "not rendering the label" isn't really enough here.

gregjopa: Great points @bluepnume! I overlooked funding eligibility requirements. I imagine PayLater and Credit do not support donations so we should make sure they are not eligible when this label is used.

arielxin: PayLater does not support donations, but Credit might do. Do we have to consider the eligibility for alternative funding source too, wechant, etc.?

arielxin: And if i'm understanding it correctly, there are two use cases:
1. For the funding source that donation txn is supported but we don't want to show the `donate` label, we should keep the button but hide the label.
2. For the funding source that donation txn is not supported, we should make it ineligible when `donate` label is being used, so the button won't show at all.

Are you suggesting we should maintain two label configs in each funding source's config, one for suppressing the button and one for hiding the label?

gregjopa (approved):

PR #1679, opened 2021-06-29
Use local web server for 'npm run dev'

gregjopa:
This PR updates the local dev server (`npm run dev`) to be used as the domain for the popup. This fixes the contents of the popup:

Screen Shot 2021-06-28 at 8 02 40 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "dependencies": {
         "@paypal/common-components": "^1.0.10",
         "@paypal/funding-components": "^1.0.17",
-        "@paypal/sdk-client": "^4.0.151",
+        "@paypal/sdk-client": "^4.0.156",
         "@paypal/sdk-constants": "^1.0.105",
         "@paypal/sdk-logos": "^1.0.26",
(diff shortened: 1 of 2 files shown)


nbierdeman:
Should I be seeing sandbox as the domain in the popup when on the master branch? 🤔 

Screen Shot 2021-06-29 at 9 21 47 AM

gregjopa:
@nbierdeman I think you need to stop the local server before switching branches. If you stop the server and checkout master and then restart it with `npm run dev` you should see the sandbox url that 404s. Let me know if that's not the case.

We don't publish our html assets in this repo so the only way to serve them up is to use the local web server. This is how it use to work before I broke it with the recent `__PAYPAL_DOMAIN__` global change 😄

nbierdeman:
I'm still getting localhost in the popup after stopping the server, switching to a feature branch, then back to `master`, and running `npm run dev` again.

nbierdeman (approved):
🚢  🥳

znomuz (commented):
[paypal](url)

PR #1685, opened 2021-07-07
Add client-side eligible checks before for Venmo experiments can start

agnjunio:
The purpose of this PR is to add an eligibility check on the Venmo experiment after the middleware has been called. The reason for this is that some eligibility checks were missing in the pre-render, which is where we have less information about the user. On the second render, on the other hand, the eligibility check was being skipped.

receive::req onInit shows the moment onInit is called on the second render phase. The `enable_venmo_ios_enable_venmo_ios_test_start` shows that the `experiment.logStart()` has been called.

Screen Shot 2021-07-07 at 16 16 57

which doesn't happen when the user is not actually eligible

Screen Shot 2021-07-07 at 16 21 30

# Steps to reproduce in production

1. Set User-Agent to mobile ios or android
2. Use this code in https://developer.paypal.com/demo/checkout/#/pattern/client
```
<!DOCTYPE html>
<html lang="en">
<head>
    <!-- Add meta tags for mobile and IE -->
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
    <title> PayPal Smart Payment Buttons Integration | Client Demo </title>
</head>
<body>
    <!-- Set up a container element for the button -->
    <div id="paypal-button-container"></div>
    <!-- Include the PayPal JavaScript SDK -->
    <script src="https://www.paypal.com/sdk/js?client-id=test&currency=USD&debug=true"></script>
    <script>
        // Render the PayPal button into #paypal-button-container
        paypal.Buttons({
            // Set up the transaction
            createOrder: function(data, actions) {
                return actions.order.create({
                    purchase_units: [{
                        amount: {
                            value: '88.44'
                        }
                    }]
                });
            },
            // Finalize the transaction
            onApprove: function(data, actions) {
                return actions.order.capture().then(function(details) {
                    // Show a success message to the buyer
                    alert('Transaction completed by ' + details.payer.name.given_name + '!');
                });
            },
            onShippingChange: function () {
              console.log(`onShippingChange callback`);
            }
    </script>
</body>
</html>
```
3. Search console for `enable_venmo_ios_enable_venmo_ios_test_start`

Screen Shot 2021-07-07 at 19 22 34

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -245,7 +245,16 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 default:  () => noop,
                 decorate: ({ props, value = noop }) => {
                     return (...args) => {
-                        if (enableVenmoExperiment) {
+                        const { onShippingChange, style = {}, fundingEligibility = getRefinedFundingEligibility(), applePaySupport, supportsPopups, supportedNativeBrowser } = props;
+
+                        const flow = determineFlow(props);
+                        const { layout } = style;
+
+                        const platform   = getPlatform();
+                        const components = getComponents();
+                        const experiment = getVenmoExperiment(enableVenmoExperiment);
+
+                        if (enableVenmoExperiment && isFundingEligible(FUNDING.VENMO, { layout, platform, fundingSource: FUNDING.VENMO, fundingEligibility, components, onShippingChange, flow, applePaySupport, supportsPopups, supportedNativeBrowser, experiment })) {
                             enableVenmoExperiment.logStart({ [ FPTI_KEY.BUTTON_SESSION_UID ]: props.buttonSessionID });
                         }
 


src/zoid/buttons/component.jsx
```diff
@@ -245,7 +245,16 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 default:  () => noop,
                 decorate: ({ props, value = noop }) => {
                     return (...args) => {
-                        if (enableVenmoExperiment) {
+                        const { onShippingChange, style = {}, fundingEligibility = getRefinedFundingEligibility(), applePaySupport, supportsPopups, supportedNativeBrowser } = props;
+
+                        const flow = determineFlow(props);
+                        const { layout } = style;
+
+                        const platform           = getPlatform();
```
gregjopa, line 253: Can you fix the indenting here? The equal signs seem so far away 😄

mstuart (approved):

gregjopa (approved):
Great work!

PR #1687, opened 2021-07-09
ppplmer-94708 label was cut between specific widths and only in German language

alan-bendjuya:
Requirements:

- PP monogram should always be shown in pay later button if the button is present. Same with the label
- `Mit "paypal" zahlen` for the Paypal Button and `Später Bezahlen` for Pay later button should have space for the labels or not be shown otherwise. 

Now:
width < 300px -> only Paypal button is being shown
300px < 320px -> If language is german, the label is smaller so it doesn't get cut from the button, for the other languages the label remains the same size
320px < width -> no changes

Screenshot: width 300px
image

Screenshot: width 315px
image

Screenshot: width 330px
image
 
Screenshot: width 300px in english
image

Screenshot: width 300px in French
image

[Ticket](https://engineering.paypalcorp.com/jira/browse/PPPLMER-94708)


src/ui/buttons/styles/responsive.js
```diff
@@ -1,29 +1,30 @@
...
+export function buttonResponsiveStyle({ height, locale } : {| height? : ?number, locale? : ?LocaleType |}) : string {
 
     return Object.keys(BUTTON_SIZE_STYLE).map(size => {
 
         const style = BUTTON_SIZE_STYLE[size];
         const buttonHeight = height || style.defaultHeight;
         const minDualWidth = Math.max(Math.round(buttonHeight * BUTTON_MIN_ASPECT_RATIO * (100 / WALLET_BUTTON_PERC)), MIN_SPLIT_BUTTON_WIDTH);
 
-        const labelHeight = max(roundUp(perc(buttonHeight, 35) + 5, 2), 12);
+        const textPercPercentage = locale?.lang === 'de' ? 32 : 36;
```
remotevision, line 24: Should the default be `35` here to match previous code?

alan-bendjuya: Yes, missed it

remotevision (approved):
👍

mstuart (approved):

src/ui/buttons/styles/responsive.js
```diff
@@ -132,6 +136,25 @@ export function buttonResponsiveStyle({ height } : {| height? : ?number |}) : st
                 }
             }
 
+            @media only screen and (min-width: ${ style.minWidth }px) and (max-width: 320px) {
+
+                .${ CLASS.CONTAINER } .${ CLASS.BUTTON_ROW } .${ CLASS.TEXT }, .${ CLASS.CONTAINER } .${ CLASS.BUTTON_ROW } .${ CLASS.SPACE } {
```
gregjopa, line 141: Do we need the class container included as well? Wondering if we can shorten up this selector like so:
```suggestion
                .${ CLASS.BUTTON_ROW } .${ CLASS.TEXT }, .${ CLASS.BUTTON_ROW } .${ CLASS.SPACE } {
```

alan-bendjuya: Actually If the container class is removed the one in line 38 will be used

gregjopa: ah good point.

PR #1698, opened 2021-07-20
Pin exact version of post-robot

nbierdeman:
### Purpose

This PR pins an exact version of `post-robot` to see if the CI will pass.


nbierdeman:
@bluepnume, it appears that the latest version of `post-robot` breaks the CI for checkout-components. Do you recommend we fix forward or roll back?

gregjopa:
Looks like we have this resolved. Closing this one out

PR #1700, opened 2021-07-23
feat(DTNAVYA-1153): Add logic to show tagline for multiple buttons

lucasfrotam:
This PR is about the Upfront Presentment project and adds a new queryParam so we can show the tagline for multiple buttons using Personalization Studio.

https://engineering.paypalcorp.com/jira/browse/DTNAVYA-1153

### DIFF
--- a/src/ui/buttons/tagline.jsx
+++ b/src/ui/buttons/tagline.jsx
@@ -30,7 +30,7 @@ export function TagLine({ fundingSource, locale, multiple, nonce, personalizatio
     return (
         <div class={ CLASS.TAGLINE }>
             {
-                (tagline && !multiple)
+                (tagline)
                     ? (
                         <Fragment>
                             <span>
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -19,7 +19,7 @@ import { isFundingEligible } from '../../funding';
 
 import { containerTemplate } from './container';
 import { PrerenderedButtons } from './prerender';
-import { applePaySession, determineFlow, isSupportedNativeBrowser, createVenmoExperiment, getVenmoExperiment, getButtonsArray } from './util';
+import { applePaySession, determineFlow, isSupportedNativeBrowser, createVenmoExperiment, getVenmoExperiment, getRenderedButtons } from './util';
 
 export type ButtonsComponent = ZoidComponent<ButtonProps>;
 
@@ -441,10 +441,10 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 value:      getMerchantID
             },
 
-            buttonsArray: {
+            renderedButtons: {
                 type:       'array',
                 queryParam: true,
-                value:      ({ props }) => getButtonsArray(props)
+                value:      ({ props }) => getRenderedButtons(props)
             },
 
             csp: {
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -19,7 +19,7 @@ import { isFundingEligible } from '../../funding';
 
 import { containerTemplate } from './container';
 import { PrerenderedButtons } from './prerender';
-import { applePaySession, determineFlow, isSupportedNativeBrowser, createVenmoExperiment, getVenmoExperiment } from './util';
+import { applePaySession, determineFlow, isSupportedNativeBrowser, createVenmoExperiment, getVenmoExperiment, getButtonsArray } from './util';
 
 export type ButtonsComponent = ZoidComponent<ButtonProps>;
 
@@ -439,6 +439,12 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 type:       'array',
                 queryParam: true,
                 value:      getMerchantID
+            },
+
+            buttonsArray: {
+                type:       'array',
+                queryParam: true,
+                value:      ({ props }) => getButtonsArray(props)
             },
 
             csp: {
--- a/src/zoid/buttons/util.js
+++ b/src/zoid/buttons/util.js
@@ -1,11 +1,13 @@
 /* @flow */
-import { supportsPopups, isAndroid, isChrome, isIos, isSafari, isSFVC, type Experiment, isDevice } from 'belter/src';
+import { supportsPopups as userAgentSupportsPopups, isAndroid, isChrome, isIos, isSafari, isSFVC, type Experiment, isDevice } from 'belter/src';
 import { FUNDING } from '@paypal/sdk-constants/src';
-import { getEnableFunding, createExperiment, getFundingEligibility } from '@paypal/sdk-client/src';
+import { getEnableFunding, createExperiment, getFundingEligibility, getPlatform, getComponents } from '@paypal/sdk-client/src';
+import { getRefinedFundingEligibility } from '@paypal/funding-components/src';
 
 import type { Experiment as VenmoExperiment } from '../../types';
 import { BUTTON_FLOW, CLASS } from '../../constants';
-import type { ApplePaySessionConfigRequest, CreateBillingAgreement, CreateSubscription } from '../../ui/buttons/props';
+import type { ApplePaySessionConfigRequest, CreateBillingAgreement, CreateSubscription, ButtonProps } from '../../ui/buttons/props';
+import { determineEligibleFunding } from '../../funding';
 
 type DetermineFlowOptions = {|
     createBillingAgreement : CreateBillingAgreement,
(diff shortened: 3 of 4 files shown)


src/zoid/buttons/util.js
```diff
@@ -106,6 +108,21 @@ export function getVenmoExperiment(experiment : ?Experiment) : VenmoExperiment {
     };
 }
 
+export function getButtonsArray(props : ButtonProps) : $ReadOnlyArray<$Values<typeof FUNDING>> {
```
gregjopa, line 111: I'm concerned about seeing the `getVenmoExperiment` call in this `getButtonsArray` function. Can you explain the purpose of this function? I'd like to see if we can simplify the logic.

gregjopa: I went through the other [SPB PR](https://github.com/paypal/paypal-smart-payment-buttons/pull/209) and it seems like we want to know if there is only one button rendered or if there are more than one. Which buttons are actually rendered doesn't seem to matter. And we need to know this information on second render so it can be passed to the personalization api call. Is that correct?

lucasfrotam: Hi Greg! So, I created this `getButtonsArray` function to get the buttons array and know if the page will render single or multiple buttons, according to the array length. And about the `getVenmoExperiment`, is because this prop is mandatory to call the `determineEligibleFunding` function, which returns the buttons array.

lucasfrotam: About the second comment, you are right! I don't use the buttons array to know which button will be rendered, but how many will be rendered. And with this information, I can assign a boolean to my variable `hasMultipleButtons`, which will be sent to the Personalization graphql call. In general, I'm using the same logic as we already use on [this part](https://github.com/paypal/paypal-checkout-components/blob/d408ae23346f7ce93797e06d1f15078592b94bff/src/ui/buttons/buttons.jsx#L104)

gregjopa: Thanks for explaining. That makes a lot more sense now. True button eligibility is a combination of the embedded server-side eligibility plus the client-side eligibility rules (support for shipping callback, tagline, layout, etc). This is unknown to SPB right now so I can understand why this new query param prop is needed.

src/zoid/buttons/component.jsx
```diff
@@ -451,6 +451,12 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 value:      getMerchantID
             },
 
+            buttonsArray: {
```
gregjopa, line 454: Can we drop the "Array" part from the prop name? Perhaps "buttons" or "renderedButtons" would work. The other existing props using `type: 'array'` do not include Array in the name so I would prefer not to do that here.

```suggestion
            buttons: {
```

src/zoid/buttons/component.jsx
```diff
@@ -451,6 +451,12 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
...
+                type:       'array',
+                queryParam: true,
+                value:      ({ props }) => getButtonsArray(props)
```
gregjopa, line 457: Let's update this name to match. Either `getButtons` or `getRenderedButtons`
```suggestion
                value:      ({ props }) => getRenderedButtons(props)
```

gregjopa:
This PR is looking good @lucasfrotam! Can you add a couple new tests for this new buttons prop?

src/ui/buttons/tagline.jsx
```diff
@@ -30,7 +30,7 @@ export function TagLine({ fundingSource, locale, multiple, nonce, personalizatio
     return (
         <div class={ CLASS.TAGLINE }>
             {
-                (tagline && !multiple)
```
ssono, line 33: Just do double check have you ensured that the default multiple buttons tagline still works after this?

src/zoid/buttons/component.jsx
ssono: agreed with a preference towards renderedButtons

ssono (commented):
Overall looks good to me. Lets make sure there is parity between both PRs before merging.
https://github.com/paypal/paypal-smart-payment-buttons/pull/209

src/ui/buttons/tagline.jsx
lucasfrotam: Yes, it's working! `Two easy ways to pay tagline` is being shown, which is the default tagline

src/zoid/buttons/component.jsx
lucasfrotam: Done!

src/zoid/buttons/component.jsx
lucasfrotam: Done!

src/zoid/buttons/util.js
```diff
@@ -106,6 +108,21 @@ export function getVenmoExperiment(experiment : ?Experiment) : VenmoExperiment {
     };
 }
 
+export function getRenderedButtons(props : ButtonProps) : $ReadOnlyArray<$Values<typeof FUNDING>> {
+    const { fundingSource, onShippingChange, style = {}, fundingEligibility = getRefinedFundingEligibility(),
+        experiment = getVenmoExperiment(createVenmoExperiment()), applePaySupport, supportsPopups = userAgentSupportsPopups(),
```
gregjopa, line 113: Currently calling `createVenmoExperiment()` is going to create a new belter client-side experiment each time. We currently call it here: https://github.com/paypal/paypal-checkout-components/blob/master/src/zoid/buttons/component.jsx#L27 and pass that value down to `getVenmoExperiment()`. 

I think the easiest solution is to memoize the `createVenmoExperiment()` function. That way we don't have to worry about if it gets called multiple times. Can you update the `createVenmoExperiment()` function to use memoize from belter?

gregjopa: Here's what I'm thinking for adding memoize to createVenmoExperiment():

```js
const createVenmoExperiment : Experiment | void = memoize(() => {
    const enableFunding = getEnableFunding();
    const isEnableFundingVenmo = enableFunding && enableFunding.indexOf(FUNDING.VENMO) !== -1;

    const fundingEligibility = getFundingEligibility();
    const isEligibleForVenmo = fundingEligibility && fundingEligibility[FUNDING.VENMO] && fundingEligibility[FUNDING.VENMO].eligible;

    // exclude buyers who are not eligible
    // exclude non-desktop integrations using enable-funding=venmo
    if (!isEligibleForVenmo || (isEnableFundingVenmo && isSupportedNativeBrowser())) {
        return;
    }

    if (isIos() && isSafari()) {
        return createExperiment('enable_venmo_ios', 90);
    }

    if (isAndroid() && isChrome()) {
        return createExperiment('enable_venmo_android', 90);
    }

    if (!isDevice()) {
        return createExperiment('enable_venmo_desktop', 0);
    }
});
```

lucasfrotam: I just tested it and all this information is coming in the props correctly. It's much simpler than I did

lucasfrotam: Could you please review it again?

ssono: @lucasfrotam I'm not sure I understand. Are you saying that the venmo information is already passed in the props so you don't have to call the venmo functions at all?

ssono:
@lucasfrotam I think your last commit broke some stuff

src/zoid/buttons/util.js
lucasfrotam: I talked to Greg and I really need to call the functions I was calling. I'm adding the functions again and I'm going to add the `memoize`

lucasfrotam: Done!

gregjopa:
@lucasfrotam can you rebase this PR with the master branch?

test/integration/tests/button/props.js
```diff
@@ -135,4 +135,33 @@ describe(`paypal button component props`, () => {
             });
         }));
     });
+
+    it('should render a button and get the renderedButtons props', () => {
+        return ZalgoPromise.try(() => {
+            return wrapPromise(({ expect, avoid }) => {
+                let onRender = ({ xprops }) => {
+                    const renderedButtons = xprops.renderedButtons;
+                    if (!renderedButtons.length) {
```
gregjopa, line 144: Can we check for the exact array values here? It would be great to confirm that all the expected funding sources show up.

lucasfrotam: Yes! I'll add this logic

lucasfrotam: Done

gregjopa (approved):
Looks great!

I tested locally and see the expected values being passed through for the new `renderedButtons` prop:

Screen Shot 2021-08-06 at 4 24 59 PM

PR #1703, opened 2021-07-26
Added styles hack to fix width getting cut off on older ipads

wizzrdd:


bluepnume:
I thought this style needed to be on the inner iframe (the one containing the card fields), not the button iframe?

dj110919871 (approved):

PR #1705, opened 2021-07-28
Add stickiness ID Checkout prop

gillycheesesteak:
Tied to [this PR](https://github.com/paypal/paypal-smart-payment-buttons/pull/211) in smart-payment-buttons

### DIFF
--- a/src/zoid/checkout/component.jsx
+++ b/src/zoid/checkout/component.jsx
@@ -87,7 +87,8 @@ export function getCheckoutComponent() : CheckoutComponent {
                 
                 stickinessID: {
                     type:       'string',
-                    queryParam: true
+                    queryParam: true,
+                    required:   false
                 },
                 
                 env: {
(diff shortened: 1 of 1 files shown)


src/zoid/checkout/component.jsx
```diff
@@ -85,6 +85,11 @@ export function getCheckoutComponent() : CheckoutComponent {
                     required:   false
                 },
                 
+                stickinessID: {
+                    type:       'string',
+                    queryParam: true
```
gregjopa, line 90: Can you provide an example value for this new stickinessID prop? Wondering how big they will be and if we need to pass through as a query string param or if we can get this from window.xprops.

gillycheesesteak: The ID (ex: `uid_85ecf248f0_mtm6mzy6ntq`) is similar to the sessionID. I believe it is required to be a queryParam since the goal is to have it accessible as part of the checkout state in UL

dj110919871 (approved):

bluepnume (approved):

PR #1708, opened 2021-08-03
Handle Venmo Desktop

mnicpt:
Revert settings for Mobile eligibility.

**Testing**
1. Tested Venmo button on iOS in a SFVC
2. Tested Venmo button on Desktop
3. Tested Venmo button on iOS outside of SFVC and successful app switch.


elizabethmv (approved):

src/zoid/buttons/util.js
```diff
@@ -77,33 +77,38 @@ export function createVenmoExperiment() : Experiment | void {
...
-    return {
-        enableVenmo: Boolean(isExperimentEnabled || isEnableFundingVenmo)
-    };
+    if (isDevice()) {
+        return {
+            enableVenmo: Boolean((isExperimentEnabled || isVenmoFundingEnabled) && isNativeSupported)
+        };
+    } else {
+        return {
+            enableVenmo: Boolean(isExperimentEnabled || isVenmoFundingEnabled)
```
gregjopa, line 109: I don't think `enable-funding=venmo` is suppose to show Venmo Desktop. Venmo Desktop should only be shown from being in the experiment test group.
```suggestion
            enableVenmo: Boolean(isExperimentEnabled)
```

mnicpt: Thanks, @gregjopa

TNuzzi: @gregjopa @mnicpt would there be another way to force Venmo to show?  The `enable-funding=venmo` is convenient for testing, specially for e2e.

test/integration/tests/funding/venmo/index.js
```diff
@@ -77,13 +77,13 @@ describe(`venmo desktop web button test `, () => {
     });
 
 
-    it(`should display button when it's passed in as instrument on desktop web`, (done) => {
+    it(`should not display button when it's passed in as instrument on desktop web`, (done) => {
```
gregjopa: Ideally we would have a test for each scenario (displaying and not displaying). What do you think about using the localstorage opt-in for the tests?

For example, we can use `localStorage.setItem('enable_venmo_desktop', true)` to opt into Venmo Desktop. And we could set this to false for the negative test case. That way the ramping percentage does not impact the test cases.

gregjopa (approved):

test/integration/tests/funding/venmo/index.js
gregjopa: ah ignore this. Just realized that this file is no longer changed.

PR #1710, opened 2021-08-04
Add queryParam for rendered buttons for use in personalization experiments

ssono:
- On initial render calculate buttons to be rendered
- Pass through smartcomponentnodeweb for the personalization call

### DIFF
--- a/src/button/config.js
+++ b/src/button/config.js
@@ -70,7 +70,7 @@ export const BUTTON_CONFIG : ButtonConfig = {
         tag:     `{ content: safer_tag }`,
         dualTag: `{ content: dual_tag|safer_tag }`,
 
-        defaultLocale:           `en_US`,
+        defaultLocale:           { country: 'US', lang: 'en' },
         defaultLabel:            BUTTON_LABEL.CHECKOUT,
         defaultVerticalLabel:    BUTTON_LABEL.PAYPAL,
         defaultColor:            BUTTON_COLOR.GOLD,
--- a/src/lib/funding.js
+++ b/src/lib/funding.js
@@ -4,13 +4,13 @@ import { ZalgoPromise } from 'zalgo-promise/src';
 
 import { FUNDING } from '../constants';
 import type { FundingSource, FundingList } from '../types';
+import { labelToFunding, getButtonConfig } from '../button/config';
+import { determineEligibleFunding } from '../funding';
+import { parseLocale } from '../button/props';
 
 import { getStorageState, getGlobalState, getSessionState } from './session';
 import { openMetaFrame } from './meta';
 import { identity } from './util';
-import { determineEligibleFunding } from '../funding';
-import { parseLocale } from '../button/props';
-import { labelToFunding, getButtonConfig } from '../button/config';
 
 // $FlowFixMe
 export function getRememberedFunding<T>(handler? : (rememberedFunding : FundingList) => T = identity) : T {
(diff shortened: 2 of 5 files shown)


ssono:
retest this please

dj110919871 (approved):
0

src/button/config.js
```diff
@@ -70,7 +70,7 @@ export const BUTTON_CONFIG : ButtonConfig = {
         tag:     `{ content: safer_tag }`,
         dualTag: `{ content: dual_tag|safer_tag }`,
 
-        defaultLocale:           `en_US`,
+        defaultLocale:           { country: 'US', lang: 'en' },
```
mnicpt, line 73: Good catch!

mnicpt (approved):

lucasfrotam:
The build error doesn’t seem to be related to your changes.  I'm investigating it

src/lib/funding.js
```diff
@@ -113,3 +116,28 @@ export function precacheRememberedFunding() : ZalgoPromise<void> {
     }
     return ZalgoPromise.resolve();
 }
+
+export function getRenderedButtons(props : Object) : string {
```
bluepnume, line 120: Please avoid using `Object` type, it provides no type safety :)

src/lib/funding.js
```diff
@@ -113,3 +116,28 @@ export function precacheRememberedFunding() : ZalgoPromise<void> {
...
+    const selected = labelToFunding(label);
+    const sources = determineEligibleFunding({
+        funding,
+        selected,
+        locale,
+        env,
+        layout,
+        commit
+    });
+    return String(sources);
```
bluepnume, line 142: Let's return an array. Zoid is smart enough to serialize that correctly for the query param.

src/lib/funding.js
ssono: I was basing it on what I had seen elsewhere. https://github.com/paypal/paypal-checkout-components/blob/v4/src/button/validate.js%20#L54

Should there be a defined propsType instead?

src/lib/funding.js
ssono: Done!

src/lib/funding.js
gregjopa: @ssono could we leverage the `normalizeProps` function here? https://github.com/paypal/paypal-checkout-components/blob/v4/src/button/props.js#L60. It seems like it returns `sources` and eliminates the need for alot of this logic in `getRenderedButtons`

ssono: Yes we can do this @gregjopa

src/lib/funding.js
```diff
@@ -113,3 +114,9 @@ export function precacheRememberedFunding() : ZalgoPromise<void> {
     }
     return ZalgoPromise.resolve();
 }
+
+export function getRenderedButtons(props : Object) : string {
+
+    const { sources } = normalizeProps(props);
+    return sources;
```
gregjopa, line 121: CI is failing with the following type error:
```

Cannot return `sources` because `FundingList` [1] is incompatible with string [2].

   src/lib/funding.js:121:12
   121|     return sources;
                   ^^^^^^^

References:
   src/button/props.js:50:15
    50|     sources : FundingList,
                      ^^^^^^^^^^^ [1]
   src/lib/funding.js:118:54
   118| export function getRenderedButtons(props : Object) : string {
                                                             ^^^^^^ [2]
```

You may need to use `sources.toString()`

ssono:
Screen Shot 2021-08-12 at 9 41 10 AM
Screenshots passing locally

PR #1718, opened 2021-08-10
Remove double checking on wallet

agnjunio:
Title.


gregjopa:
Nice catch @agnjunio!

PR #1720, opened 2021-08-13
Add logs for tagline display

ssono:


src/ui/buttons/buttons.jsx
```diff
@@ -118,6 +119,15 @@ export function Buttons(props : ButtonsProps) : ElementNode {
         ((__WEB__ && userIDToken) || Object.keys(instruments).length)
     );
 
+    if (!(tagline && layout === BUTTON_LAYOUT.HORIZONTAL && !fundingSource)) {
+        getLogger().track({
+            [FPTI_KEY.STATE]:      'SMART_BUTTON',
+            [FPTI_KEY.TRANSITION]: 'PROCESS_BUTTON_LOAD',
+            tagline_shown:         false,
```
gregjopa, line 126: Are there any existing FPTI_KEYs we can use? I worry that these new keys will not be searchable in reporting tools like kibana.

gregjopa: Could we update the transition name to indicate if the tagline is shown or not? Ex:

```js
[FPTI_KEY.TRANSITION]: 'BUTTON_TAGLINE_SHOWN',
```

and for when it's hidden:

```js
[FPTI_KEY.TRANSITION]: 'BUTTON_TAGLINE_HIDDEN',
```

And maybe use `FPTI_KEY.OPTION_SELECTED` for the actual tagline message

PR #1725, opened 2021-08-24
Added experiment for removing paylater button

ssono:
https://engineering.paypalcorp.com/jira/browse/DTUPSTREAM-16

- These changes are to create an experiment for removing the paylater button.
- All flows that have paylater button and haven't forced it on or off should be eligible
    - US buyers only (we don't want to run this experiment in other buyer regions where we have the BNPL button)
    - Horizontal or Vertical stacked buttons only. We should not run this on Standalone button integrations.
    - We should have the ability to exclude certain merchants from this experiment
- Used the venmo experiment as a template for this one

Related to https://github.paypal.com/Checkout-R/clientsdknodeweb/pull/285

### DIFF
--- a/src/funding/paylater/config.jsx
+++ b/src/funding/paylater/config.jsx
@@ -20,8 +20,15 @@ export function getPaylaterConfig() : FundingSourceConfig {
             BUTTON_LAYOUT.VERTICAL
         ],
 
-        eligible: ({ experiment }) => {
-            if (experiment && experiment.disablePaylater === true) {
+        eligible: ({ experiment, fundingEligibility, fundingSource }) => {
+            const { paylater } = fundingEligibility;
+            if (
+                experiment
+                && experiment.disablePaylater === true
+                && !fundingSource  // Exclude standalone buttons
+                && (paylater?.products?.paylater?.variant === 'experimentable'
+                    || paylater?.products?.payIn4?.variant === 'experimentable') // Only use tagged eligibility
+            ) {
                 return false;
             }
 
--- a/src/funding/paylater/config.jsx
+++ b/src/funding/paylater/config.jsx
@@ -20,6 +20,14 @@ export function getPaylaterConfig() : FundingSourceConfig {
             BUTTON_LAYOUT.VERTICAL
         ],
 
+        eligible: ({ experiment }) => {
+            if (experiment && experiment.disablePaylater === true) {
+                return false;
+            }
+
+            return true;
+        },
+
         Label: ({ logo }) => logo,
 
         Logo: ({ logoColor, nonce, fundingEligibility }) => {
--- a/src/funding/paylater/config.jsx
+++ b/src/funding/paylater/config.jsx
@@ -24,14 +24,13 @@ export function getPaylaterConfig() : FundingSourceConfig {
             const { paylater } = fundingEligibility;
             if (
                 experiment
-                && experiment.disablePaylater === true
+                && experiment.disablePaylater
                 && !fundingSource  // Exclude standalone buttons
                 && (paylater?.products?.paylater?.variant === 'experimentable'
                     || paylater?.products?.payIn4?.variant === 'experimentable') // Only use tagged eligibility
             ) {
                 return false;
             }
-
             return true;
         },
 
--- a/src/funding/paylater/config.jsx
+++ b/src/funding/paylater/config.jsx
@@ -20,14 +20,11 @@ export function getPaylaterConfig() : FundingSourceConfig {
             BUTTON_LAYOUT.VERTICAL
         ],
 
-        eligible: ({ experiment, fundingEligibility, fundingSource }) => {
-            const { paylater } = fundingEligibility;
+        eligible: ({ experiment, fundingSource }) => {
             if (
                 experiment
                 && experiment.disablePaylater
                 && !fundingSource  // Exclude standalone buttons
-                && (paylater?.products?.paylater?.variant === 'experimentable'
-                    || paylater?.products?.payIn4?.variant === 'experimentable') // Only use tagged eligibility
             ) {
                 return false;
             }
--- a/src/zoid/buttons/util.js
+++ b/src/zoid/buttons/util.js
@@ -1,10 +1,10 @@
 /* @flow */
 import { supportsPopups as userAgentSupportsPopups, isAndroid, isChrome, isIos, isSafari, isSFVC, type Experiment, isDevice } from 'belter/src';
 import { FUNDING } from '@paypal/sdk-constants/src';
-import { getEnableFunding, createExperiment, getFundingEligibility, getPlatform, getComponents } from '@paypal/sdk-client/src';
+import { getEnableFunding, getDisableFunding, createExperiment, getFundingEligibility, getPlatform, getComponents } from '@paypal/sdk-client/src';
 import { getRefinedFundingEligibility } from '@paypal/funding-components/src';
 
-import type { Experiment as VenmoExperiment } from '../../types';
+import type { Experiment as EligibilityExperiment } from '../../types';
 import { BUTTON_FLOW, CLASS } from '../../constants';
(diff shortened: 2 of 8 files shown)


src/funding/paylater/config.jsx
```diff
@@ -20,6 +20,21 @@ export function getPaylaterConfig() : FundingSourceConfig {
             BUTTON_LAYOUT.VERTICAL
         ],
 
+        eligible: ({ experiment, fundingEligibility, fundingSource }) => {
+            const { paylater } = fundingEligibility;
+            if (
```
westeezy, line 25: Minor nitpick but for readability sake, and if code is not in hotpath, I sometimes will make some functions such as `hasTaggedEligibility` or `hasStandaloneButtons` to make these conditionals a tad easier to glance over. The comments you added are great though so thanks for those! Just sharing an alternative we might explore later. No change required

src/zoid/buttons/util.js
```diff
@@ -115,6 +115,54 @@ export function getVenmoExperiment(experiment : ?Experiment) : VenmoExperiment {
...
+        const isDisableFundingPaylater = disableFunding && disableFunding.indexOf(FUNDING.PAYLATER) !== -1;
+        const enableFunding = getEnableFunding();
+        const isEnableFundingPaylater = enableFunding && enableFunding.indexOf(FUNDING.PAYLATER) !== -1;
+
+        const fundingEligibility = getFundingEligibility();
+        const isEligibleForPaylater = fundingEligibility && fundingEligibility[FUNDING.PAYLATER] && fundingEligibility[FUNDING.PAYLATER].eligible;
+
+        if (isDevice()) {
+            // No experiment because ineligible, already forced on or off, unsupported browser
+            if (!isEligibleForPaylater
```
westeezy, line 130: I think this is probably a better example where the function calls might help cleanup the traceabiltiy aspect of what a 'complex' conditional is testing for.

westeezy (approved):
lgtm. just some minor comments for discussion. not blocking

src/zoid/buttons/util.js
ssono: I agree that it isn't pretty, but it follows the convention set by the venmoExperiment. I don't think that creating a function is really necessary tbh

src/funding/paylater/config.jsx
ssono: Hi, that's definitely a good point. One thing I'm trying to keep in mind is that this code is shipped to the client so anything to minimize the bundle size would be ideal. I think if we don't absolutely need a helper it would be best to avoid it. But maybe my understanding of how additional helper functions impact build size is wrong

src/zoid/buttons/util.js
gregjopa: I think these eligibility checks need to apply to both conditions (isDevice() is true or false). Venmo has some unique rules for desktop eligibility that we should not apply here.

src/funding/paylater/config.jsx
westeezy: In my opinion, I think readability should be a major focus as should keeping code tight. A small number of extra lines is worth understanding/maintainability to me. I am, however, certainly glad its on your mind to keep our bundle sizes in check! I'm gonna mark this as resolved as I think it's an out of scope conversation for the changeset, but my intent was to get this on folks minds as a discussion and thank you so much for sharing thoughts with me! Happy to continue a discussion in the future!

ssono:
retest this please

ssono:
All tests passing on local
Screen Shot 2021-09-07 at 1 58 58 PM

mnicpt:
@ssono It looks like your new tests aren't covering a good portion of your new code in `util.js` and `/paylater/config.jsx`. Can write some tests to cover these new scenarios?

mnicpt (changes requested):
Left a comment regarding Definition of Done around new code written to be resolved.

mnicpt:
@ssono You really don't want to use Venmo entirely as the template for this as it has unique logic for each native flow: `Venmo Native` and `Venmo Desktop`.

src/zoid/buttons/util.js
```diff
@@ -121,6 +121,56 @@ export function getVenmoExperiment() : VenmoExperiment {
...
+}
+
+export function getNoPaylaterExperiment() : EligibilityExperiment {
+    const experiment = createNoPaylaterExperiment();
+
+    const disableFunding = getDisableFunding();
+    const isDisableFundingPaylater = disableFunding && disableFunding.indexOf(FUNDING.PAYLATER) !== -1;
+    const isNativeSupported = isSupportedNativeBrowser();
+    const isExperimentEnabled = experiment && experiment.isEnabled();
+    if (isDevice()) {
```
mnicpt, line 163: Why is this logic here for PayLater?

src/funding/paylater/config.jsx
```diff
@@ -20,6 +20,17 @@ export function getPaylaterConfig() : FundingSourceConfig {
             BUTTON_LAYOUT.VERTICAL
         ],
 
+        eligible: ({ experiment, fundingSource }) => {
+            if (
+                experiment
+                && experiment.disablePaylater
+                && !fundingSource  // Exclude standalone buttons
```
mnicpt, line 27: `fundingSource` can be set in button stack as well.  This isn't exact.  I'm not so sure we can detect this.  @gregjopa Do you know if there is a more reliable way to detect standalone button integration?

ssono: Ok, I was following @gregjopa advice to use fundingSource to indicate standalone, but I can check if it's vertical or horizontal

src/zoid/buttons/util.js
ssono: I did not realize that the native/desktop logic was unique to venmo and thought all buttons needed to be checked for supportedNative. I will remove if all native and desktop renders can be considered equivalent

src/funding/paylater/config.jsx
mnicpt: That won't be enough either.  Will see what we can do.

src/zoid/buttons/util.js
```diff
@@ -121,6 +121,49 @@ export function getVenmoExperiment() : VenmoExperiment {
...
+    // No experiment because ineligible, already forced on or off
+    if (!isEligibleForPaylater
+        || !isExperimentable
+        || isDisableFundingPaylater
+        || isEnableFundingPaylater
+    ) {
+        return;
+    }
+
+    if (isDevice()) {
```
mnicpt, line 142: Do we need to have all of these types of disables by platform?  Or can we just have one for all?

ssono: Not sure yet, confirming with analytics, but since the logs I gave them followed this pattern I was planning on leaving it unless they confirm the generalized form

mnicpt (approved):

src/funding/paylater/config.jsx
gregjopa: Good point @mnicpt. It this context, `fundingSource` cannot be used to check if its a standalone integration. The style layout may work. Let's dig in and see how we can uniquely identify a standalone integration in this context.

ssono: Hey @gregjopa after speaking with @mnicpt offline we came to the conslusion that !fundingSource would be ok for this case since that field is only populated if the merchant has set it. Does that align with your understanding?

gregjopa: Great to hear you guys paired on this and `!fundingSource` should work! Let's move forward with it.

test/integration/tests/button/props.js
```diff
@@ -17,6 +17,112 @@ describe(`paypal button component props`, () => {
         destroyTestContainer();
     });
 
+    it('should not render a paylater button when the experiment disables it', () => {
```
gregjopa, line 20: Thanks for adding tests! Could you move them from props.js over to eligibility.js?

https://github.com/paypal/paypal-checkout-components/blob/master/test/integration/tests/button/eligibility.js

gregjopa (approved):

PR #1733, opened 2021-09-07
adding dimensions values to checkout function

aditya27marri:
Dimensions added to checkout function as a value object


src/zoid/checkout/component.jsx
```diff
@@ -232,9 +232,15 @@ export function getCheckoutComponent() : CheckoutComponent {
                 }
             },
         
-            dimensions: isDevice()
-                ? { width:  '100%', height: `${ DEFAULT_POPUP_SIZE.HEIGHT }px` }
-                : { width:  `${ DEFAULT_POPUP_SIZE.WIDTH }px`, height: `${ DEFAULT_POPUP_SIZE.HEIGHT }px` }
+            dimensions: ({ props }) => {
```
gregjopa, line 235: I don't think this is currently working since it's not part of the props object. Can you move it up into the props? That way we can leverage the zoid prop helpers. Ex:

```js
                dimensions: {
                    type:       'object',
                    required:   false,
                    decorate:   ({ props }) => {
                        if (props.dimensions) {
                            return { width: `${ props.dimensions.width }px`, height: `${ props.dimensions.height }px` };
                        } else if (isDevice()) {
                            return { width:  '100%', height: `${ DEFAULT_POPUP_SIZE.HEIGHT }px` };
                        } else {
                            return { width:  `${ DEFAULT_POPUP_SIZE.WIDTH }px`, height: `${ DEFAULT_POPUP_SIZE.HEIGHT }px` };
                        }
                    },
                    default: () => ({})
                },
```

test/integration/tests/checkout/dimensions.js
```diff
@@ -0,0 +1,35 @@
...
+    });
+
+    afterEach(() => {
+        destroyTestContainer();
+    });
+
+    it('should render checkout, then complete the payment with dimensions', () => {
+        return wrapPromise(({ expect, error }) => {
+            return runOnClick(() => {
+                return window.paypal.Checkout({
```
gregjopa, line 24: Could you use the `onRender` test hook to assert that xprops.dimensions is the expected value as the input width/height?

src/zoid/checkout/component.jsx
gregjopa: Ignore the above comment. I just learned that the `dimensions` function is part of zoid: https://github.com/krakenjs/zoid/commit/4d179e123f914cfa8cdfa902977f4ebf502c196d

src/zoid/checkout/component.jsx
```diff
@@ -232,9 +232,15 @@ export function getCheckoutComponent() : CheckoutComponent {
...
+                if (props.dimensions) {
```
gregjopa, line 236: Could we type check here to ensure `props.dimensions.width` never errors
```suggestion
                if (typeof props.dimensions === 'object') {
```

test/integration/tests/checkout/dimensions.js
gregjopa: @aditya27marri and I synced up offline. The plan is to use this approach for testing the popup size: https://github.com/paypal/paypal-checkout-components/pull/1741

test/integration/tests/checkout/dimensions.js
```diff
@@ -0,0 +1,53 @@
...
+
+        runOnClick(() => {
+            let childWindow;
+            onWindowOpen().then(win => {
+                childWindow = win;
+                // $FlowFixMe
+                const { outerWidth: width, outerHeight: height } = childWindow;
+
+                if (height !== APM_DEFAULT_POPUP_SIZE.HEIGHT) {
+                    done(`height does not match. expected ${ APM_DEFAULT_POPUP_SIZE.HEIGHT }, got ${ height }`);
```
gregjopa, line 35:
```suggestion
                    done(new Error(`height does not match. expected ${ APM_DEFAULT_POPUP_SIZE.HEIGHT }, got ${ height }`));
```

test/integration/tests/checkout/dimensions.js
```diff
@@ -0,0 +1,53 @@
...
+                }
+                if (width !== APM_DEFAULT_POPUP_SIZE.WIDTH) {
+                    done(`width does not match. expected ${ APM_DEFAULT_POPUP_SIZE.WIDTH }, got ${ width }`);
```
gregjopa, line 38:
```suggestion
                    done(new Error(`width does not match. expected ${ APM_DEFAULT_POPUP_SIZE.WIDTH }, got ${ width }`));
```

gregjopa (approved):
Looks good! 💯

PR #1735, opened 2021-09-08
chore(documentation): Add PR Template

westeezy:
First swing at a basic and generic PR template. If this looks good I'll add to all relevant public repos.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
@@ -1,0 +1,22 @@
+### Your checklist for this pull request
+🚨 Please review the [guidelines for contributing](../../CONTRIBUTING.md) to this repository.
+🚨 Please review the [guidelines for a good PR Description](https://www.pullrequest.com/blog/writing-a-great-pull-request-description/)
+
+- [ ] Ensure you have a PR description that answers: What? Why? How?
+- [ ] Link to any reference tickets if they are available. This includes Jira ticket or Github issues.
+- [ ] Include any relevant screenshots for the change set if it modifies any visual functionality.
+- [ ] Ensure you have added passing tests to cover new functionality, bug fixes, etc.
+
+### Description
+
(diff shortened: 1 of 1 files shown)


ssono (approved):
Could the what, why, how be their own sections in place of description? Feels more explicit and will be more prominent?

westeezy:
@ssono thats a great suggestion! I wanted to be careful about being too prescriptive, but that sounds like a good balance to me.

elizabethmv (approved):
nice! non blocking suggestion: I like to see steps to reproduce in the description for those that want to pull the changes down and confirm the behavior locally.

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -0,0 +1,22 @@
+### Your checklist for this pull request
+🚨 Please review the [guidelines for contributing](../../CONTRIBUTING.md) to this repository.
```
gregjopa, line 2: I think we need to pump the brakes here @westeezy. 🚨 is the police car light emoji. Reviewing the guides as a pre-req is not an emergency. We don't want flashing lights to scare off future contributors.

How about 🔬 or 🍏 instead?

bluepnume: 🦆 please

westeezy: i like a little intimidation in my prs. 🦁

westeezy:
> nice! non blocking suggestion: I like to see steps to reproduce in the description for those that want to pull the changes down and confirm the behavior locally.

@elizabethmv love it. I'll add that in

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
gregjopa: The lion is pretty sweet too 🦁. We could use that

PR #1738, opened 2021-09-09
enable experiment to animate paypal logo and a placeholder

erick2014:


src/constants/experiment-one.js
```diff
@@ -0,0 +1,5 @@
+/* @flow */
+
+export const EXPERIMENT_ONE = {
+    PLACE_HOLDER: ('experiment-one-label' : 'experiment-one-label')
```
ssono, line 4: Can we have a more descriptive name than experiment-one?

ssono:
It also looks like they are starting at different times. Can you get them to start at the same time?

src/funding/paypal/template.jsx
```diff
@@ -147,10 +147,60 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
...
+            100% {
+                transform: translateX(220px);
+                opacity: 1;
+            }
+        }
+    ` }
+    />);
+    return (
+        <Fragment>
+            <div class="experiment-one-label"> <span>Earn rewards</span></div>
```
ravishekhar, line 193: Are there localization requirements for "Earn rewards"?

ravishekhar (changes requested):

src/funding/paypal/template.jsx
erick2014: this label will come from a machine learning model prediction, but for now will be harcoded

src/constants/experiment-one.js
```diff
@@ -0,0 +1,5 @@
+/* @flow */
+
+export const EXPERIMENT_ONE = {
```
ssono, line 3: Can you change all instances of EXPERIMENT ONE to the new class name?

erick2014: sure

src/constants/index.js
```diff
@@ -3,3 +3,4 @@
 export * from './button';
 export * from './misc';
 export * from './class';
+export * from './experiment-one';
```
ssono, line 6: here

src/funding/common.jsx
```diff
@@ -46,7 +46,8 @@ export type LabelOptions = {|
     personalization : ?Personalization,
     nonce : ?string,
     tagline : ?boolean,
-    content : ?ContentType
+    content : ?ContentType,
+    enableExperimentOne : ?boolean
```
ssono, line 50: here

src/funding/paypal/template.jsx
```diff
@@ -147,10 +147,60 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
 }
 
 
+const getExperimentOneAnimation = (opts) => {
```
ssono, line 150: here

src/funding/paypal/template.jsx
```diff
@@ -147,10 +147,60 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
...
+    if (!opts.enableExperimentOne) {
```
ssono, line 151: here

src/funding/paypal/template.jsx
```diff
@@ -147,10 +147,60 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
...
+        return;
+    }
+    const experimentStyles = (<style innerHTML={ `
+        .${ CLASS.DOM_READY } .${ LOGO_CLASS.LOGO }{
+            animation: 1s experiment-one-move-logo-to-the-left 2s forwards;
```
ssono, line 156: here

src/funding/paypal/template.jsx
```diff
@@ -147,10 +147,60 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
...
+            position: relative;
+        }
+
+        .${ DIVIDE_ANIMATE_LOGO_EXPERIMENT.PLACE_HOLDER } {
+            display: inline-block;
+            position: absolute;
+            opacity: 0; 
+            animation: 1s experiment-one-move-placeholder-to-the-right 2s forwards;
```
ssono, line 164: here

src/funding/paypal/template.jsx
```diff
@@ -147,10 +147,60 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
...
+        }
+
+        .${ CLASS.BUTTON_LABEL } .${ DIVIDE_ANIMATE_LOGO_EXPERIMENT.PLACE_HOLDER } span {
+            font-size: 14px;
+            color: #142C8E;
+            padding-top: 3px;
+            font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+        }
+
+        @keyframes experiment-one-move-logo-to-the-left {
```
ssono, line 174: here

src/funding/paypal/template.jsx
```diff
@@ -147,10 +147,60 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
...
+            100% {
+                transform: translateX(-300px)
+            }
+        }
+        
+        @keyframes experiment-one-move-placeholder-to-the-right {
```
ssono, line 180: here

src/funding/paypal/template.jsx
```diff
@@ -147,10 +147,60 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
...
+            { experimentStyles }
+        </Fragment>
+    );
+};
+
 export function Label(opts : LabelOptions) : ChildType {
     return (
         <Fragment>
             <BasicLabel { ...opts } />
+            {getExperimentOneAnimation(opts)}
```
ssono, line 203: here

src/ui/buttons/button.jsx
```diff
@@ -112,6 +112,7 @@ export function Button({ fundingSource, style, multiple, locale, env, fundingEli
 
     let labelNode = (
         <Label
+            enableExperimentOne={ false }
```
ssono, line 115: here

src/constants/experiment-one.js
```diff
@@ -1,5 +1,5 @@
 /* @flow */
 
-export const EXPERIMENT_ONE = {
+export const DIVIDE_ANIMATE_LOGO_EXPERIMENT = {
```
ssono, line 3: here

src/constants/experiment-one.js
```diff
@@ -1,5 +1,5 @@
...
     PLACE_HOLDER: ('divide-animate-logo-experiment' : 'divide-animate-logo-experiment')
```
ssono, line 4: here

src/funding/common.jsx
ssono: Is there a reason this one was left?

ssono (approved):
LGTM theres one thing that might not pass lint, but it does not block getting more reviews

src/funding/paypal/template.jsx
```diff
@@ -126,7 +126,7 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
 
     const personalizationText = getPersonalizationText(opts);
     const personalizationTracker = getPersonalizationTracker(opts);
-
+
```
ssono, line 129: Does this pass lint with the extra spaces there?

erick2014:
> LGTM theres one thing that might not pass lint, but it does not block getting more reviews

I just pushed that change, thanks

src/funding/paypal/template.jsx
```diff
@@ -147,10 +146,52 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
 }
 
 
+const getTagLineForDivideLogoAnimationExperiment = (opts) => {
```
ssono, line 149: Hmmm should this be named tagline? Tagline normally refers to the text below the button. Could we call it labelText or something like that?

src/funding/paypal/template.jsx
```diff
@@ -147,10 +146,52 @@ function ButtonPersonalization(opts : LabelOptions) : ?ChildType {
...
+        return;
+    }
+
+    const experimentStyles = (<style innerHTML={ `
+        .${ CLASS.DOM_READY } .${ LOGO_CLASS.LOGO }{
+            animation: 1s divide-logo-animation-experiment-left-side 2s forwards;
+            position: relative;
+        }
+
+        .${ DIVIDE_LOGO_ANIMATION_EXPERIMENT.PLACE_HOLDER } {
```
ssono, line 160: What does PLACE_HOLDER refer to?

src/ui/buttons/script.jsx
```diff
@@ -145,6 +144,46 @@ function getComponentScript() : () => void {
             }
         }
 
+        function addDivideLogoAnimationExperiment() {
+            const buttonElement = document ? document.querySelector('.paypal-button-label-container') : null;
```
ssono, line 148: Instead of using a Ternary could you do this instead? Or is it important that the default value is null?
`const buttonElement = document && document.querySelector('.paypal-button-label-container');`

ssono (commented):
Left some comments on naming and code style. I have some questions on the overall design of this experiment.

From the overall code design perspective, I think the UI components in checkout components should not know about the experiment. Instead they should render according to a flag/animationId. Experimentation decision could be made in the `checkoutCustomization` call in xobuyernodeserv and then the proper animation should be applied based on that. 

See these PRs for a reference

src/ui/buttons/script.jsx
erick2014: it does not matter if it is null or a boolean

ssono: Using && is easier to read than a ternary so we generally avoid ternaries unless they are necessary

src/ui/buttons/buttonAnimations.js
```diff
@@ -0,0 +1,116 @@
...
+    return {
+        'params': {
+            large: { min: BUTTON_SIZE_STYLE.large.minWidth, max: BUTTON_SIZE_STYLE.large.maxWidth },
+            huge:  { min: BUTTON_SIZE_STYLE.huge.minWidth, max: BUTTON_SIZE_STYLE.huge.maxWidth }
+        },
+        'fn': animation
+    };
+};
+
+export const getAnimationLabelAndStyles = (opts) => {
```
ssono, line 77: Can you set this function to work generically when there are multiple animations available? Instead of checking options and returning DivideLogoAnimation you could extract to check if there is an `enabledAnimation` and then based on that value return the correct result? You could abstract out the specific animation stuff into constants like 
`
export const ANIMATIONS = {
    animation1: {
        labelText:  'Earn rewards',
        labelClass:  DIVIDE_LOGO_ANIMATION.LABEL_TEXT,
        {styles}
    },
    animation2: {
        labelText:  'Earn rewards',
        labelClass:  DIVIDE_LOGO_ANIMATION.LABEL_TEXT,
        {styles}
    }
}
`

ssono (commented):
Looks great. I have one code design request but otherwise good

src/ui/buttons/buttonAnimations.js
erick2014: @ssono  I change the structure to make it more flexible, please check!

src/ui/buttons/script.jsx
```diff
@@ -166,11 +163,21 @@ function getComponentScript() : () => void {
...
+    nonce : ?string,
+    // eslint-disable-next-line flowtype/no-weak-types
+    buttonAnimation : any
 |};
 
-export function Script({ nonce } : ScriptProps) : ElementNode {
+export function Script({ nonce, buttonAnimation } : ScriptProps) : ElementNode {
+    const htmlForScript = `
+        const scriptFns = ${ getComponentScript().toString() }
+        const animation = ${ buttonAnimation && buttonAnimation.fn ? buttonAnimation.fn.toString() : null }
```
ssono, line 174: Doesn't need to be a ternary. Only use a ternary when you have relevant return values for both cases
(buttonAnimation && buttonAnimation.fn) || buttonAnimation.fn.toString()

src/ui/buttons/script.jsx
```diff
@@ -166,11 +163,21 @@ function getComponentScript() : () => void {
...
+        scriptFns()
+        if (animation) {
+            animation(${ buttonAnimation && buttonAnimation.params ?  JSON.stringify(buttonAnimation.params) : null })
```
ssono, line 177: Same as above, probably shouldn't be a ternary

src/ui/buttons/script.jsx
```diff
@@ -166,11 +163,21 @@ function getComponentScript() : () => void {
...
 type ScriptProps = {|
-    nonce : ?string
+    nonce : ?string,
+    // eslint-disable-next-line flowtype/no-weak-types
+    buttonAnimation : any
 |};
 
-export function Script({ nonce } : ScriptProps) : ElementNode {
+export function Script({ nonce, buttonAnimation } : ScriptProps) : ElementNode {
+    const htmlForScript = `
```
ssono, line 172: Can you pull some of this js logic out before defining the htmlForScript? Or add comments explaining what i going on because it is not very readable

src/ui/buttons/buttons.jsx
```diff
@@ -189,6 +190,7 @@ export function Buttons(props : ButtonsProps) : ElementNode {
             }
 
             <Script
+                buttonAnimation={ createAddDivideLogoAnimation() }
```
ssono, line 193: This is still not generic really. I think what you'd want is `createLogoAnimation()` here and have internal logic to decide which animation gets returned.

ssono (commented):
I don't know if the changes have made this more generic. It is still checking in the component for the divideLogoLabel. Imagine you have multiple possible animations that you could receive from personalization. In this case, what would happen is that `Buttons` would receive an animation ID in the personalization object here https://github.com/paypal/paypal-checkout-components/blob/2e5458d6ed9fa3091f703f01742c55a104bebcea/src/ui/buttons/button.jsx#L44

Then based on the animationID would render the correct animation. There should be no explicit reference to a specific animation in either `button.jsx` or `buttons.jsx`

src/ui/buttons/props.js
```diff
@@ -163,6 +163,10 @@ export type Personalization = {|
             impression : string,
             click : string
         |}
+    |},
+    buttonAnimation : {|
```
ssono, line 167: Should be optional and match other personalization types
`
buttonAnimation? : {|
        id : string,
        labelText : string,
        tracking : {|
            impression : string,
            click : string
        |}
    |}
    `

src/ui/buttons/script.jsx
```diff
@@ -166,11 +163,25 @@ function getComponentScript() : () => void {
...
 
-export function Script({ nonce } : ScriptProps) : ElementNode {
+export function Script({ nonce, buttonAnimation } : ScriptProps) : ElementNode {
+    const scripts = `
+        const scriptFns = ${ getComponentScript().toString() }
+        scriptFns()
+        function onDomLoad(){
+            const animation = ${ buttonAnimation && buttonAnimation.fn ? buttonAnimation.fn.toString() : null }
+            if (animation) {
+                animation(${ buttonAnimation && buttonAnimation.params ?  JSON.stringify(buttonAnimation.params) : null })
```
ssono, line 178: Why does this need to be a ternary instead of || ?

src/ui/buttons/button.jsx
```diff
@@ -8,7 +8,7 @@ import { LOGO_COLOR, LOGO_CLASS } from '@paypal/sdk-logos/src';
 import { noop, preventClickFocus, isBrowser, isElement } from 'belter/src';
 
 import type { ContentType, Wallet, Experiment, WalletInstrument } from '../../types';
-import { ATTRIBUTE, CLASS, BUTTON_COLOR, BUTTON_NUMBER, TEXT_COLOR, BUTTON_FLOW } from '../../constants';
+import { ATTRIBUTE, CLASS, DIVIDE_LOGO_ANIMATION, BUTTON_COLOR, BUTTON_NUMBER, TEXT_COLOR, BUTTON_FLOW } from '../../constants';
```
ssono, line 11: is `DIVIDE_LOGO_ANIMATION` being used anywhere?

src/constants/divide-logo-animation.js
```diff
@@ -0,0 +1,6 @@
+/* @flow */
```
ssono, line 1: Is this being used anywhere?

src/constants/index.js
```diff
@@ -3,3 +3,4 @@
 export * from './button';
 export * from './misc';
 export * from './class';
+export * from './divide-logo-animation';
```
ssono, line 6: Is this being used?

src/funding/paypal/template.jsx
```diff
@@ -151,6 +150,7 @@ export function Label(opts : LabelOptions) : ChildType {
     return (
         <Fragment>
             <BasicLabel { ...opts } />
+            {opts.buttonAnimation &&  opts.buttonAnimation.animationComponent ? opts.buttonAnimation.animationComponent : null }
```
ssono, line 153: Does this need to be a ternary?

src/ui/buttons/button-animations/divide-logo-animation.jsx
```diff
@@ -0,0 +1,139 @@
...
+                PAYPAL_LOGO:                LOGO_CLASS.LOGO,
+                ANIMATION_LABEL_CONTAINER:        ANIMATION.LABEL_CONTAINER,
+                PAYPAL_BUTTON_LABEL:        CLASS.BUTTON_LABEL
+            }
+        },
+        'fn': animation
+    };
+}
+
+export function setupAnimation (animationLabelText : string) : void {
```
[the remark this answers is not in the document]
ssono: Maybe add an animation type

Roselynalan (approved):

src/ui/buttons/script.jsx
erick2014: I think we need the Ternary because we need to convert the function into string and the || buttonAnimation.fn.toString() only gets evaluated when (buttonAnimation && buttonAnimation.fn) is false

PR #1741, opened 2021-09-15
Add test for default dimensions

gregjopa:
This PR adds a test for the default dimensions of the popup.

### DIFF
--- a/test/integration/tests/checkout/happy.js
+++ b/test/integration/tests/checkout/happy.js
@@ -204,10 +204,10 @@ describe(`paypal checkout component happy path`, () => {
                 const { outerWidth: width, outerHeight: height } = childWindow;
 
                 if (height !== DEFAULT_POPUP_SIZE.HEIGHT) {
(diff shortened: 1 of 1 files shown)


test/integration/tests/checkout/happy.js
```diff
@@ -189,4 +189,36 @@ describe(`paypal checkout component happy path`, () => {
...
+
+    it('should render checkout with default dimensions', (done) => {
+        const DEFAULT_POPUP_SIZE = {
+            WIDTH:  500,
+            HEIGHT: 590
+        };
+
+        runOnClick(() => {
+            let childWindow;
+            onWindowOpen().then(win => {
```
gregjopa, line 201: The goal is to assert on the width and height of the popup. Currently it seems like `onWindowOpen` along with `done()` will solve it. But I'm definitely open to other suggestions. I tried using Promises but couldn't get it to error from the `onWindowOpen` callback.

PR #1745, opened 2021-09-16
Add github action for publishing to npm

gregjopa:
### Description

Add a new action responsible for publishing this package to npm.

#### What is being changed from a technical perspective?

The `npm run release` command is being run with GitHub Actions. This action securely publishes to npm using an automation token. This action is not automatically run. Instead it is manually run using the workflow dispatch feature which is only available to employees with elevated GitHub permissions on this repo.

#### Why are we making these changes? Include and reference tickets (Jira, Github Issue, etc)

Because deploying with GitHub Actions is easier than deploying locally in your terminal.


westeezy (approved):
🔥

PR #1747, opened 2021-09-16
Venmo Desktop - Consumer Feedback collection

andregardi:
### Description
Feedback on Desktop XO would need to be collected at only one point on the QR code scanning page when the consumer hits the X option.  We do not want to collect the feedback post the QR scan because we are already collecting the feedback on the mobile experience.

#### Linked PR:
https://github.com/paypal/paypal-smart-payment-buttons/pull/247

#### Changes
- Passing the onSubmitFeedback event as a props to the QRCodeComponent.
- After the iframe is loaded, we remove the close button. A new identical close button is rendered on the QRCodeComponent that controls the survey.

### DIFF
--- a/src/zoid/qr-code/component.jsx
+++ b/src/zoid/qr-code/component.jsx
@@ -73,6 +73,9 @@ export function getQRCodeComponent() : QRCodeComponent {
                 onEscapePath: {
                     type: 'function'
                 },
+                onSubmitFeedback: {
+                    type: 'function'
+                },
                 qrPath: {
                     type:       'string',
                     queryParam: true,
@@ -142,6 +145,9 @@ export function QRCodeContainer({
 
         frame.classList.remove(CLASS.INVISIBLE);
         frame.classList.add(CLASS.VISIBLE);
+
+        // Close is now controlled by QRCard on smart-payment-buttons
+        document.querySelector('#close').remove();
 
         setTimeout(() => {
             destroyElement(prerenderFrame);


src/zoid/qr-code/component.jsx
```diff
@@ -73,6 +73,9 @@ export function getQRCodeComponent() : QRCodeComponent {
                 onEscapePath: {
                     type: 'function'
                 },
+                onSubmitFeedback: {
```
remotevision, line 76: @andregardi - let's be explicit in setting `required` to `true` or `false`.

andregardi: Sure @remotevision !
For the moment I am keeping `required` as `false` for that.
I just tried making it true, and it does not play well with older versions of smart-payment-buttons.

mnicpt: @andregardi But it is set to `true` currently.  Can you set it to `false`?

mnicpt:
@andregardi Why are we doing any of these: 
```
- Passing the onSubmitFeedback event as a props to the QRCodeComponent.
- After the iframe is loaded, we remove the close button. A new identical close button is rendered on the QRCodeComponent that controls the survey.
```

src/zoid/qr-code/component.jsx
```diff
@@ -185,7 +195,7 @@ export function QRCodeContainer({
                 box-shadow: 0px 0px 8px rgba(0, 0, 0, 0.4);
                 border-radius: 16px;                        
                 width: 720px;
-                height: 544px;  
+                height: 612px;
```
mnicpt, line 198: @andregardi Why are we changing the dimensions of the QRCodeContainer?

andregardi:
> @andregardi Why are we doing any of these:
> 
> ```
> - Passing the onSubmitFeedback event as a props to the QRCodeComponent.
> - After the iframe is loaded, we remove the close button. A new identical close button is rendered on the QRCodeComponent that controls the survey.
> ```

This is the result of this slack Thread:
https://paypal.slack.com/archives/C015CRRV23Y/p1631276747156800

To sumarize, on the first time we click the close button, it does not actually close the modal, but opens a survey.
Because we can't emit events from the `containerTemplate` (on **paypal-checkout-components**) to the child iframe, Daniel suggested to move the close button to the `paypal-smart-payment-buttons`.

At same time we need to keep the close button here while the iframe content has not ready, because it would not allow the user to close the modal while it is loading.

src/zoid/qr-code/component.jsx
andregardi: We had a redesign last week on the size of the modal:
https://www.figma.com/file/g9Pg0vQJK0EVeiBLiuMoC2/Universal-Venmo-Pay?node-id=7269%3A33206

Now that I am revisiting that, this new modal height is only `612px` if the survey is displayed. I am sorry for that, I got confused because the design kept changing. I am not sure yet how could we implement this dynamic height controlled by the child iframe. If you have any suggestion on that, let me know!

src/zoid/qr-code/component.jsx
andregardi: Sorry for that!
I fixed it now.

src/zoid/qr-code/component.jsx
```diff
@@ -143,6 +147,12 @@ export function QRCodeContainer({
         frame.classList.remove(CLASS.INVISIBLE);
         frame.classList.add(CLASS.VISIBLE);
 
+        // Close is now controlled by QRCard on smart-payment-buttons
+        const closeButton = document.querySelector('#close');
```
mnicpt, line 151: Just remove this.

src/zoid/qr-code/component.jsx
bluepnume: Not sure why we need `onSubmitFeedback`? Can't we just close the modal, does the parent need to care about whether feedback was sent or not?

src/zoid/qr-code/component.jsx
```diff
@@ -143,6 +147,12 @@ export function QRCodeContainer({
...
+        if (closeButton) {
```
bluepnume, line 152: This is gonna be true 100% of the time right? Why not just remove the close button from this template entirely and just not render it in the first place?

src/zoid/qr-code/component.jsx
mnicpt: We don't need this either so you can remove.

src/zoid/qr-code/component.jsx
mnicpt: Style it so that there is a min-height and max-height and let the container expand for survey.

src/zoid/qr-code/component.jsx
```diff
@@ -208,7 +218,7 @@ export function QRCodeContainer({
                 position: absolute;
                 left: 8px;
                 content: ' ';
-                height: 16px;
+                height: 20px;
```
mnicpt, line 221: Remove the close styling.

src/zoid/qr-code/component.jsx
andregardi: image
@bluepnume 
While we are on the loading state the iframe component is not ready to be displayed. We need someway to close the modal during that time or even if we have an error on loading the child component.

bluepnume: Sure, then can we put the close button in the prerender template? The prerender will be automatically removed when the full component is rendered.

`document.querySelector` isn't very safe here, as you could end up removing elements from the merchant page this way if they have the same ID

src/zoid/qr-code/component.jsx
andregardi: @mnicpt 
I think that make sense if we decide to entirely remove the close button from checkout-components.
I replied the bellow comment from Daniel, explain why maybe we should remove it only after the child iframe is loaded.
If we decide to move forward with that, I will be happy to remove the close button on this PR.

src/zoid/qr-code/component.jsx
andregardi: Sure, know that you are saying, it makes complete sense to not use `document.querySelector`.
I guess moving it to prerender template is a great idea!

src/zoid/qr-code/component.jsx
andregardi: Since we are moving the close button to the `prerenderTemplate`, maybe we will need to keep this style. I will need to work a bit on the code try check that. Is that ok?

src/zoid/qr-code/component.jsx
andregardi: That is great, I will handle this event entirely on the component side.

src/zoid/qr-code/component.jsx
andregardi: Thank you for the suggestion @mnicpt ,
But that is not so easy in that case.
Between the `containerTemplate` and the `QRCard` there is an iframe. I believe that there is no CSS property that allow the iframe to resize according to its content. That means, expand the survey or any other container on SPB would not affect the size of the `containerTemplate` here.

src/zoid/qr-code/component.jsx
```diff
@@ -33,15 +32,52 @@ export function getQRCodeComponent() : QRCodeComponent {
...
+                    position: absolute;
+                    right: 16px;
+                    top: 16px;
+                    width: 16px;
+                    height: 0;
+                    opacity: 0.6;
+                    z-index: 10;
+                    padding: 0;
+                    border: none;
+                    cursor: pointer;
```
gregjopa, line 47: I think `<button>` uses a pointer by default. Can you check? We should keep these styles down to the bare minimum.
```suggestion
```

src/zoid/qr-code/component.jsx
gregjopa: @andregardi can you remove this line? Based on the convo, it seems like prerender is handlnig it.

PR #1749, opened 2021-09-17
Add section to PR Template to describe dependent changes

gregjopa:
### Description

Update the PR template to include a section to call out dependent changes so we can better coordinate production deployments and ensure code changes account for unexpected input.

#### What is being changed from a technical perspective?

We experienced an issue yesterday where a new prop was only partially deployed while being marked as required.


elizabethmv (approved):

westeezy (approved):
lgtm. lets update in SPB too :D

PR #1761, opened 2021-10-08
Resize PayPal button animation two ( experiment one )

erick2014:
### Description
This animation is part of a series of animations that we are going to implement in order to try to increase the click rating. 

In this second animation we resize the PayPal button and add a blue background and also a label inside it. The animation was implemented using keyframes only. We are not loading extra resources, just creating a stylesheet  that includes keyframes inside, then we append this to the dom.

To increase click rating
https://engineering.paypalcorp.com/jira/browse/DTUPSTREAM-68

https://user-images.githubusercontent.com/7244087/136867965-857ec942-0ff9-4d7a-9818-fab46a516eec.mov

https://github.com/paypal/paypal-checkout-components/pull/1738/files


mnicpt:
See comments here: https://github.com/paypal/paypal-checkout-components/pull/1767.

PR #1765, opened 2021-10-13
chore: update bundlesize tracking to use bundlemon

westeezy:
### Description

Use [bundlemon](https://github.com/LironEr/bundlemon) to enable richer bundle size tracking. Our intent is to move to a percentage based increase model.

We run into the scenario of updating our bundlesize budget pretty consistently and its output is hidden away in the build. We would like output to be first class on new PRs as well as track via a percent increase.

❤️  Thank you!

### DIFF
--- a/.bundlemonrc.json
+++ b/.bundlemonrc.json
@@ -1,0 +1,21 @@
+{
+  "baseDir": "./dist",
+  "files": [
+    {
+      "path": "size.min.js",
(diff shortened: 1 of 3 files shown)


westeezy:
merging with failing check as it might be secrets related. chatted offline w/ @gregjopa related to this. If any ill effects we will revert.

PR #1781, opened 2021-11-01
Fix BundleMon configuration

LironEr:
### Description
Currently, BundleMon runs on pull requests & commits to master.
When running on a branch there is no problem and BundleMon creates check run & commit status without a problem.
But when running BundleMon on PR you need to override `CI_COMMIT_SHA` env var to point BundleMon to the right commit SHA, otherwise, you won't see the report. You can read more about it [here](https://frontside.com/blog/2020-05-26-github-actions-pull_request/#how-does-pull_request-affect-actionscheckout)

Also because GitHub check run creates commit status there is no need to use both, otherwise, you will have 2 statues.
image

### DIFF
--- a/.bundlemonrc.json
+++ b/.bundlemonrc.json
@@ -13,7 +13,7 @@
       "github",
       {
         "checkRun": true,
-        "commitStatus": true,
+        "commitStatus": false,
         "prComment": false
       }
     ]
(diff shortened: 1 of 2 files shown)


westeezy:
Oh this is awesome! Thanks for explaining it!

westeezy (approved):

PR #1790, opened 2021-11-09
Fix destructuring bug

ssono:
### Description

- Bug when destructuring a null value
- Fixed by checking falsy before destructuring

❤️  Thank you!

### DIFF
--- a/src/ui/buttons/button-animations/index.js
+++ b/src/ui/buttons/button-animations/index.js
@@ -8,8 +8,14 @@ import { type ButtonAnimationOutputParams } from './types';
 
 
 export function getButtonAnimation(personalization : ?Personalization) : ButtonAnimationOutputParams | Object {
-    // Only show animations for SSR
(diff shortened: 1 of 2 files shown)


src/ui/buttons/button-animations/index.js
```diff
@@ -9,7 +9,7 @@ import { type ButtonAnimationOutputParams } from './types';
 
 export function getButtonAnimation(personalization : ?Personalization) : ButtonAnimationOutputParams | Object {
     // Only show animations for SSR
-    if (__WEB__ || !personalization) {
+    if (__WEB__ || !personalization || !personalization.buttonAnimation) {
```
gregjopa, line 12: What do you think about moving this validation to a variable? Here's what I'm thinking:
```js
const isValidPersonalization = personalization && personalization === 'object' && personalization.buttonAnimation;
if (__WEB__ || !isValidPersonalization)  {   
    return {};
}
```

PR #1791, opened 2021-11-09
paymentfields prerender bg color update

andyfen:
adds subtle gray background to prerender stage for payment fields

before (empty)
Screen Shot 2021-11-09 at 1 35 04 PM

doc example gray bg
image001

on white 
image001

### DIFF
--- a/src/zoid/payment-fields/prerender.jsx
+++ b/src/zoid/payment-fields/prerender.jsx
@@ -19,35 +19,12 @@ export function PaymentFieldsPrerender({ nonce } : PrerenderedPaymentFieldsProps
                             margin: 0;
                             width: 100%;
                             height: 100%;
-                            background: transparent;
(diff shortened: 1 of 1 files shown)


src/zoid/payment-fields/prerender.jsx
```diff
@@ -19,36 +19,13 @@ export function PaymentFieldsPrerender({ nonce } : PrerenderedPaymentFieldsProps
                             margin: 0;
                             width: 100%;
                             height: 100%;
-                            background: transparent;
+                            background: rgba(240,240,240,0.5);
```
gregjopa, line 22: Is the merchant able to control the background color here when `transparent` is used?

andyfen: yes. background color is the page

aditya27marri (approved):

PR #1792, opened 2021-11-10
[VMORECOMBO-345]- Venmo App Label

fedperezp:
### Description

As a Venmo buyer, I see "Venmo App" mentioned on the Venmo button and I am given greater context so I know to expect an app switch.
As a low intent Venmo buyer, I see "Venmo app" mentioned on the Venmo button and I am given more information about the transaction experience with Venmo and am less curious about experiencing the app switch without following through.

https://engineering.paypalcorp.com/jira/browse/DTCHKINT-1032

| | |
| ---- | ---- |
image Pay With `horizontal` | image Pay With `vertical` |
| image Donate `Horizontal` | image Donate `Vertical` |
| image Checkout `Horizontal` | image Checkout `Vertical` |
| image Buy Now `horizontal` | image Buy Now `vertical` |

❤️  Thank you!


src/zoid/buttons/util.js
```diff
@@ -135,6 +135,12 @@ export function getNoPaylaterExperiment(fundingSource : ?$Values<typeof FUNDING>
     };
 }
 
+export function getVenmoAppLabelExperiment() : EligibilityExperiment  {
+    return {
+        enableVenmoAppLabel: true
+    };
```
fedperezp, line 141: hardcoded until passing a param from elmo exp

src/zoid/buttons/util.js
```diff
@@ -135,6 +135,12 @@ export function getNoPaylaterExperiment(fundingSource : ?$Values<typeof FUNDING>
     };
 }
 
+export function getVenmoAppLabelExperiment() : EligibilityExperiment  {
+    return {
+        enableVenmoAppLabel: false
```
mnicpt, line 140: Hardcoded blocker so we don't forget about it

mnicpt (changes requested):

src/funding/venmo/template.jsx
```diff
@@ -3,10 +3,37 @@
...
+        <Style css={ css }>
+            <VenmoLogo logoColor={ logoColor } />
+            <Text className={ [ 'app-label' ] }>
+                App
+            </Text>
+        </Style>
+    );
+
+    return (
+        <BasicLabel { ...opts } logo={ AppLogo } />
```
mnicpt, line 28: Nice solution to allow reuse of tagline!

mnicpt:
@fedperezp Looking good! Just have a blocker to ensure we update the hardcoded placeholder.

src/zoid/buttons/component.jsx
```diff
@@ -20,7 +20,7 @@ import { isFundingEligible } from '../../funding';
 import { containerTemplate } from './container';
 import { PrerenderedButtons } from './prerender';
 import { applePaySession, determineFlow, isSupportedNativeBrowser, createVenmoExperiment,
-    getVenmoExperiment, createNoPaylaterExperiment, getNoPaylaterExperiment, getRenderedButtons } from './util';
+    getVenmoExperiment, createNoPaylaterExperiment, getNoPaylaterExperiment, getRenderedButtons, getVenmoAppLabelExperiment } from './util';
```
mnicpt, line 23: Can we reduce these three experiments to a single collection of them and operate on them later?

mnicpt (changes requested):

src/funding/venmo/template.jsx
fedperezp: Thanks!

src/zoid/buttons/util.js
fedperezp: Steve, I forgot to comment on this, the intention is to display the "App" label during the final rendering of the component, with that in mind the `experiment.enableVenmoAppLabel` that counts is the one coming from scnw -> pspb and then forwarded to the zoid component. Let me know if the implementation is correct in that sense or if we should add a new prop for this experiment (for your comments on doing the `getExperiments` thing I'm assuming we want to keep it as one prop)

fedperezp: This is the desired behaviour 
https://user-images.githubusercontent.com/89796119/142225790-ebe5260a-b2a2-4ba4-87dd-46bba3535a06.mov

src/zoid/buttons/component.jsx
fedperezp: +1, pushed that change

src/zoid/buttons/util.js
mnicpt: It looks like the `App` suffix should be positioned lower.

src/funding/venmo/style.scoped.scss
```diff
@@ -1,4 +1,7 @@
 
+.app-label {
+    margin-left: 0.25rem;
```
mnicpt, line 3: Looks like it needs to be positioned down a bit as well.

src/zoid/buttons/component.jsx
```diff
@@ -20,7 +20,7 @@ import { isFundingEligible } from '../../funding';
 import { containerTemplate } from './container';
 import { PrerenderedButtons } from './prerender';
 import { applePaySession, determineFlow, isSupportedNativeBrowser, createVenmoExperiment,
-    getVenmoExperiment, createNoPaylaterExperiment, getNoPaylaterExperiment, getRenderedButtons } from './util';
+    createNoPaylaterExperiment, getRenderedButtons, getButtonsExperiments } from './util';
```
mnicpt, line 23: Can we change to `getButtonExperiments`?

src/zoid/buttons/component.jsx
```diff
@@ -383,10 +380,7 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 queryParam: true,
                 value:      ({ props }) => {
                     const { fundingSource } = props;
-                    const experimentTreatments = {
-                        ...getVenmoExperiment(),
-                        ...getNoPaylaterExperiment(fundingSource)
-                    };
+                    const experimentTreatments = getButtonsExperiments(fundingSource);
```
mnicpt, line 383: Can we rename this to `experiments`?

src/zoid/buttons/util.js
mnicpt: > 

Setting this false to indicate that it occurs on second render is not clear here.  Can we not make a decision in it's prerender to ignore?

src/zoid/buttons/component.jsx
fedperezp: sure!

src/funding/venmo/style.scoped.scss
fedperezp: I think you're right, going to take this back with design because the result was approved but it looks higher than the venmo logo

mnicpt (approved):

TNuzzi (approved):

PR #1793, opened 2021-11-11
Generalize button personalization design

ssono:
### Description

- Refactor code for existing button design experiments to be more reusable
- Handle resizing better
- Code is written on the assumption that we can get button size on prerender
- added data-design-experiment tag for more accurate logging

https://engineering.paypalcorp.com/jira/browse/DTUPSTREAM-185

❤️  Thank you!


src/ui/buttons/buttonDesigns/buttonDesignComponent.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+import { CLASS } from '../../../constants';
+
+import { type LabelOptions } from './types';
+
+export function ButtonDesignComponent({ designLabelText } : LabelOptions) : ChildType {
+    return (
+        <Fragment>
+            <div class={ 'personalized-label-container' } data-animation-experiment="Varied_Button_Design"> <span>{designLabelText}</span></div>
+            <style innerHTML={ `
```
erick2014, line 14: This style tag should be optional( since the styles can be created at keyframes creation time ), and would be helpful to receive it as a property in case we need them. 
Take a look at this animation, where I was able to remove the styles from here, and I add them within the keyframes creation script.
Screen Shot 2021-11-15 at 12 55 54 PM

src/ui/buttons/buttonDesigns/inlineLogoTextDesign.jsx
```diff
@@ -0,0 +1,93 @@
...
+                  color: #142C8E;
+                  font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+                  font-size: 14px;
+              }
+
+              .personalized-design-container .personalized-label-container span {
+                  display: flex;
+                  flex-direction: column;
+                  justify-content: space-around;
+              }
```
erick2014, line 89: I think all these styles can be integrated inside `getInlineLabelTextDesign` and you will get the same result.

ssono: They definitely can be moved, but I'm not sure I understand what the advantage of doing this would be?

src/ui/buttons/buttonDesigns/buttonDesignComponent.jsx
ssono: They definitely can be moved, but I'm not sure I understand what the advantage of doing this would be?

src/ui/buttons/buttonDesigns/divideLogoAnimation.jsx
```diff
@@ -0,0 +1,154 @@
...
+    const paypalLabelContainerElement = designContainer.querySelector(`.${ PAYPAL_BUTTON_LABEL }`) || null;
+    if (!paypalLabelContainerElement) {
+        return null;
+    }
+
+    // get starting position for element so it doesn't flicker when animation begins
+    const paypalLogoElement = (paypalLabelContainerElement && paypalLabelContainerElement.querySelector(`.${ PAYPAL_LOGO }`)) || null;
+    const paypalLogoStartingLeftPosition = paypalLogoElement
+        ? `${ (paypalLogoElement.offsetLeft / paypalLabelContainerElement.offsetWidth) * 100 }%`
+        : '44.5%';
```
jrod1032, line 53: Nit: where does this number come from? a comment can help

src/ui/buttons/buttonDesigns/inlineLogoTextDesign.jsx
ssono: After looking into this more, it is better to keep them separate for the resize use case

ssono: After looking into this more, it is better to keep them separate for the resize use case

jrod1032 (approved):
Small nit, but LGTM

src/funding/paypal/template.jsx
```diff
@@ -164,16 +164,16 @@ export function Label(opts : LabelOptions) : ChildType {
 }
 
 export function AnimatedExperimentLabel(opts : AnimatedExperimentLabelOptions) : ChildType {
-    const { buttonAnimationComponent } = opts;
-    delete opts.buttonAnimationComponent;
+    const { buttonDesignComponent } = opts;
```
gregjopa, line 167: Could we remove the `delete` operation and replace using `...rest`?
```suggestion
    const { buttonDesignComponent, ...updatedOpts } = opts;
```

src/ui/buttons/buttonDesigns/control.jsx
```diff
@@ -0,0 +1,9 @@
+/* @flow */
+/** @jsx node */
+import { node, type ChildType } from 'jsx-pragmatic/src';
+
+export function ControlDesignComponent(experimentId : string) : ChildType {
+    return (
+        <div data-design-experiment={ experimentId } />
```
gregjopa, line 7: I like the data attribute approach 👍

src/ui/buttons/buttonDesigns/divideLogoAnimation.jsx
ssono: Eyeballed. Will replace with a check for existence and remove ternary

src/ui/buttons/buttonDesigns/divideLogoAnimation.jsx
```diff
@@ -0,0 +1,129 @@
...
+    `;
+
+    if (paypalLabelContainerElement) {
+        const style = document.createElement('style');
+        paypalLabelContainerElement.appendChild(style);
+        style.appendChild(document.createTextNode(designCss));
+    }
+}
+
+export function getDivideLogoAnimationScript() : string {
```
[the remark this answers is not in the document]
ssono: Not sure if this can/should be generalized. I'll think about it. The limiting factors would be the specifics of the css, which element the design is applied to, and the fact that it is being passed in as a string

src/ui/buttons/buttonDesigns/divideLogoAnimation.jsx
```diff
@@ -0,0 +1,129 @@
...
+    max:                            BUTTON_SIZE_STYLE.huge.maxWidth,
+    cssUtilClasses: {
+        DOM_READY:                  CLASS.DOM_READY,
+        PAYPAL_LOGO:                LOGO_CLASS.LOGO,
+        PAYPAL_BUTTON_LABEL:        CLASS.BUTTON_LABEL
+    }
+};
+
+// Returns props necessary to render the animation as long as they are valid
+const getValidDesignProps = function(document, configuration) : DivideLogoAnimationProps | null {
```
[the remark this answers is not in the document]
ssono: Maybe? Certain parts can be shared but not all of them. Since it is passed in as a string I'm not sure if it would be better or worse...

src/funding/paypal/template.jsx
ssono: Yes we can do this

src/ui/buttons/buttonDesigns/control.jsx
ssono: Yeah it should be better. Only worry I have with logging is how we ignore the resize cases for logs. Right now the designs will be removed whenever the button doesn't match the design requirements. We should send some signal in the logs, but not sure how to go about it

gregjopa (approved):

src/ui/buttons/buttonDesigns/divideLogoAnimation.jsx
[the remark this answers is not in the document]
erick2014: @jrod1032 it can be generalized, but I am afraid for some animations these properties are going to change, so creating a util for this is not very useful ( in my humble opinion )

src/ui/buttons/buttonDesigns/divideLogoAnimation.jsx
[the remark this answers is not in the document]
erick2014: I am afraid this script is different for every animation, so creating a util for this is not very useful ( in my humble opinion )

erick2014 (approved):

PR #1797, opened 2021-11-18
chore: upgrade to grumbler-scripts v5

westeezy:
### Description

Upgrade grumbler-scripts to v5 which includes an eslint upgrade. 

This tooling updates will not directly effect bundles that are generated as webpack and babel were unmodified.  **However the lint fixes may impact production behavior**. We **MUST** ensure these lint changes are accurate so please read carefully. No code was handwritten here it is the result of `eslint --fix` and the old lint rules being more correctly applied. If we want to change this behavior I can publish a v5.x.x of `grumbler-scripts` with the updated rules.

Good to stay up to date

`npm run test`

❤️  Thank you!


src/zoid/buttons/util.js
```diff
@@ -189,9 +189,9 @@ export function applePaySession() : ?ApplePaySessionConfigRequest {
                 listeners.paymentauthorized({ payment });
             };
 
-            session.oncancel = () => {
+            session.addEventListener('cancel', () => {
```
westeezy, line 192: should we ignore this to prevent the prod change?

src/ui/buttons/buttons.jsx
```diff
@@ -110,7 +110,7 @@ export function Buttons(props : ButtonsProps) : ElementNode {
     }
 
     if (fundingSources.indexOf(FUNDING.CARD) !== -1) {
-        fundingSources = fundingSources.filter(src => src !== FUNDING.CARD).concat([ FUNDING.CARD ]);
+        fundingSources = [ ...fundingSources.filter(src => src !== FUNDING.CARD),  FUNDING.CARD ];
```
westeezy, line 113: should we ignore this to prevent the prod change?

wsbrunson: what rule is this

src/zoid/buttons/util.js
wsbrunson: is this even right? I don't think that method exists on session: https://braintree.github.io/braintree-web/current/ApplePay.htm

wsbrunson (commented):
Does this repo have pre-commit hooks to run eslint --format? if not, I feel like now would be a good time to introduce it

westeezy:
> Does this repo have pre-commit hooks to run eslint --format? if not, I feel like now would be a good time to introduce it

it does not but we could explore using husky. Id like to explore prettier though as well so maybe we can table the pre-commit for another PR/discussion?

src/ui/buttons/buttons.jsx
westeezy: I'll look into it. I didn't change any rules in grumbler-scripts so I don't know off hand but should be easy to find. I guessing maybe the prefers-spread from unicorn is set to something. thats off the top of my head

src/zoid/buttons/util.js
westeezy: yeah i think the rule blindly applies to `oncancel` which kinda stinks... let's ignore this for now. I'll push an update

src/ui/buttons/buttons.jsx
westeezy: you know what i just want to ignore this too for now the more i think about it so gonna lint ignore here as well.

src/zoid/buttons/util.js
westeezy: `unicorn/prefer-add-event-listener`

wsbrunson (approved):

package.json
```diff
@@ -65,7 +65,7 @@
         "conventional-changelog-cli": "^2.0.34",
         "flow-bin": "0.155.0",
         "fs-extra": "^9.0.1",
-        "grumbler-scripts": "^3.0.70",
+        "grumbler-scripts": "5.0.0",
```
gregjopa, line 68: Should we keep the carrot here so we get patches and minor versions for free?
```suggestion
        "grumbler-scripts": "^5.0.0",
```

gregjopa (approved):
💯

package.json
westeezy: consider it done

PR #1802, opened 2021-12-01
Animation: Fadeout logo show label text

erick2014:
### Description

This is an animation where we fadeout the PayPal logo and show up a label text. The animation can be play infinite times or on hover:

Infinite Loop:

https://user-images.githubusercontent.com/7244087/145302798-cc06e958-b869-453c-8d42-baa542ca79f8.mov

On hover:

https://user-images.githubusercontent.com/7244087/145302769-081a2d5b-63cd-43aa-a979-6b3f910b9d82.mov

JIRA TICKET: https://engineering.paypalcorp.com/jira/browse/DTUPSTREAM-191

❤️  Thank you!

PR #1804, opened 2021-12-02
Animation: Fadeout logo show label text hover

erick2014:
### Description

This is an animation where we fadeout the PayPal logo and show up a label text on hover.

https://user-images.githubusercontent.com/7244087/144509010-856548b5-f304-4e0f-9d6a-4898d4ee2281.mov

JIRA TICKET: https://engineering.paypalcorp.com/jira/browse/DTUPSTREAM-110

Depends on pr: https://github.com/paypal/paypal-checkout-components/pull/1802

❤️  Thank you!


src/ui/buttons/buttonDesigns/index.js
```diff
@@ -31,6 +41,18 @@ const DESIGN_MAP : Object = {
...
+    'alternate-slide-logo-animation': {
+        designFn:              getAlternateSlideLogoAnimation,
+        getValidDesignProps:   getAlternateSlideLogoProps,
+        designConfig:          ALTERNATE_SLIDE_LOGO_CONFIG,
+        ButtonDesignComponent: AlternateSlideLogoComponent
+    },
+    'hover-slide-logo-animation': {
+        designFn:              getHoverSlideLogoAnimation,
+        getValidDesignProps:   getHoverSlideLogoProps,
+        designConfig:          ALTERNATE_SLIDE_LOGO_CONFIG,
```
ssono, line 54: Can you add a comment to explain why we are naming this ALTERNATE_SLIDE for the hover animation or name it something that fits both?

ssono (commented):
Mostly looks good. Left a few comments. Also wondering if it would be easy/worth it to control the hover vs infinite with a flag

src/ui/buttons/buttonDesigns/index.js
erick2014: yeah is because I am reusing the same config, but I am going to create a general config structure to re use it wherever we can.

erick2014:
> Mostly looks good. Left a few comments. Also wondering if it would be easy/worth it to control the hover vs infinite with a flag

I think I already tried, but was complicated so I ended up creating its own animation. Although I'm gonna spend some time to see if I can do something about it.

mnicpt (changes requested):
Please hold on merging

PR #1808, opened 2021-12-07
Personalization

mnicpt:
### Description
Update to use new Personalization architecture and repo.


erick2014:
@mnicpt it has linter errors

mnicpt:
It's in Draft and only have Daniel reviewing for now. I will let you know when I'm done.

PR #1809, opened 2021-12-08
Card holder name field

leogedler:
### Description

Defined a new Zoid component for the Card Holder's Name input for the card fields.

- https://engineering.paypalcorp.com/jira/browse/DTNOR-288

For the card fields we need an optional hosted field for the card holder's name.

The name is not PCI compliance but for the auto-fill browser autofill feature all the card fields related inputs need to be on an iframe for it to work as expected. 

It will be used as a child component of the CardField parent component.

image

This change will be used on SPB (paypal-smart-payment-buttons).
- https://github.com/paypal/paypal-smart-payment-buttons/pull/307

❤️  Thank you!

### DIFF
--- a/src/zoid/card-fields/component.jsx
+++ b/src/zoid/card-fields/component.jsx
@@ -20,7 +20,8 @@ const CARD_FIELD_TYPE = {
     SINGLE: 'single',
     NUMBER: 'number',
     CVV:    'cvv',
-    EXPIRY: 'expiry'
+    EXPIRY: 'expiry',
+    NAME:   'name'
 };
 
 type CardFieldsProps = {|
@@ -66,7 +67,8 @@ type CardFieldsExports = {|
 type CardFieldsChildren = {|
     NumberField : CardFieldComponent,
     CVVField : CardFieldComponent,
-    ExpiryField : CardFieldComponent
+    ExpiryField : CardFieldComponent,
+    NameField : CardFieldComponent
 |};
 
 const url = () => `${ getPayPalDomain() }${ __PAYPAL_CHECKOUT__.__URI__.__CARD_FIELD__ }`;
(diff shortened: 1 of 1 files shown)


bluepnume:
Interesting. Is this something that Braintree also supports right now?

leogedler:
@bluepnume yeah, actually was @crookedneighbor the one who suggested this approach because they had a similar implementation on Braintree.

In our case Rahul requested that we need to be able to set the card holder's name and billing address.

Here are the docs about the fields supported --> https://braintree.github.io/braintree-web/current/module-braintree-web_hosted-fields.html#~fieldOptions

crookedneighbor:
@bluepnume yes, it's supported both as a hosted field (to accommodate browser autofill) and as a raw text value that the merchant can collect on their own.

The billing address fields are all supported as raw text that the merchant collects and get attached to the resulting payment method token in the tokenization process.

Technically, in Braintree, postal code is supported as a hosted field, but there's no technical requirement for that other than it's slightly more convenient for the merchant if they have fraud rules set up to require it already.

bluepnume:
Good to know! In that case, if it helps with auto-fill, does it make sense to build all of the fields including billing/shipping fields?

crookedneighbor:
IMO, we shouldn't need to capture billing address within the hosted payment fields. Here's why.

If a merchant adds their own cardholder name input, it's impossible to link that up to the payment fields, so the browser autofill won't kick in until you've entered the hosted payment field. And when the autofill menu does appear within the hosted payment field, it's not able to add the cardholder name data to the raw input on the merchant page (we could technically make a pretty good guess about which input it is and fill it in automatically, but that would be making a lot of assumptions about how the merchant set up their form, and would be easy to get wrong).

For billing address inputs, those are treated as a whole unit separate from the card. So while they won't get automatically added via the browser credit card autofill, the customer will still be able to autofill them when they start to fill out the address portion of the form. By leaving the billing address inputs to the merchant, we let them make any complex data validation decisions, and make it easier for users with screen-readers to fill out the form. 

The obvious downside is that the customer has to go through two separate autofill events instead of just one, but a workaround we could do is emit an event when the browser autofill happens in the hosted payment fields and expose any billing address data that gets filled in so the merchant can fill in their own fields as needed.

Ultimately, if we want to expose a hosted payment field for each billing address piece, we can, but I think it's a better API to try and scope the hosted payment fields to things that require PCI compliance.

nbierdeman (approved):
Nice work, @leogedler! 🚀

crookedneighbor (approved):

PR #1814, opened 2021-12-13
fix: use unicode emoji for version message

crookedneighbor:
### Description

Very minor. Just uses the 🎉 unicode emoji for the npm version commit message so it appears correctly on all platforms. 

Just thought we should see the party popper instead of `:tada:` when looking at `git log` output.


elizabethmv (approved):

crookedneighbor:
Anything stopping this from getting merged? (I don't have the permission to do it myself)

PR #1826, opened 2022-01-05
Support payIn3 labels

elizabethmv:
### Description
Add new labels for Pay In 3

To support BNPL expansion to new countries 

This change is safe to release independently, but the GraphQL changes have also already been released. Currently payIn3 is defaulted false so these changes will not be visible in sandbox/prod yet.

### DIFF
--- a/src/ui/buttons/styles/responsive.js
+++ b/src/ui/buttons/styles/responsive.js
@@ -22,10 +22,13 @@ export function buttonResponsiveStyle({ height, fundingEligibility } : {| height
         const minDualWidth = Math.max(Math.round(buttonHeight * BUTTON_MIN_ASPECT_RATIO * (100 / WALLET_BUTTON_PERC)), MIN_SPLIT_BUTTON_WIDTH);
 
         const { paylater } = fundingEligibility;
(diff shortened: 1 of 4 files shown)


src/ui/buttons/styles/responsive.js
```diff
@@ -22,10 +22,13 @@ export function buttonResponsiveStyle({ height, fundingEligibility } : {| height
         const minDualWidth = Math.max(Math.round(buttonHeight * BUTTON_MIN_ASPECT_RATIO * (100 / WALLET_BUTTON_PERC)), MIN_SPLIT_BUTTON_WIDTH);
 
         const { paylater } = fundingEligibility;
-        const isInGerman = paylater?.products?.paylater?.variant === 'DE';
+        const resizeLabel =
```
gregjopa, line 25: Could we prefix this boolean variable name with `is`, `has`, or `should`?
```suggestion
        const shouldResizeLabel =
```

elizabethmv: Done!

gregjopa (approved):

PR #1829, opened 2022-01-07
Test belter version with CI

gregjopa:
The following test seems to be failing with `belter@1.0.179`:

https://github.com/paypal/paypal-checkout-components/runs/4742920827?check_suite_focus=true
```
 1) should render multiple buttons including venmo, click on the venmo button, and send the correct url params
     paypal multiple button component happy path on popup
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

This Draft PR pins to `belter@1.0.177` and the tests pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
         "@paypal/sdk-client": "^4.0.156",
         "@paypal/sdk-constants": "^1.0.107",
         "@paypal/sdk-logos": "^1.0.26",
-        "belter": "^1.0.2",
+        "belter": "1.0.177",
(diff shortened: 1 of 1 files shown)


gregjopa:
@TNuzzi @mnicpt, it looks like latest belter (`belter@1.0.179`) is breaking a test in checkout-components. Venmo is not rendering when it should. Can you take a look?

gregjopa:
I just published `belter@1.0.180`. Testing it with CI now.

PR #1838, opened 2022-01-11
Accept data-popups-disabled attribute as valid prop 

ravishekhar:
### Description
* Pass the data-popups-disabled script attribute to SPB, so that on button click an iframe is used to show checkout UI instead of popup
* Disable popup in case if a buyer clicks the button before second render and let iframe open up once SPB is loaded in second render. 

* This new "data-popups-disabled" data attribute will be used by merchants so they can use our components in non-standard browsers. For example, merchant can use SPB inside a C++ desktop application that embeds Chrome which does not support popups.

https://github.com/paypal/paypal-sdk-client/pull/99  **Needs to be merged first**
https://github.com/paypal/paypal-smart-payment-buttons/pull/318 

❤️  Thank you!


gregjopa (approved):
Great work @ravishekhar! 💯

src/zoid/buttons/prerender.jsx
```diff
@@ -43,7 +43,7 @@ export function PrerenderedButtons({ nonce, onRenderCheckout, props } : Prerende
 
         if (fundingSource === FUNDING.VENMO) {
             // wait for button to load
-        } else if (supportsPopups()) {
+        } else if (supportsPopups() && !props.merchantRequestedPopupsDisabled) {
```
gregjopa, line 46: Could you try adding a test for this option with prerender?

PR #1839, opened 2022-01-12
Menu Placement Style Prop for MenuButtons

westeezy:
Create a menuPlacement prop to be used to position the dropdown menu on
the paypal button. Needs continued effort for other buttons such as
card vaulting.

New Props:
- Buttons.style.menuPlacement = 'above' | 'below'
- Default to 'below' to ensure no changes to users until they opt in to
  new positions

Example Usage:
```
paypal.Buttons({
  style: {
    menuPlacement: 'above'
  },

  createOrder: function (data, actions) {
    return '1234';
  },

  onApprove: function (data, actions) {
    console.log(data, actions);
  }
```

Issues:
- Currently this works for the PayPal button only due the positioning
  logic being mixed between checkout-components and SPB. Functionality
  wise it always works but the menu looks disconnected on other button
  types.
- This PR sets the checkout-components wrapper css to ensure all drop downs appear at the
  top of the button container. Later we would want to have SPB report
  the coordinates to move `.smart-menu` wrapper to via either `top:
  <y>px;` or `bottom: <y>px;` then remove the `margin-top` from the
  `verticalOffset` prop in SPB


src/ui/buttons/props.js
```diff
@@ -391,7 +392,8 @@ export function normalizeButtonStyle(props : ?ButtonPropsInputs, style : ButtonS
         shape = fundingConfig.shapes[0],
         tagline = (layout === BUTTON_LAYOUT.HORIZONTAL && !fundingSource),
         height,
-        period
+        period,
+        menuDirection = MENU_PLACEMENT.BELOW
```
gregjopa, line 396: Should this be `menuPlacement`?
```suggestion
        menuPlacement = MENU_PLACEMENT.BELOW
```

westeezy: oh shoot thanks!

gregjopa (approved):
Looks good to me! I'm cool with this solution 👍

PR #1845, opened 2022-01-18
Add "publish alpha release to npm" workflow

nbierdeman:
### Description

This PR adds a GitHub Actions workflow for publishing an alpha release to npm. This will only work for base branches, as there are security risks with running workflows from forks.

We are making these changes in order to avoid having to grant publishing rights to every new contributor, and also to do as little locally as possible.

❤️  Thank you!

### DIFF
--- a/.github/workflows/alpha.yml
+++ b/.github/workflows/alpha.yml
@@ -1,0 +1,35 @@
+name: publish alpha release to npm
+# manually run this action using the GitHub UI
+# https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/
+on: workflow_dispatch
+jobs:
+  main:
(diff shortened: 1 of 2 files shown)


nbierdeman:
@gregjopa, I'm not entirely sure we want to do this as part of an alpha release: https://github.com/paypal/paypal-checkout-components/blob/master/scripts/postversion.sh#L6. Should it be `git push git@github.com:paypal/paypal-checkout-components.git HEAD;` instead?

nbierdeman:
@gregjopa, an alternative to `npm run release:alpha` would be:

- Update the version (manually do this in the `package.json` on one's branch)

Run the GitHub Actions alpha workflow, which entails:

- Create an npm build version: `npm run build`
- Publish the package: `npm publish --tag alpha`

PR #1836, opened 2022-01-10
Re-add button size queryParam

ssono:
### Description

- Add a query Param that calculates the buttons size to use for upstream presentment
- Dependent on https://github.com/krakenjs/zoid/pull/384
- https://engineering.paypalcorp.com/jira/browse/DTUPSTREAM-262
- Removal of the code the first time https://github.com/paypal/paypal-checkout-components/pull/1787

### Dependent Changes (if applicable)

❤️  Thank you!

PR #1846, opened 2022-01-19
Re-Add button size query param prop

gregjopa:
This reverts commit e0142429b997810fc614397d0237304506db59d2.

### Description

This PR adds the button width prop back to the component from #1836. We fixed the issue in zoid/belter so we should be safe to add this code back in 🤞 

cc @ssono @bluepnume


gregjopa:
This PR fixed the tests: https://github.com/krakenjs/belter/pull/80.

PR #1863, opened 2022-01-31
add .vscode config

andyfen:
currently when you open the project in vscode it will interpolate the typesystem as being typescript rather than flow by default (i've seen multiple people share their screens looking like this)

Screen Shot 2022-01-31 at 10 56 14 AM

We want to disable this for the workspace  and let the flow extension then process the types, rather than force the user to disable the typescript extension in their ide.

This is detailed here 
https://stackoverflow.com/questions/48859169/js-types-can-only-be-used-in-a-ts-file-visual-studio-code-using-ts-check

After (the project is now running flow)
Screen Shot 2022-01-31 at 10 56 37 AM

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "javascript.validate.enable": false
+}


gregjopa:
Thanks for trying to fix these vscode errors @andyfen. Have you tried out the Flow plugin for vscode? https://flow.org/en/docs/editors/vscode/. Wondering if we can try that instead.

andyfen:
> Thanks for trying to fix these vscode errors @andyfen. Have you tried out the Flow plugin for vscode? https://flow.org/en/docs/editors/vscode/. Wondering if we can try that instead.

yes i have that running, thats not the issue, the issue is that it will interpret these files as typescript, this resolves that and lets the flow plugin work.

andyfen:
it even tells you to do this in the plugin readme
https://marketplace.visualstudio.com/items?itemName=flowtype.flow-for-vscode
```
Set javascript.validate.enable option to false or completely disable the built-in TypeScript extension for your project (see gif below):
```

andyfen:
we want to avoid having to "completely disable the built-in TypeScript extension"

gregjopa:
> it even tells you to do this in the plugin readme https://marketplace.visualstudio.com/items?itemName=flowtype.flow-for-vscode
> 
> ```
> Set javascript.validate.enable option to false or completely disable the built-in TypeScript extension for your project (see gif below):
> ```

That makes sense. What do you think about adding a `.vscode/extensions.json` file to recommend the flow extension? It will recommend that devs install this specific flow plugin. https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

andyfen:
> > it even tells you to do this in the plugin readme https://marketplace.visualstudio.com/items?itemName=flowtype.flow-for-vscode
> > ```
> > Set javascript.validate.enable option to false or completely disable the built-in TypeScript extension for your project (see gif below):
> > ```
> 
> That makes sense. What do you think about adding a `.vscode/extensions.json` file to recommend the flow extension? It will recommend that devs install this specific flow plugin. https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

yes i would defiantly add a setting.json also and add the recommended plugins to that, the whole purpose of having a  .vscode config in the source code so the editor will work correctly.

elizabethmv:
Just a note but this can also be configured in your personal workspace vs this project with CMD + Shift + P and searching for Open Settings JSON

andyfen:
> Just a note but this can also be configured in your personal workspace vs this project with CMD + Shift + P and searching for Open Settings JSON

yeah but whenever anyone starts working on it they need to go through the process of configuring it all, seems a no brainer to me.

mnicpt (changes requested):
Don't add this.

mnicpt:
When we're working across repos, we will run into many configurations in which teams have built their repos.  VSCode gives us the ability to configure our own workspaces to accommodate our specific needs.  Putting `.vscode` into source control creates the potential for anyone to add their personal settings that are not necessarily conducive for the repo.  I don't think we should include this in our repos, and keep personal settings separate.  We should always aspire to place `nice` in other peoples spaces and not push individual needs.  To maintain your own private settings takes 2s to make.  It's part of respecting others domains to accommodate.  I'm open to hearing others thoughts as well, but I don't see this as a great need and has potential to get out of control once the door is open.  This is the first request we've ever received in 3 years since I've been here regarding a change like this.

andyfen:
Firstly, the repo simply doesn't work out the box because you are not doing what the readme tells you to do.

Secondly, you actually think the best approach is to leave it broken and leave people to figure out why it doesn't work and fix it for themselves just to get it to work yeah ? no.


>  Putting `.vscode` into source control creates the potential for anyone to add their personal settings that are not necessarily conducive for the repo. 

no one will update it, and if they do then you could just request they don't.

> This is the first request we've ever received in 3 years since I've been here regarding a change like this.

and ?

gregjopa:
Hey Andy, I do think these .vscode settings are great for a "Getting Started" guide for sdk development. But I don't think we are ready to commit them into the repos. Here's the reasoning for that decision:
1. Currently we ignore the `.vscode` folder in the `.gitigore` for the sdk repos. There may be some developers contributing who have settings saved in `.vscode` and I would worry that there would be conflicts. The .vscode folder can contain settings.json, launch.json, and extensions.json.
2. Checking in a standard `.vscode` folder would be difficult to maintain since we have 20+ sdk-related repos.

andyfen:
closing as this is a total joke

andyfen:
and to point out this isn't gitignored for this repo, so what you mention about conflicts is a hypothetical situation that has never occurred https://github.com/paypal/paypal-checkout-components/blob/main/.gitignore

PR #1867, opened 2022-02-07
Fix failing venmo test

gregjopa:
This PR resolves the following failing test: 

```
   paypal multiple button component happy path on popup
    ✗ should render multiple buttons including venmo, click on the venmo button, and send the correct url params
	Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

### DIFF
--- a/test/integration/tests/button/index.js
+++ b/test/integration/tests/button/index.js
@@ -1,22 +1,19 @@
 /* @flow */
-
+import './happy';
+import './props';
+import './validation';
+import './error';
+import './drivers';
+import './frame';
+import './size';
 import './multiple';
-
-// import './happy';
-// import './props';
-// import './validation';
-// import './error';
-// import './drivers';
-// import './frame';
-// import './size';
-
-// import './layout';
-// import './style';
-// import './instant';
-// import './popupBridge';
-// import './standalone';
-// import './clone';
-// import './renderOrder';
-// import './nonce';
-// import './eligibility';
-// import './animation';
+import './layout';
+import './style';
+import './instant';
+import './popupBridge';
+import './standalone';
+import './clone';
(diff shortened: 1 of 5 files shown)


gregjopa:
@elizabethmv @TNuzzi @mnicpt this venmo test is mysteriously failing. Could you take a look?

elizabethmv (commented):
It looks like SFVC is being detected for this user agent string https://github.com/paypal/paypal-checkout-components/blob/main/test/integration/tests/common.js#L45 

karma-logs

gregjopa:
> It looks like SFVC is being detected for this user agent string https://github.com/paypal/paypal-checkout-components/blob/main/test/integration/tests/common.js#L45
> 
> karma-logs

Good find @elizabethmv! 

How do think we should solve? Wondering if this is a bug with `isSFVC()` in belter or an issue with the karma test environment mocking. This user agent string use to pass.

elizabethmv:
We made a change in the latest belter to detect SFVC to prevent rendering the button if the device is unsupported:
https://github.com/krakenjs/belter/commit/7b1b78ae426bc78f8fff1690058d4d2aa11fcda6#diff-00b079f6a4e9ad71474c45460b40f61ad49e5df4b4fb493f3700253d2b22aa6aR101

It looks like we are getting `undefined` after checking the dimensions because the test is getting a screen height of 1440 and that does not match our matrix / isn't really a valid iPhone height: https://github.com/krakenjs/belter/blob/main/src/screenHeights.js#L4.

I think its safe to update the test, I'll push a change up @gregjopa

gregjopa (commented):
Awesome job fixing the failing test @elizabethmv! 💯

PR #1880, opened 2022-02-22
feat(incognito): Log FTPI event if browser can't access localstorage

TNuzzi:
### Description

Added event logging for VMORECOMBO-440 to possibility detect if browser is in private mode (Incognito).  This is primarily for Android but good to log all browsers.

Screen Shot 2022-02-22 at 12 51 51 PM

NOTE: By default Android Chrome has `Block third-party cookies in Incognito` enabled by default.  When accessing cookies from a third party site an exception is thrown because site-data (localStorage) is also disabled.  We use this to detect the possibility of incognito mode.  

To reproduce on the desktop:
1) In Google Chrome go to settings/cookies: chrome://settings/cookies
2) Enable: Block all cookies (not recommended)
3) Hit the test page.  You should see the following log in the console: `localstroage_inaccessible_possible_private_browsing `
To test:
1) 

❤️  Thank you!

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -50,7 +50,7 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                     [ FPTI_KEY.CONTEXT_TYPE ]:       'button_session_id',
                     [ FPTI_KEY.CONTEXT_ID ]:         buttonSessionID,
                     [ FPTI_KEY.TRANSITION ]:         'localstorage_inaccessible_possible_private_browsing'
-                }).flush();
+                });
             }
 
             return (
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -49,7 +49,7 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                     [ FPTI_KEY.BUTTON_SESSION_UID ]: buttonSessionID,
                     [ FPTI_KEY.CONTEXT_TYPE ]:       'button_session_id',
                     [ FPTI_KEY.CONTEXT_ID ]:         buttonSessionID,
-                    [ FPTI_KEY.TRANSITION ]:         'localstoage_inaccessible_possible_private_browsing'
+                    [ FPTI_KEY.TRANSITION ]:         'localstorage_inaccessible_possible_private_browsing'
                 }).flush();
             }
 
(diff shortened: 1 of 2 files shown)


src/zoid/buttons/util.js
```diff
@@ -138,7 +138,14 @@ export function getNoPaylaterExperiment(fundingSource : ?$Values<typeof FUNDING>
 
 export function getVenmoAppLabelExperiment() : EligibilityExperiment  {
     const isEnvForTest = getEnv() === ENV.LOCAL || getEnv() === ENV.TEST || getEnv() === ENV.STAGE;
-    const isEnabledForTest = isEnvForTest ? window.localStorage.getItem('enable_venmo_app_label') : false;
+
+    let isEnabledForTest = false;
+
+    isEnabledForTest = window.localStorage.getItem('enable_venmo_app_label');
```
gregjopa, line 144: Is this line a duplicate?
```suggestion
```

TNuzzi:
@gregjopa good catch, missed that in my second commit.

src/zoid/buttons/component.jsx
```diff
@@ -42,6 +42,17 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
         logger: getLogger(),
 
         prerenderTemplate: ({ state, props, doc }) => {
+            const { buttonSessionID } = props;
```
gregjopa, line 45: Looks good. I think you could also log this logic in the onInit() decorator on line 260.

TNuzzi: @gregjopa I had originally it there but wanted it as early as possible.  The reason is my next update will suppress the Venmo button (on prerender) if `localStorage` is not accessible.

src/zoid/buttons/component.jsx
```diff
@@ -42,6 +42,17 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
...
+            
+            if (!isLocalStorageEnabled()) {
+                getLogger().info('localstoage_inaccessible_possible_private_browsing').track({
+                    [ FPTI_KEY.BUTTON_SESSION_UID ]: buttonSessionID,
+                    [ FPTI_KEY.CONTEXT_TYPE ]:       'button_session_id',
+                    [ FPTI_KEY.CONTEXT_ID ]:         buttonSessionID,
+                    [ FPTI_KEY.TRANSITION ]:         'localstoage_inaccessible_possible_private_browsing'
```
gregjopa, line 52:
```suggestion
                    [ FPTI_KEY.TRANSITION ]:         'localstorage_inaccessible_possible_private_browsing'
```

gregjopa (approved):
Great work Tony! 💯 

This belter local storage check is great with using a try/catch: https://github.com/krakenjs/belter/blob/main/src/dom.js#L306-L327

gregjopa (approved):

PR #1894, opened 2022-03-18
Change inlinexo to inline

mnicpt:
### Description
Making the external facing prop customer focussed and not internal language.


westeezy (approved):
lgtm, but I would love to start documenting what some of these props do in the codebase. Would you be open to adding an inline comment on `buttons/props.js` above inline to give a quick ELI5 of what this prop would do? `inline` as a word can be interpreted in many ways especially for newcomers. I also recognize this isn't something we did in the past so this is a proposal that is net new from me. Feel free to merge without it but if you like this idea I think it would be helpful for others.

PR #1895, opened 2022-03-25
Keep inlinexo internally

mnicpt:
### Description
We allow the merchant to set the `inline` prop to use Inline XO.  Because `xo` is an internal naming convention and `inline` is already used in Hermes, we will use `inlinexo` internally back to Hermes.


remotevision (approved):
👍

PR #1904, opened 2022-03-30
Update Styling for Inline XO

mnicpt:
### Description
- Allow merchant to just supply custom styling and get custom button on top.
- Changes `inline` prop to `experience`.

image

### DIFF
--- a/src/constants/button.js
+++ b/src/constants/button.js
@@ -61,3 +61,7 @@ export const MENU_PLACEMENT = {
     ABOVE:   ('above' : 'above'),
     BELOW: ('below' : 'below')
 };
+
+export const EXPERIENCE = {
+    INLINE: ('inline' : 'inline')
+};
--- a/src/ui/buttons/button.jsx
+++ b/src/ui/buttons/button.jsx
@@ -45,7 +45,7 @@ type IndividualButtonProps = {|
 export function Button({ fundingSource, style, multiple, locale, env, fundingEligibility, i, nonce, flow, vault,
     userIDToken, personalization, onClick = noop, content, tagline, commit, experiment, instrument, experience } : IndividualButtonProps) : ElementNode {
 
-    const inline = experience === EXPERIENCE.INLINE;
+    const inlineExperience = experience === EXPERIENCE.INLINE;
     const fundingConfig = getFundingConfig()[fundingSource];
 
     if (!fundingConfig) {
@@ -131,7 +131,7 @@ export function Button({ fundingSource, style, multiple, locale, env, fundingEli
             personalization={ personalization }
             tagline={ tagline }
             content={ content }
-            custom={ inline ? custom : null }
+            custom={ inlineExperience ? custom : null }
             experiment={ experiment }
         />
     );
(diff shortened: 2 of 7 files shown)


src/constants/button.js
```diff
@@ -61,3 +61,7 @@ export const MENU_PLACEMENT = {
     ABOVE:   ('above' : 'above'),
     BELOW: ('below' : 'below')
 };
+
+export const EXPERIENCE = {
```
gregjopa, line 65: I like the new `EXPERIENCE` addition. It will scale with supporting more possible values in the future 💯

src/ui/buttons/button.jsx
```diff
@@ -39,12 +39,13 @@ type IndividualButtonProps = {|
...
     instrument : ?WalletInstrument,
-    inline? : boolean
+    experience? : string
 |};
 
 export function Button({ fundingSource, style, multiple, locale, env, fundingEligibility, i, nonce, flow, vault,
-    userIDToken, personalization, onClick = noop, content, tagline, commit, experiment, instrument, inline } : IndividualButtonProps) : ElementNode {
+    userIDToken, personalization, onClick = noop, content, tagline, commit, experiment, instrument, experience } : IndividualButtonProps) : ElementNode {
 
+    const inline = experience === EXPERIENCE.INLINE;
```
gregjopa, line 48: It would be nice to make this variable more descriptive.
```suggestion
    const isInlineExperience = experience === EXPERIENCE.INLINE;
```

src/ui/buttons/button.jsx
```diff
@@ -199,6 +200,7 @@ export function Button({ fundingSource, style, multiple, locale, env, fundingEli
         <div
             class={ [
                 CLASS.BUTTON_ROW,
+                inline && fundingSource === FUNDING.CARD ? CLASS.CUSTOM : '',
```
gregjopa, line 203:
```suggestion
                isInlineExperience && fundingSource === FUNDING.CARD ? CLASS.CUSTOM : '',
```

src/ui/buttons/buttons.jsx
```diff
@@ -121,6 +127,18 @@ export function Buttons(props : ButtonsProps) : ElementNode {
     );
 
     const { buttonDesignScript = '' } = getButtonDesign(personalization);
+    const index = (i) => {
+        if (!inline) {
```
gregjopa, line 131:
```suggestion
        if (!isInlineExperience) {
```

src/ui/buttons/styles/custom.js
```diff
@@ -10,11 +10,19 @@ export const customStyle = ({ custom } : {| custom? : CustomStyle |}) : string =
 
     const { css } = custom || {};
 
+    let heightStyle = '';
     let style = Object.keys(css).reduce((acc, key) => {
+        if (key === 'height') {
```
gregjopa, line 15: I'm not sure the exact custom styling requirements, but there may be some overlap with what's in SPB for the new card fields here: https://github.com/paypal/paypal-smart-payment-buttons/blob/ee31f1d4ee9bea646dcb21b5b7cf88fe4f8d974c/src/card/lib/card-utils.test.js#L372. Just wanted to mention it incase there's some logic we can copy/paste.

gregjopa (approved):
Looks good! I left a some non-blocking comments.

src/ui/buttons/button.jsx
mnicpt: Updating all suggestions with `inlineExperience`. Just think it reads better in an if-condition.

src/ui/buttons/styles/custom.js
mnicpt: Actually will be having the merchant enter in custom styling as they would in CSS with no conversion.

PR #1906, opened 2022-04-05
onComplete Callback

mnicpt:
### Description
Adds an `onComplete` callback for all supported components in cases where PayPal captures order and `onApprove` is not required.  This allows a combination of APM and branded FI's to have the appropriate callback to take next steps in the checkout flow.


remotevision (approved):
👍

PR #1918, opened 2022-04-12
Inline XO Global Usage

mnicpt:
### Description
Uses inline xo eligibility global variable for setting experience === 'inline'.


src/zoid/buttons/component.jsx
```diff
@@ -608,10 +608,28 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
...
-                    return isInlineXO ? EXPERIENCE.INLINE : '';
+                    const { commit, createBillingAgreement, currency, disableFunding, fundingEligibility, locale, merchantID, style: { layout }, vault } = props || {};
+                    
+                    let pageType;
+                    try {
+                        pageType = getPageType() || '';
+                    } catch (e) {
+                        pageType = '';
+                    }
+                    const inlineCheckoutEligibility = __INLINE_CHECKOUT_ELIGIBILITY__ || {
```
wsbrunson, line 619: should we be getting this value from `sdk-client` like 
```
function getInlineCheckoutEligility () {
  return __INLINE_CHECKOUT_ELIGIBILITY__
}
```

mnicpt: We don't have to and thought the same. It's not related to the script and was why I didn't put a helper method there.

mnicpt: I'm open to it though if we think it is best there.  The client acts as a way to get data across multiple components but we don't need to do that here since only checkout-components needs to know about it.

wsbrunson: one benefit of `sdk-client` for sure is making these build variables available to multiple components. One of the big benefits though is abstracting away how these variables become available to the components. These top-level build variables are hard to test properly since they techincally don't change after a script is compiled. We are also discussing different ways to pass down `__FUNDING_ELIGIBILITY__` and pull it out of the main cache key because we are getting to a point of too many unique values

PR #1922, opened 2022-04-15
remove vuln

wsbrunson:


src/setup.js
```diff
@@ -136,11 +136,6 @@ function configure({ env, stage, stageUrl, apiStage, localhostUrl, checkoutUri,
         config.apiStage = apiStage;
     }
 
-    localhostUrl = localhostUrl || (Button.xprops && Button.xprops.localhostUrl) || (Checkout.xprops && Checkout.xprops.localhostUrl);
```
gregjopa, line 139: Looks good to me! Let's also delete this `localhostUrl` from the props too. We shouldn't need it at all.

```
src/button/component.jsx
```

src/setup.js
```diff
@@ -136,11 +136,6 @@ function configure({ env, stage, stageUrl, apiStage, localhostUrl, checkoutUri,
...
-    if (localhostUrl) {
-        config.localhostUrl = localhostUrl;
-    }
-
     checkoutUri = checkoutUri || (Button.xprops && Button.xprops.checkoutUri) || (Checkout.xprops && Checkout.xprops.checkoutUri);
```
gregjopa, line 139: Let's remove `checkoutUri` as well.

src/setup.js
```diff
@@ -196,8 +191,6 @@ if (currentScript) {
         stage:              currentScript.getAttribute('data-stage'),
         apiStage:           currentScript.getAttribute('data-api-stage'),
         stageUrl:           currentScript.getAttribute('data-stage-url'),
-        localhostUrl:       isPayPalDomain() ? currentScript.getAttribute('data-localhost-url') : undefined,
```
gregjopa, line 204: I think we need to keep support for these `data-localhost-url` and `data-checkout-uri` attributes. They are safe to keep because they have the `isPayPalDomain()` check in there.

gregjopa (approved):
💯

src/button/component.jsx
```diff
@@ -355,30 +355,6 @@ export const Button : Component<ButtonOptions> = create({
             }
         },
 
-        localhostUrl: {
```
gregjopa, line 358: I think you can delete these props from the Types as well: https://github.com/paypal/paypal-checkout-components/blob/v4/src/button/component.jsx#L108-L109

westeezy:
There appears to be a lot of flake on GH Actions for v4 branch and this PR. Can confirm that this all passes locally. We should investigate the flake as time permits. The bulk of flake is coming from screenshots timing out.

westeezy (approved):

PR #1928, opened 2022-04-21
fix: remove configuration of localhostUrl

westeezy:
This is a draft and not ready to merge.

### DIFF
--- a/src/setup.js
+++ b/src/setup.js
@@ -137,7 +137,7 @@ function configure({ env, stage, stageUrl, apiStage, localhostUrl, checkoutUri, 
     }
 
     if (localhostUrl) {
-        config.localhostUrl = localhostUrl;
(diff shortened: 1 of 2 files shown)


src/setup.js
```diff
@@ -137,7 +137,7 @@ function configure({ env, stage, stageUrl, apiStage, localhostUrl, checkoutUri,
     }
 
     if (localhostUrl) {
-        config.localhostUrl = localhostUrl;
+        // localhostUrl is not configurable and hardcoded to localhost.paypal.com
```
westeezy, line 140: leaving config signature intact and passing on the setter here. i could be convinced to fully remove it as well, but I was hoping this would leave more historical information for the next person.

westeezy:
We discovered what we believe to be the root cause and will open a new PR.

PR #1929, opened 2022-04-21
fix: remove configuration of stageUrl via props

westeezy:
This is a continued effort from PR#1922.

We had assumed we caught all of the xss vector around the `localhostUrl`
being used in the window.name of the iframe. It appears there still
existed an attack surface area around `stageUrl` that was reported to us.

PR 1922: https://github.com/paypal/paypal-checkout-components/pull/1922/files
sha: https://github.com/paypal/paypal-checkout-components/commit/e5aa4090a8257a04416f3ce28669c452a6ab4585

### DIFF
--- a/src/setup.js
+++ b/src/setup.js
@@ -114,14 +114,6 @@ function configure({ env, stage, stageUrl, apiStage, localhostUrl, checkoutUri, 
         delete config.stageUrl;
         // $FlowFixMe
         config.stageUrl = stageUrl;
-    } else if (Button.xprops && Button.xprops.stageUrl) {
-        delete config.stageUrl;
-        // $FlowFixMe
-        config.stageUrl = Button.xprops.stageUrl;
-    } else if (Checkout.xprops && Checkout.xprops.stageUrl) {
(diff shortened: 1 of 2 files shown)


src/setup.js
```diff
@@ -198,7 +190,7 @@ if (currentScript) {
         env:                currentScript.getAttribute('data-env'),
         stage:              currentScript.getAttribute('data-stage'),
         apiStage:           currentScript.getAttribute('data-api-stage'),
-        stageUrl:           currentScript.getAttribute('data-stage-url'),
+        stageUrl:           isPayPalDomain() ? currentScript.getAttribute('data-stage-url') : undefined,
```
gregjopa, line 193: Good to see a single place for setting `stageUrl` with `data-stage-url` and enforcing the regex domain check.

gregjopa (approved):

TNuzzi (approved):

westeezy:
CI on v4 has been failing but this is confirmed passing tests locally and smoke tested locally as well. We will proceed with the merge.

PR #1930, opened 2022-04-27
Update styling for InlineXO

mnicpt:
### Description
Fixes a `css` check on the `custom` styling attribute and prevents free text entry for label.  Will provide more option in the future that we will localize like we currently do with PayPal FI.

### DIFF
--- a/src/funding/card/config.jsx
+++ b/src/funding/card/config.jsx
@@ -122,10 +122,20 @@ export function getCardConfig() : FundingSourceConfig {
         },
 
         Label: ({ logo, locale, content, custom }) => {
-            if (custom && custom.label) {
+            if (custom) {
+                const validLabels = {
+                    checkout: 'Checkout'
+                };
+
+                let label = 'Checkout';
+
+                if (custom.label && validLabels[custom.label]) {
+                    label = validLabels[custom.label];
+                }
+
                 return (
                     <Fragment>
-                        <Text>{ custom.label }</Text>
+                        <Text>{ label }</Text>
                         <Space />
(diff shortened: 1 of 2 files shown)


src/funding/card/config.jsx
```diff
@@ -122,10 +122,20 @@ export function getCardConfig() : FundingSourceConfig {
...
         Label: ({ logo, locale, content, custom }) => {
-            if (custom && custom.label) {
+            if (custom) {
+                const validLabels = {
+                    checkout: 'Checkout'
+                };
+
+                let label = 'Checkout';
+
+                if (custom.label && typeof custom.label === 'string' && validLabels[custom.label.toLowerCase()]) {
```
westeezy, line 132: can flow catch type of label or does this come from external source?

westeezy (approved):

src/funding/card/config.jsx
```diff
@@ -122,10 +122,20 @@ export function getCardConfig() : FundingSourceConfig {
         },
 
         Label: ({ logo, locale, content, custom }) => {
-            if (custom && custom.label) {
+            if (custom) {
+                const validLabels = {
+                    checkout: 'Checkout'
+                };
+
+                let label = 'Checkout';
```
gregjopa, line 130: Let's reuse validLabels for the default value:
```suggestion
                let label = validLabels.checkout;
```

gregjopa (approved):

src/funding/card/config.jsx
mnicpt: external source

PR #1944, opened 2022-05-16
Inline payment-fields

AnshuChaudhari:
### Description
Branded to Unbranded migration via SPB vertical stack.
- Exported `PaymentFields` zoid component to get included in `paypal` global object.
- Added few new props to `payment-fields` zoid component required for this feature.

APMs have branded (PayPal logos / styles / guest user creation) flow via smart payment buttons.
We also built [unbranded](https://developer.paypal.com/docs/checkout/apm/) (no PayPal logos / configurable style / inline payment fields)  flow through [standalone buttons](https://developer.paypal.com/docs/multiparty/checkout/standard/customize/standalone-buttons/).

Our product requirement is to move away from branded flows and bring in an inline checkout experience for APMs that we built for unbranded(existing) to vertical stack as well.

1. Go to go/smartbuttons, add the custom stage as `te-apm-unbranded-vertical`
2. Select country as Austria and integration tye as OrdervsV2
3. Hit Refresh Buttons
4. Click on eps funding source once vertical stack buttons are rendered
5. Inline-payment-fields should render

**Before:**

before-eps

**After:**

After-eps-local

Not depending but corresponding changes in SPB for this feature - https://github.com/paypal/paypal-smart-payment-buttons/pull/405/files

❤️  Thank you!


src/zoid/payment-fields/component.jsx
```diff
@@ -122,6 +122,16 @@ export function getPaymentFieldsComponent() : PaymentFieldsComponent {
                     required: false
                 },
 
+                onContinue: {
+                    type:     'function',
+                    required: false
+                },
+
+                createOrder: {
```
AnshuChaudhari, line 130: Do we need this `createOrder` prop here?

src/ui/buttons/buttons.jsx
```diff
@@ -203,6 +203,12 @@ export function Buttons(props : ButtonsProps) : ElementNode {
                     : null
             }
 
+            {
+                (fundingSources.indexOf(FUNDING.EPS) !== -1)
```
AnshuChaudhari, line 207: Should we add a condition that will incorporate any future funding sources? Currently, this is EPS specific

src/zoid/payment-fields/component.jsx
faiz121: No Anshu, we don't need that anymore

src/interface/button.js
```diff
@@ -32,6 +33,10 @@ export const CardForm : LazyProtectedExport<CardFormComponent> = {
     __get__: () => protectedExport(getCardFormComponent())
 };
 
+export const PaymentFields : LazyProtectedExport<PaymentFieldsComponent> = {
+    __get__: () => getPaymentFieldsComponent()
```
AnshuChaudhari, line 37: Should we add the domain check `protectedExport` wrapper around `getPaymentFieldsComponent`? All other components have it.
By adding protectedExport we were getting this error sometimes
image

src/ui/buttons/buttons.jsx
```diff
@@ -106,9 +106,11 @@ export function Buttons(props : ButtonsProps) : ElementNode {
         nonce, components, onShippingChange, onShippingAddressChange, onShippingOptionsChange, personalization, userIDToken, content, flow, experiment, applePaySupport,
         supportsPopups, supportedNativeBrowser, experience } = normalizeButtonProps(props);
     const { custom, layout, shape, tagline } = style;
- 
+
     const inlineExperience = experience === EXPERIENCE.INLINE && custom && custom.label && custom.label.length !== 0;
 
+    const fundingSourcesWithInlinePaymentFields = [ FUNDING.EPS ];
```
weihou, line 112: can we use the `inlineExperience` above or some similar function so that it is easy to maintain and ramp up/down.

src/interface/button.js
mnicpt: The `protectedExport` means it is not added to the `window.paypal.Buttons` global object.

src/zoid/payment-fields/component.jsx
```diff
@@ -122,6 +122,11 @@ export function getPaymentFieldsComponent() : PaymentFieldsComponent {
                     required: false
                 },
 
+                onContinue: {
```
mnicpt, line 125: What is the purpose of this callback? Wondering if it is the same as `onComplete` that was created for flows where we capture or authorize on our servers.  `onComplete` would be great for Apple Pay, for example.

jeffreysmooths: ..yh..

src/interface/button.js
AnshuChaudhari: @mnicpt Thanks for the clarification! 
So in that case how are we able to access all these components like `CardForm`, `Checkout`, `Menu` etc [here in spb](https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/button/props.js#L123) from `window.paypal` global object?
We are trying to access `PaymentFields` similar fashion [here](https://github.com/paypal/paypal-smart-payment-buttons/pull/405/files#diff-e3a4efeee1834a0c5a8750c378aef4deb5d1a4d1638b72565c8b91759ab23211R124)

src/zoid/payment-fields/component.jsx
AnshuChaudhari: @mnicpt We are trying to render the payment-fields inline for APMs, for now lets say `eps`. If you see in the screenshot after onclick of eps from vertical stack `PaymentFields` iframe will be rendered inline and it has a Continue button. This `onContinue` callback is been used for this button.
image

AnshuChaudhari:
retest this please

src/ui/buttons/buttons.jsx
```diff
@@ -210,6 +210,8 @@ export function Buttons(props : ButtonsProps) : ElementNode {
                     : null
             }
 
+            <div id="payment-fields-container" className="payment-fields-container"/>
```
mnicpt, line 213:
```
(apmSources.indexOf(fundingSource) !== -1
```

mnicpt: Move APM constant out of SPB and into sdk-constants repo.

wsbrunson (approved):

wsbrunson (approved):

PR #1955, opened 2022-06-21
Inline XO: Experiment

mnicpt:
### Description
Inline XO experiment.

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -5,7 +5,8 @@
 import { getLogger, getLocale, getClientID, getEnv, getIntent, getCommit, getVault, getDisableFunding, getDisableCard,
     getMerchantID, getPayPalDomainRegex, getCurrency, getSDKMeta, getCSPNonce, getBuyerCountry, getClientAccessToken, getPlatform,
     getPartnerAttributionID, getCorrelationID, getEnableThreeDomainSecure, getDebug, getComponents, getStageHost, getAPIStageHost, getPayPalDomain,
-    getUserIDToken, getClientMetadataID, getAmount, getEnableFunding, getStorageID, getUserExperienceFlow, getMerchantRequestedPopupsDisabled, getVersion } from '@paypal/sdk-client/src';
+    getUserIDToken, getClientMetadataID, getAmount, getEnableFunding, getStorageID, getUserExperienceFlow, getMerchantRequestedPopupsDisabled,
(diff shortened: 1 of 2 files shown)


src/zoid/buttons/component.jsx
```diff
@@ -711,7 +712,7 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                         }
                     }
 
-                    return inlineCheckoutEligibility &&  inlineCheckoutEligibility.eligible && alphaEligible && isInlineXOEligible({ props: {
+                    const eligible = inlineCheckoutEligibility &&  inlineCheckoutEligibility.eligible && alphaEligible && isInlineXOEligible({ props: {
```
gregjopa, line 715: nit: extra space
```suggestion
                    const eligible = inlineCheckoutEligibility && inlineCheckoutEligibility.eligible && alphaEligible && isInlineXOEligible({ props: {
```

gregjopa (approved):

src/zoid/buttons/component.jsx
```diff
@@ -720,7 +721,32 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
...
+                            .info(treatment)
+                            .track({
+                                [FPTI_KEY.EXPERIMENT_NAME]: 'inlinexo',
+                                [FPTI_KEY.TREATMENT_NAME]:  treatment
+                            }).flush();
+
+                        return inlinexoExperiment.isEnabled() ? EXPERIENCE.INLINE : '';
+                    }
+
+                    return '';
```
gregjopa, line 749: Do we need to explicitly return an empty string? Or will `undefined` be okay?

src/zoid/buttons/component.jsx
```diff
@@ -720,7 +721,32 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
...
+                    if (eligible) {
+                        const inlinexoExperiment = createExperiment('inlinexo', 50, logger);
+                        const treatment = inlinexoExperiment.getTreatment();
+
+                        logger
+                            .info(treatment)
+                            .track({
+                                [FPTI_KEY.EXPERIMENT_NAME]: 'inlinexo',
+                                [FPTI_KEY.TREATMENT_NAME]:  treatment
+                            }).flush();
```
wsbrunson, line 744: I'm not sure that we'd want to potentially flush twice. we might want to remove both flushes above and only call it right before the return statement

PR #1963, opened 2022-07-07
add payment request prop

andyfen:
ref https://github.com/paypal/paypal-smart-payment-buttons/pull/437

adds 

- paymentRequest config prop 

- applepaySession abort and cancel

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -461,6 +461,12 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 }
             },
 
+            paymentRequest: {
+                type:       'object',
+                queryParam: false,
+                required: false
+            },
+
             flow: {
                 type:       'string',
                 queryParam: true,
--- a/src/ui/buttons/props.js
+++ b/src/ui/buttons/props.js
@@ -350,6 +350,8 @@ export type ApplePayPaymentAuthorizationResult = {|
 
 export type ApplePaySessionConfig = {|
     begin : () => void,
+    abort : () => void,
+    oncancel : () => void,
     addEventListener : (string, Function) => void,
     // eslint-disable-next-line flowtype/no-weak-types
     completeMerchantValidation : (validatedSession : any) => void,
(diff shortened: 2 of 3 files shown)


mnicpt (changes requested):
Why is abort and new button prop added? Are we updating ApplePay version we are supporting? Please give more background to the "why" we are doing this change.

src/zoid/buttons/component.jsx
```diff
@@ -461,6 +461,12 @@ export const getButtonsComponent : () => ButtonsComponent = memoize(() => {
                 }
             },
 
+            paymentRequest: {
+                type:       'object',
+                queryParam: true,
```
mnicpt, line 466: @andyfen If this is approved, I'm not so sure we should have it passed as a queryParam.  Are we getting this from the query param anywhere?  We're trying to bring the query param down so it doesn't get truncated. We saw this occurring on native app switching.

mnicpt:
@andyfen Why was a screenshot updated?

src/zoid/buttons/component.jsx
andyfen: yup it is only accessed via xprops, will remove, great point.

andyfen:
> @andyfen Why was a screenshot updated?

i have not idea, i thought i was part of a githook or something, will remove, i guess its not gitignored

src/zoid/buttons/component.jsx
andyfen: just to confirm `queryParam: false` we still have it available as a xprop it's just not in the queryparam right ?

mnicpt: Correct

gregjopa:
Hey @andyfen this PR is looking good. Seems like there are two changes needed and then this will be good to merge:
1. remove `queryParams: true` since this is not needed server-side, only client-side inside the iframe via window.xprops.
2. remove the new image from the screenshot test. This should not be needed.

gregjopa (approved):

wsbrunson (approved):

PR #1967, opened 2022-07-13
PostalCode Name Field

jplukarski:
### Description

A zoid component for a Postal Code input for unbranded multi-card fields

We want to ensure we have parity with the current PayPal Hosted Fields offerings. Right now, Hosted Fields has an optional Postal Code field for merchants that need to collect this information for Fraud Checks

[Jira - DTNOR-528](https://engineering.paypalcorp.com/jira/browse/DTNOR-528)

Screen Shot 2022-07-13 at 4 48 47 PM

- https://github.com/paypal/paypal-smart-payment-buttons/pull/442

### Co-Authored By:
-@bywood
-@jplukarski 

❤️  Thank you!


src/zoid/card-fields/component.jsx
```diff
@@ -197,6 +201,18 @@ export const getCardFieldsComponent : () => CardFieldsComponent = memoize(() : C
                     }
                 },
 
+                minLength: {
```
gregjopa, line 204: I worry that `minLength` and `maxLength` are too generic. Could we rename them so its clear that they are used for the postalCode?
```suggestion
                postalCodeMinLength: {
```

src/zoid/card-fields/prerender.jsx
```diff
@@ -8,7 +8,7 @@ type PrerenderedCardProps = {|
     height : ?number
 |};
 
-const DEFAULT_HEIGHT = 30;
+const DEFAULT_HEIGHT = 78;
```
gregjopa, line 11: Does this impact the single card field height? Wondering if we need to give that one a different prerender height since it's only one text box.

bywood: Yes, this is used by the single card field as well. From the [Figma designs](https://www.figma.com/file/78svOAGmEt7fk4OIU6Qde1/UP_-UDCC-Drop-in?node-id=472%3A3161), the single card field should have the same height as each multi card field. This is the height for each individual card field, with some extra to allow the focus drop shadow to display.

src/zoid/card-fields/component.jsx
bywood: These map directly to the `<input>` element's `minlength` and `maxlength` attributes. So I'd prefer to keep them generic. This is an example of how they are set:
```js
cardFields.PostalCodeField({
    minLength: 5,
    maxLength: 8
});
``` 
So it should be pretty clear that these are for the `PostalCodeField`. Also, in the future we may want to allow merchant's to set them on other card fields. For example:
```js
cardFields.NameField({
    maxLength: 500
});
```
So it would be nice to keep them consistent.

gregjopa: ahh I think I understand now. The card field props are generic in checkout-components and the implementation for each child component is in SPB.

gregjopa (approved):

PR #1978, opened 2022-07-27
Adding eligibility check to CardFields

bywood:
### Description

This adds an eligibility check to the `CardFields`, which allows the merchant to call `isEligible` before rendering any of the card fields.

[DTNOR-497 - [JS] Supported HCFs Methods: isEligible](https://engineering.paypalcorp.com/jira/browse/DTNOR-497)

Checkout SDK

### Co-Authored By:
-@bywood
-@jplukarski

❤️  Thank you!

### DIFF
--- a/src/zoid/card-fields/component.jsx
+++ b/src/zoid/card-fields/component.jsx
@@ -306,6 +306,19 @@ export const getCardFieldsComponent : () => CardFieldsComponent = memoize(() : C
             }
         },
 
+        eligible: () => {
+            const fundingEligibility = getRefinedFundingEligibility();
+            if (fundingEligibility?.card?.eligible) {
+                return {
+                    eligible: true
+                };
+            }
+            return {
+                eligible: false,
+                reason: 'card payments are not eligible'
+            };
+        },
+
         props: {
             type: {
                 type:       'string',


src/zoid/card-fields/component.jsx
```diff
@@ -306,6 +306,19 @@ export const getCardFieldsComponent : () => CardFieldsComponent = memoize(() : C
             }
         },
 
+        eligible: () => {
```
gregjopa, line 309: Can we have this return a boolean value? That's the standard right now and how it's defined in the spec: 
```js
const cardFields = paypal.CardFields();

if (cardFields.isEligible()) {
    cardFields.render('#card-field-container')
}
```

https://github.com/paypal/paypal-sdk-spec/blob/1b9b34cb379b836fd92ed5b20b252e24592ab30f/spec/client/javascript/components/card-fields/eligibility.md

bywood: @gregjopa `cardFields.isEligible()` does return a boolean value. The [Zoid documentation](https://github.com/krakenjs/zoid/blob/main/docs/api/create.md#eligible--props---string--any----boolean) appears to be outdated, but in the code you can see that it expects `eligible` to return an object:

https://github.com/krakenjs/zoid/blob/3974c52a880e8b7a72201c9ad205b576611e7c65/src/component/component.js#L128-L131

The `eligible` property of that object, which is a boolean, is what is returned from the `isEligible` call:

https://github.com/krakenjs/zoid/blob/3974c52a880e8b7a72201c9ad205b576611e7c65/src/component/component.js#L481-L482

gregjopa: Hey @bywood, great points. I didn't realize `eligible()` is an interface/pattern from zoid. 

I do see the buttons doing the same thing here: https://github.com/paypal/paypal-checkout-components/blob/main/src/zoid/buttons/component.jsx#L107 

Thanks for teaching me about zoid here 😄

PR #1979, opened 2022-07-28
fix: add clientMetadataID prop to standard card fields

allen-tong:
### Description

Adding the clientMetadataID prop to the inline guest component to send client-metadata-id as a query param. The inline guest server will use this, if present, to initialize fraudnet.

We want to use client-metadata-id to correctly associate RDA data.
https://engineering.paypalcorp.com/jira/browse/PPBLRAC-44501

Testing locally: add a data attribute like `data-client-metadata-id="test-id"` to the script element and check that the request to /inline includes the query param `client-metadata-id=test-id`.

❤️  Thank you!


mnicpt:
@allen-tong So you don't need this as a button prop as well?

mnicpt (approved):
LGTM if not needed for SPB as well.

allen-tong:
@mnicpt the button prop already exists here: https://github.com/paypal/paypal-checkout-components/blob/main/src/zoid/buttons/component.jsx#L578

PR #1984, opened 2022-08-05
applepay - oncancel

andyfen:
changes
https://github.com/paypal/paypal-smart-payment-buttons/pull/452

### DIFF
--- a/src/zoid/buttons/util.js
+++ b/src/zoid/buttons/util.js
@@ -192,6 +192,11 @@ export function applePaySession() : ?ApplePaySessionConfigRequest {
                 listeners.paymentauthorized({ payment });
             };
 
+            // eslint-disable-next-line  unicorn/prefer-add-event-listener
+            session.oncancel = () => {
+                listeners.oncancel();
+            };
+                        
             return {
                 addEventListener: (name, handler) => {
                     listeners[name] = handler;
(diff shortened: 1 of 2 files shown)


wsbrunson (approved):

src/zoid/buttons/util.js
```diff
@@ -214,8 +219,7 @@ export function applePaySession() : ?ApplePaySessionConfigRequest {
                     session.completePayment(newUpdate);
                 },
                 begin: () => session.begin(),
-                abort: () => session.abort(),
-                oncancel: () => session.oncancel()
```
mnicpt, line 218: @andyfen Why are we refactoring this?  I think the issue is that when I look at SPB, the `oncancel` callback isn't being brought in from the response: https://github.com/paypal/paypal-smart-payment-buttons/blob/c12f7f31540246d593e53e3bbd4e68d29218817b/src/payment-flows/applepay/applepay.js#L345

mnicpt (changes requested):
@andyfen Let's get together and talk through this and issue 2.

src/zoid/buttons/util.js
mnicpt: I see what happened.  This wasn't here before and was added.  Interesting this doesn't work though.  This should be fine.  Did we update the Apple Pay version.  I don't recall `abort` being supported last year.

mnicpt (approved):

PR #1989, opened 2022-08-16
Update APM buttons with new default colors

ericyupaypal:
### Description
- Update default button colors for APMs to match individual payment scheme branding

- https://engineering.paypalcorp.com/jira/browse/DTALTPAY-993

screencapture-localhost-paypal-9001-webpack-dev-server-demo-dev-index-htm-2022-09-22-10_17_38

- Dependency on `paypal-sdk-logos` PR, https://github.com/paypal/paypal-sdk-logos/pull/68, to use new APM logos created for these new background colors. If used with old version of paypal-sdk-logos, logos may be illegible on new colored button backgrounds.

- AltPay Experience
- Checkout SDK

❤️  Thank you!


jscheinhorn (approved):

wsbrunson:
Image 2022-08-22 at 16 43 41

Does the Boleto test look okay?

jscheinhorn:
Is that typical for the screenshots from the test to have that shaded box?

wsbrunson:
> Is that typical for the screenshots from the test to have that shaded box? I see it on most of the test screenshots https://github.com/paypal/paypal-checkout-components/pull/1989/files

It might be graying out the label for the sake of flakeyness but we will verify that.

ericyupaypal:
> 

I'll double check this. I would expect the logo to show up and the label greyed out. It's only showing the greyed out label, so will look into why the logo is not rendering. I am wondering if it's due to how we have both ```FUNDING.BOLETO``` and ```FUNDING.BOLETOBANCARIO``` in the test config.

ericyupaypal:
I didn't realize both ```FUNDING.BOLETO``` and ```FUNDING.BOLETOBANCARIO``` were used. Before, I just enabled the new styling on ```FUNDING.BOLETO```, which led to the bad screenshot test for ```FUNDING.BOLETOBANCARIO```. Updated and committed changes to support ```FUNDING.BOLETOBANCARIO``` also now.

wsbrunson:
> I didn't realize both FUNDING.BOLETO and FUNDING.BOLETOBANCARIO were used

I think `BOLETOBANCARIO` replaced `BOLETO` but the work to remove `BOLETO` from the codebase still needs to be completed

dtjones404 (approved):
Looks great, thanks!

PR #1993, opened 2022-08-24
chore: upgrade to latest puppeteer

westeezy:
Draft PR to see if we can get screenshots to work on both m1 and intel macbooks


wsbrunson (approved):
hell ya. heads up with this PR: https://github.com/paypal/paypal-checkout-components/pull/1989

westeezy:
This will be tough cause images are generated ever so slightly differently on m1 architectures. It appears to be anywhere from 1-5% drift

PR #2010, opened 2022-09-22
Second Button Experiment Logging

onghaim:
### Description

Adding logging support to track the result of the second button experiments run in xobuyernodeserv (already merged).

https://engineering.paypalcorp.com/jira/browse/DTXOWALCAP-428

The main point of inserting the logging here is because we have access to the button session id, which is not present in the xoybuyer segment that contains the elmo logic.

❤️  Thank you!

### DIFF
--- a/src/funding/paylater/config.jsx
+++ b/src/funding/paylater/config.jsx
@@ -1,7 +1,7 @@
 /* @flow */
 /** @jsx node */
 
-import type { FundingEligibilityType } from '@paypal/sdk-client/src';
+import { FundingEligibilityType, getLogger, getLocale, getSessionID} from '@paypal/sdk-client/src';
 import { FUNDING } from '@paypal/sdk-constants/src';
 import { node, Style } from '@krakenjs/jsx-pragmatic/src';
 import { PPLogo, LOGO_COLOR } from '@paypal/sdk-logos/src';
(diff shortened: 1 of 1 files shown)


src/funding/paylater/config.jsx
```diff
@@ -49,6 +56,19 @@ function getLabelText(fundingEligibility : FundingEligibilityType) : ?string {
         labelText = '4X PayPal';
     }
 
+    if(getLocale().includes("US")) {
+        const payload = {
+            state_name: "elmo_check",
+            transition_name: "process_elmo_check",
+            experimentation_experience: SECOND_BTN_EXP.EXPERIMENTATION_EXPERIENCE,
+            experimentation_treatment: paylater?.products?.payIn4?.reasons.includes(SECOND_BTN_EXP.REASON) ? SECOND_BTN_EXP.TREATMENT_ID : SECOND_BTN_EXP.CONTROL_ID,
+            context_id: getSessionID(),
```
matthewonghai13, line 65: From what I understand, this is the main session ID that people refer to as the "button session ID" (please correct me if I'm wrong on this

src/funding/paylater/config.jsx
```diff
@@ -49,6 +56,19 @@ function getLabelText(fundingEligibility : FundingEligibilityType) : ?string {
         labelText = '4X PayPal';
     }
 
+    if(getLocale().includes("US")) {
```
matthewonghai13, line 59: please let me know if there is a cleaner way to check that this instance of the buttons are rendering in a US transaction

elizabethmv: Locale doesn't resolve to the true buyer country, so this can't be used as a reliable check

src/funding/paylater/config.jsx
```diff
@@ -49,6 +56,19 @@ function getLabelText(fundingEligibility : FundingEligibilityType) : ?string {
...
+        const payload = {
+            state_name: "elmo_check",
+            transition_name: "process_elmo_check",
+            experimentation_experience: SECOND_BTN_EXP.EXPERIMENTATION_EXPERIENCE,
+            experimentation_treatment: paylater?.products?.payIn4?.reasons.includes(SECOND_BTN_EXP.REASON) ? SECOND_BTN_EXP.TREATMENT_ID : SECOND_BTN_EXP.CONTROL_ID,
```
elizabethmv, line 64: This check can't be used, `reasons` isn't part of the query and also can't be accessed externally

elizabethmv (changes requested):
The button session ID is passed in funding eligibility, so we should have access to it in the Pay Later resolver. Let's pair up on getting that added to the experiment in GraphQL

PR #2029, opened 2022-11-09
Disable vaulted card menu for token type

elizabethmv:
### Description
This PR disables the vaulted card menu for in the scenario that is not currently supported for different vault token types and the ID token from the merchant.

We don't want to display a bad UX, where there is an option to unlink your card but actually

- Vault a card
- Create a client token and pass it to the SDK
- See vaulted button with menu
- Pass and ID token or both the client token and the ID token
- See vaulted button without menu

The jest screenshots reflect the behavior

N/A

N/A


sioked (approved):
Looks good, screenshot tests are useful here.

wsbrunson (approved):

PR #2046, opened 2022-12-14
feat: save action validates inputs

cgdibble:

DTNOR-546

These are part of the progressive implementation of the Save action. These changes enable the action to:
- be initialized
- accept config from merchant
- validate that config
- return skeleton of the action (implemented in tickets to come)

Additionally, this PR adds unit testing and the tools to use it to this repository.

n/a

n/a

n/a

❤️  Thank you!

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "reinstall": "rimraf flow-typed && rimraf node_modules && npm install && flow-typed install",
     "release": "./scripts/publish.sh",
     "start": "npm run webpack -- --progress --watch",
-    "test": "npm run jest-ssr && npm run karma && npm run jest-screenshot",
+    "test": "npm run test:unit && npm run jest-ssr && npm run karma && npm run jest-screenshot",
     "test:unit": "vitest",
     "typecheck": "npm run flow-typed && npm run flow",
     "version": "./scripts/version.sh",
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "release": "./scripts/publish.sh",
     "start": "npm run webpack -- --progress --watch",
     "test": "npm run jest-ssr && npm run karma && npm run jest-screenshot",
+    "test:unit": "vitest",
     "typecheck": "npm run flow-typed && npm run flow",
     "version": "./scripts/version.sh",
     "webpack": "babel-node $(npm bin)/webpack",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -27,6 +27,7 @@
     "release": "./scripts/publish.sh",
     "start": "npm run webpack -- --progress --watch",
     "test": "npm run jest-ssr && npm run karma && npm run jest-screenshot",
```
wsbrunson, line 29:
```suggestion
    "test": "npm run test:unit && npm run jest-ssr && npm run karma && npm run jest-screenshot",
```

wsbrunson (approved):

src/actions/save/index.js
```diff
@@ -1,17 +1,46 @@
...
+  createVaultSetupToken: () => Promise<string>,
+  onApprove: ({| vaultSetupToken: string |}) => void,
 |};
 
-export type SaveAction = (SaveActionHandlers) => ({|
+export type SaveAction = (SaveActionConfig) => ({|
   type: "save",
-  ...SaveActionHandlers,
+  /* TODO: 
+      - We need to define how paymentSourceDetails is typed here
```
cgdibble, line 12: A subsequent ticket will handle the definition of paymentSourceDetails. We'll type out the current expectation based on card fields data then, in the future, provide others as options.

src/actions/save/index.js
```diff
@@ -1,17 +1,46 @@
...
+      throw new ValidationError(`Save action is missing the required '${inputProp}' callback`)  
+    }
+  }
+}
+
+/**
+ * Creating a Save action allows us to validate initial inputs from the merchant, and then return the resulting object. 
+ */
+export const createSaveAction: SaveAction = (config: SaveActionConfig) => {
+  const { createVaultSetupToken } = config;
```
thebastedo, line 38: Would we want to validate prior to destructuring anything off of the config? Maybe it's a bit early to get that picky about the code?

cgdibble: I need to push some changes, but yeah this is just left from when I did it here initially and destructured the props. I forgot to delete this line when I moved validation to a function, so yeah it'll be removed here shortly

thebastedo (approved):

PR #2047, opened 2022-12-15
feat: save uses merchant vault token callback to get token

cgdibble:

DTNOR-547
We need to use the merchant provided callback to retrieve the _empty_ setup token. While it's as simple as us calling an async function and resolving it, the callback itself is defined by the merchant and needs to hit their own API so they can hit the PayPal vaulting API to create an empty setup token.

n/a
n/a
n/a

❤️  Thank you!

### DIFF
--- a/src/actions/save/index.js
+++ b/src/actions/save/index.js
@@ -23,7 +23,7 @@ export type onErrorCallback = (error: string) => void
 
 export type SaveAction = (SaveActionConfig) => ({|
   type: "save",
-  save: (onError: onErrorCallback, paymentSourceDetails: CardDetails) => ZalgoPromise<void> | Promise<void>,
+  save: (onError: onErrorCallback, paymentSourceDetails: CardDetails) => ZalgoPromise<void>,
 |});
 
 /**
@@ -59,9 +59,9 @@ export const createSaveAction: SaveAction = (config: SaveActionConfig) => {
 
       return ZalgoPromise.try(() => {
         createVaultSetupToken()
-        .then((emptySetupToken) => {
+        .then((/* emptySetupToken */) => {
           // take token and call our PP endpoint with it
-        }).catch((error) => {
+        }).catch((/* error */) => {
           // TODO: Let's make sure we stringify this error safely/idiomatically - Do we define errors elsewhere?
           return onError("Unable to retrieve setup token from 'createVaultSetupToken'")
         })
--- a/src/actions/save/index.js
+++ b/src/actions/save/index.js
@@ -19,12 +19,11 @@ type SaveActionConfig = {|
   onApprove: ({| vaultSetupToken: string |}) => void,
 |};
 
+export type onErrorCallback = (error: string) => void
+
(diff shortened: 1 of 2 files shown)


test/unit/actions/save.test.js
```diff
@@ -1,9 +1,10 @@
 /* @flow */
 
 import { ZalgoPromise } from "@krakenjs/zalgo-promise/dist/zalgo-promise";
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi } from "vitest";
 
 import { createSaveAction } from "../../../src/actions/save"
+import type { onErrorCallback } from "../../../src/actions/save"
```
cgdibble, line 7: Linting breaks on having this import done twice, but Flow breaks if I try to combine them. I'm not sure the solution. We _could_ have an explicit `types.js` file wherein we define the types and use them in both places. Thoughts? I'm sure there is some Flow-fu I'm missing here, too.

thebastedo (approved):

src/actions/save/index.js
```diff
@@ -51,10 +50,22 @@ export const createSaveAction: SaveAction = (config: SaveActionConfig) => {
...
+
       return ZalgoPromise.try(() => {
-        // basics for typing requirements. Implementation to come in next ticket.
-      })
+        createVaultSetupToken()
+        .then((/* emptySetupToken */) => {
+          // take token and call our PP endpoint with it
+        }).catch((/* error */) => {
+          // TODO: Let's make sure we stringify this error safely/idiomatically - Do we define errors elsewhere?
+          return onError("Unable to retrieve setup token from 'createVaultSetupToken'")
```
siddy2181, line 69: How can we make this simpler for merchants to parse?
Eg: we would want to send `errorCode: INTERNAL_SERVER`,`type:INVALID_TOKEN`&`errorMessage: createVaultSetupToken: Unable to retrieve setup token`. 
How do we pass this in the best possible manner?

siddy2181 (approved):

src/actions/save/index.js
```diff
@@ -51,10 +53,22 @@ export const createSaveAction: SaveAction = (config: SaveActionConfig) => {
 
   return {
     type: "save",
-    save: () => {
+    save: (onError, paymentSourceDetails) => {
+      const { createVaultSetupToken } = config;
+    
+      if (!onError || !paymentSourceDetails) {
+        throw new ValidationError("Missing args to #save")
```
bywood, line 60: Should this return a rejected ZalgoPromise instead? (Assuming there is such a thing. Forgive my ZalgoPromise ignorance.)

cgdibble: Honestly, uh, maybe? There are no docs on ZalgoPromise so I'm really not sure. But, I think you may be right, it should probably be an explicit reject instead.

src/actions/save/index.js
cgdibble: This is a great question. I'll raise that to the team in Slack.

wsbrunson (approved):

PR #2051, opened 2022-12-16
Add a11y attributes to MenuButton

nbierdeman:
### Description

This PR adds a11y attributes to the PayPal menu button for all relevant flows:

https://user-images.githubusercontent.com/20399044/208185874-373401b3-90cf-48b4-9915-66b31f07dc2d.mov

### DIFF
--- a/src/ui/buttons/button.jsx
+++ b/src/ui/buttons/button.jsx
@@ -263,7 +263,7 @@ export function Button({ fundingSource, style, multiple, locale, env, fundingEli
                 <Spinner />
             </div>
 
-            { shouldShowWalletMenu ? <MenuButton textColor={ textColor } /> : null }
+            { shouldShowWalletMenu ? <MenuButton textColor={ textColor } content={ content } /> : null }
         </div>
     );
 }
--- a/src/ui/buttons/menu-button/menu-button.jsx
+++ b/src/ui/buttons/menu-button/menu-button.jsx
@@ -5,10 +5,12 @@ import { node, type ChildType, Style } from '@krakenjs/jsx-pragmatic/src';
 
 import { ATTRIBUTE, TEXT_COLOR } from '../../../constants';
 import { Chevron } from '../../chevron';
+import type { ContentType } from '../../../types';
 
 import css from './menu-button.scoped.scss';
 
-export function MenuButton({ textColor = TEXT_COLOR.BLACK } : {| textColor? : $Values<typeof TEXT_COLOR> |} = { }) : ChildType {
+export function MenuButton({ textColor = TEXT_COLOR.BLACK, content } : {| textColor? : $Values<typeof TEXT_COLOR>, content : ?ContentType |} = { }) : ChildType {
+    const labelText = content?.moreOptions;
     return (
         <Style css={ css }>
             <div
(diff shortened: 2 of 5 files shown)


src/ui/buttons/menu-button/menu-button.jsx
```diff
@@ -5,18 +5,23 @@ import { node, type ChildType, Style } from '@krakenjs/jsx-pragmatic/src';
 
 import { ATTRIBUTE, TEXT_COLOR } from '../../../constants';
 import { Chevron } from '../../chevron';
+import type { ContentType } from '../../../types';
 
 import css from './menu-button.scoped.scss';
 
-export function MenuButton({ textColor = TEXT_COLOR.BLACK } : {| textColor? : $Values<typeof TEXT_COLOR> |} = { }) : ChildType {
+export function MenuButton({ textColor = TEXT_COLOR.BLACK, content } : {| textColor? : $Values<typeof TEXT_COLOR>, content : ?ContentType |} = { }) : ChildType {
+    const labelText = content?.moreOptions;
```
ravishekhar, line 13:
```suggestion
    const labelText = content?.moreOptions || '';
```

src/ui/buttons/button.jsx
```diff
@@ -263,7 +263,7 @@ export function Button({ fundingSource, style, multiple, locale, env, fundingEli
                 <Spinner />
             </div>
 
-            { shouldShowWalletMenu ? <MenuButton textColor={ textColor } /> : null }
+            { shouldShowWalletMenu ? <MenuButton textColor={ textColor } content={ content } /> : null }
```
jshawl, line 266: will content ever be used for anything other than `aria-label`? I was thinking maybe:

```
<MenuButton aria-label={content?.moreOptions} ... />
```

ravishekhar (approved):
looks good, just few nit pick

jshawl (approved):
Looks good! It would be rad to add a unit test but at first glance I didn't see any existing coverage for the menu button specifically - so we could always tackle in the future.

One non-blocking comment/suggestion about simplifying the label and type from `ContentType` to `string`.  But LGTM :+1:

src/ui/buttons/button.jsx
nbierdeman: Thanks for the feedback, @jshawl! At the moment, `content` isn't used for anything in `MenuButton` except the aria-label, but that might not always be the case. I'm not seeing any existing patterns where individual `content` properties are passed down to child components, but instead, I do see a pattern of passing down the entire `content` object. I'm leaning towards keeping it as is, unless you can think of any other reason for only passing down `content.moreOptions` to `MenuButton`.

test/integration/tests/button/eligibility.js
```diff
@@ -283,6 +283,7 @@ describe('paypal on file eligibility', () => {
                 onRender: () => {
                     assert.equal(getElementRecursive('.balance .paypal-button-text').innerHTML, 'Balance &amp;');
                     assert.equal(getElementRecursive('.fi-label .paypal-button-text').innerHTML, '••1234');
+                    assert.equal(getElementRecursive('.menu-button').getAttribute('aria-label'), 'More options');
```
jshawl, line 286: 💯

PR #2061, opened 2022-12-22
WIP Feat/update vault setup token

thebastedo:

❤️  Thank you!


cgdibble (changes requested):
Requesting changes until we have proper unit tests in place.

src/actions/save/index.js
```diff
@@ -51,22 +54,64 @@ const validateSaveConfig = (config: SaveActionConfig): void => {
...
+        logger.error('pp_sdk_actions_save_error', stringifyError(error))
+        onError(error.message)
+      }
+
       return ZalgoPromise.try(() => {
         createVaultSetupToken()
-        .then((/* emptySetupToken */) => {
+        .then(({ vaultSetupToken }) => {
+          // const vaultUrl = `${ getPayPalDomain() }/v3/vault/setup-tokens/${vaultSetupToken}/update`
+          const vaultUrl = `https://msmaster.qa.paypal.com:18582/v3/vault/setup-tokens/${vaultSetupToken}/update`
```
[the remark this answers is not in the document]
gregjopa: Good question. We should avoid hardcoding this url. Please continue to use the [getPayPalDomain() function](https://github.com/paypal/paypal-sdk-client/blob/main/src/global.js#L41). It uses a global variable that gets filled in at build time by webpack and has support for testing in all non-prod environments. We can discuss more internally if needed.

src/actions/save/index.js
```diff
@@ -51,22 +54,64 @@ const validateSaveConfig = (config: SaveActionConfig): void => {
...
+          // call the vault api to update the setup token with our payment details
+          return request({
+            method: 'post',
+            url: vaultUrl,
+            headers: {
+              // Figure out where we can get authToken
+              'Authorization': `Basic ${lowScopedAccessToken}`,
```
cgdibble, line 84: How is this type of thing handled already within the SDK?

src/zoid/card-fields/component.jsx
```diff
@@ -289,6 +289,10 @@ export const getCardFieldsComponent : () => CardFieldsComponent = memoize(() : C
         },
 
         props: {
+            action: {
+                type: 'object',
+                required: true
```
cgdibble, line 294: I was referencing this PR for config stuff and noted that this shouldn't be required, right? It's a bit of a confusing scenario because if no `createOrder`, etc. callbacks are provided, then it _is_ required. 🤔 hmm that's tricky.

wsbrunson: you're right, we'll have to use Zoid's `validate` prop to check validity for the action prop: https://github.com/krakenjs/zoid/blob/main/docs/api/prop-definitions.md#validate

src/actions/save/index.js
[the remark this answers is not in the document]
wsbrunson: this was just hard-coded for now because we were testing locally but needed the msmaster domain

src/actions/save/index.js
wsbrunson: https://github.com/paypal/paypal-sdk-client/blob/main/src/api.js#L40

PR #2072, opened 2023-01-12
feat: support >= 250px button widths for PPoF

nbierdeman:
### Description

This PR adds support for >= 250px button widths for PPoF.

### Screenshots

Screen Shot 2023-01-12 at 9 34 34 AM

### Related Changes

https://github.com/paypal/paypal-smart-payment-buttons/pull/535


gregjopa:
Hey @nbierdeman just to clarify, currently the max width for the vaulted button is 251px and this change updates it to support 250px. Is that correct?

nbierdeman:
> ...just to clarify, currently the max width for the vaulted button is 251px and this change updates it to support 250px. Is that correct?

@gregjopa, there are two button width issues that this PR is addressing:

- < 300px the FI logo disappears during 2nd render
- at exactly 250px the menu button doesn’t render

The minimum vault button width is correctly set to 250px, however, the `@media` `max-width` CSS rule is also, which isn't correct. It needs to be applied at 249px and below, thus the `- 1` code change (as opposed to trying to rename the variable and setting it to `249`).

wsbrunson (approved):

src/ui/buttons/styles/button.js
```diff
@@ -126,10 +126,7 @@ export const buttonStyle = `
         margin-top: 10px;
     }
 
-    @media only screen and (max-width: ${ MIN_VAULT_BUTTON_WIDTH }px) {
-        .menu-button {
-            display: none;
-        }
```
nbierdeman, line 132: @ravishekhar @jshawl @gregjopa @wsbrunson, it appears this isn't used. When the menu button isn't visible (when the button container width is <= 249px), it's being pushed out of view by the paypal button (since its width is the same as the button container width; see line 136/133 being applied to the `paypal-button` class below). The menu button is still in the DOM, however, with `display: inline-block`, and it can be right clicked and scrolled into view. Should we go ahead and remove this then?

ravishekhar (approved):
lgtm 👍

PR #2073, opened 2023-01-19
feat: add 250px rect & pill PPoF screenshots

nbierdeman:
### Description

This PR adds a couple of screenshots for more coverage for PPoF. They both have `Balance &` + FI label rendered with the generic card logo, which is wider than the generic bank logo, but differ in button shape (rectangle vs pill).


jshawl (approved):
looks great! :shipit:

PR #2085, opened 2023-02-08
interface changes for standalone card vault

wsbrunson:
add fields to CardFields component for standalone vault, like action and userIDToken

### DIFF
--- a/src/zoid/card-fields/component.jsx
+++ b/src/zoid/card-fields/component.jsx
@@ -8,7 +8,7 @@ import { create, type ZoidComponent } from '@krakenjs/zoid/src';
 import type { CrossDomainWindowType } from '@krakenjs/cross-domain-utils/src';
 import { memoize, uniqueID } from '@krakenjs/belter/src';
 import { getLocale, getEnv, getSDKMeta, getDisableCard, getPayPalDomain, getClientID, getDebug, getCurrency, getIntent,
-    getCommit, getVault } from '@paypal/sdk-client/src';
+    getCommit, getVault, getUserIDToken } from '@paypal/sdk-client/src';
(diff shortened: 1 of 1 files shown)


bywood (approved):

src/zoid/card-fields/component.jsx
```diff
@@ -316,7 +328,13 @@ export const getCardFieldsComponent : () => CardFieldsComponent = memoize(() : C
                     queryParam: true,
                     required:   false,
                     value:      ({ props }) => props.parent.props.branded
-                }
+                },
+
+                userIDToken: {
```
cgdibble, line 333: Non-blocking at all but...what is this for in relation to standalone vaulting?

src/zoid/card-fields/component.jsx
```diff
@@ -150,6 +151,17 @@ export const getCardFieldsComponent : () => CardFieldsComponent = memoize(() : C
             },
 
             props: {
+                action: {
+                    type: 'object',
+                    value: ({props}) => {
+                        if (props.action) {
+                            return props.action
+                        } else {
+                            return props.parent.props.action
```
cgdibble, line 160: Did we sort out why this was going on or is it kinda hand-wavy "zoid stuff sometimes"?

cgdibble (approved):

elizabethmv (approved):

src/zoid/card-fields/component.jsx
wsbrunson: this is because we don't have a "leader" or "facilitator" iframe with Zoid. We define the parent component but we don't actually render it so it never creates an iframe. I think the "leader" iframe becomes a random child component that does have access to its parent. Its really weird and we should really re-think using Zoid for CardFields once we have time to breathe

PR #2115, opened 2023-02-27
Harden logos implementation for supporting both external and inline images

dtjones404:
[DTPPSDK-1083](https://paypal.atlassian.net/browse/DTPPSDK-1083)

* Removed code related to logo cdn experiment
* Removed experiment wrapper function from logos
* Updated logos to use new `paypal-sdk-logos` components

### Dependent Changes
Waiting on https://github.com/paypal/paypal-sdk-logos/pull/98

❤️  Thank you!


gregjopa:
@dtjones404 awesome work with this! 💯 

I am wondering if we can add a note to the README or a code comment explaining this behavior. Something like:

### Logo Strategy with paypal-sdk-logos

Our usage of svg logos is optimized for performance. Here's how it works for the two-phased render for the Buttons component:
1. **First Render** - Buttons are first rendered on the client-side inside an `<iframe>` tag using the code bundled inside the JS SDK script. This `<iframe>` tag has no `src` attribute at this point in the rendering life cycle. We leverage the `__WEB__` global variable to determine if we are rendering client-side or server-side. The `__WEB__` global variable will equal `true` when rendering client-side. To minimize the bundle size of the JS SDK script, we load these svg logos from the www.paypalobjects.com CDN as external images (ex:  `<PayPalLogoExternalImage />`).
2. **Second Render** - The second phase of rendering happens on the server-side. This `<iframe>` tag is fully rendered by setting the `src` attribute and making an http request to the server to populate the contents which are securely hosted by PayPal's servers on www.paypal.com. The same code in this repo is used to render the buttons on the server-side. To ensure there are no issues with the logos, we have decided to inline the svg code for this server-side render (ex: `<PayPalLogoInlineSVG />`). The `__WEB__` global variable is set to false to control what code executes during the server render.

gregjopa (approved):

src/funding/credit/config.jsx
```diff
@@ -35,18 +37,34 @@ export function getCreditConfig(): FundingSourceConfig {
...
+            )}
             <Space />
           </span>
-          {enableLogoCDNExperiment(CreditLogo, { locale, logoColor })}
+          {__WEB__ ? (
+            <CreditLogoExternalImage locale={locale} logoColor={logoColor} />
+          ) : (
+            <CreditLogoInlineSVG locale={locale} logoColor={logoColor} />
+          )}
         </Fragment>
```
xuanwan05, line 68: should we do something like this? so we only need to check `__WEB__` once
{__WEB__ ? (
<Fragment>
  <PPLogoExternalImage logoColor={logoColor} />
  <Space />
  <span optional>
    <PayPalLogoExternalImage logoColor={logoColor} />
    <Space />
  </span>
  <CreditLogoExternalImage locale={locale} logoColor={logoColor} />
</Fragment>
) : (
<Fragment>
  <PPLogoInlineSVG logoColor={logoColor} />
  <Space />
  <span optional>
    <PayPalLogoInlineSVG logoColor={logoColor} />
    <Space />
  </span>
  <CreditLogoInlineSVG locale={locale} logoColor={logoColor} />
</Fragment>
)
}

dtjones404: Good call!  Refactored.

PR #2144, opened 2023-03-30
chore: remove __FUNDING_ELIGIBILITY__ global

dtjones404:
### Description

Removed the `__FUNDING_ELIGIBILITY__` global variable from `checkout-components`, as well as the tree-shaking logic it enabled at build-time.
Removing the `__FUNDING_ELIGIBILITY__` global variable from the build process is a precursor for caching changes.

❤️ Thank you!

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,7 +18,6 @@ module.exports = {
     __HOST__: true,
     __PATH__: true,
     __COMPONENTS__: true,
-    __FUNDING_ELIGIBILITY__: true,
     __INLINE_CHECKOUT_ELIGIBILITY__: true,
   },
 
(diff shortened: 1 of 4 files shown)


gregjopa:
Hey @dtjones404, I think we need to increase our bundle size a little bit to support this change. Can you update it?

dtjones404:
> Hey @dtjones404, I think we need to increase our bundle size budget a little bit to support this change. Can you update it?

Looks like it's the percentage increase that's causing the bundle-size check to fail, rather than the actual size.  I can approve the increase through bundlemon but I don't think I have the ability to increase the limits.

gregjopa (approved):
Looks great! Excited to see the performance improvements this enables with limiting the webpack build we do at runtime.

wsbrunson (approved):

PR #2150, opened 2023-04-20
mark card button ineligible when using card-fields

wsbrunson:
### Description

the black bcdc button or card button should not be eligible in the smart stack when using the new Hosted Card Fields component (card-fields)

I added a description to explain the purpose of the order of the if statements when checking for eligibility and a test to verify

### DIFF
--- a/src/funding/card/config.test.jsx
+++ b/src/funding/card/config.test.jsx
@@ -1,0 +1,120 @@
+/* @flow */
+import { describe, test, expect } from "vitest";
+import { FUNDING, COMPONENTS } from "@paypal/sdk-constants/src";
+
+import { getCardConfig } from "./config";
+
+const getEligibility = ({
+  eligible = true,
+  branded = false,
+  vendors = {},
+} = {}) => ({
+  card: {
+    eligible,
+    branded,
+    vendors,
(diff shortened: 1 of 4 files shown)


src/funding/card/config.test.jsx
```diff
@@ -0,0 +1,120 @@
...
+  },
+  paypal: {
+    instruments: [],
+  },
+  venmo: {
+    instruments: [],
+  },
+});
+
+describe("card eligibility", () => {
```
elizabethmv, line 34: Can we add a test for if `card-fields` is requested and there is a vaulted card as another safe guard for any potential changes to the ordering?

elizabethmv (approved):
Lgtm! One minor suggestion

src/funding/card/config.test.jsx
wsbrunson: ooooh good call

wsbrunson: these tests kinda suck, its easy to get them wrong. We need something like Property Testing to really make sure every edge case is covered

siddy2181 (approved):
Looks good!!

PR #2158, opened 2023-05-03
move label content to string map

jshawl:

❤️ Thank you!

### DIFF
--- a/src/funding/content.jsx
+++ b/src/funding/content.jsx
@@ -26,12 +26,23 @@ export type ContentMap = {
   |},
 };
 
+const componentLabelText = {
+  ar: {
+    Checkout: "السداد بواسطة",
+    Subscribe: "اشتراك",
+    SaferTag: "الطريقة الأسهل والأكثر أماناً في الدفع",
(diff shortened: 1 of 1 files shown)


nbierdeman:
@jshawl, this is looking good so far. I'm wondering if you've been able to validate the button content after these changes, especially the ones with line breaks. Are there existing screenshots for each of these label variations?

jshawl:
Thanks, nate! It looks like the break tags are broken in sandbox right now:

labels

```html
<script src="https://www.paypal.com/sdk/js?client-id=test&currency=USD&locale=es_US"></script>
<script>
  paypal.Buttons({
    style: {
      label: 'installment'
    },
  })
</script>
```

I took a look at https://github.com/paypal/paypal-checkout-components/blob/main/test/percy/files/buttonConfigs.json and we could add in these localized and labelized variations to establish a percy baseline, wdyt?

nbierdeman:
@jshawl, yeah, I do think establishing a Percy baseline would be ideal. This would be a huge win for future generations of button label text developers. 😁

PR #2182, opened 2023-06-10
feat(venmo): venmo component

mnicpt:
### Description
Used for Venmo Web experience.
https://paypal.atlassian.net/browse/DTCHKINT-1534

❤️ Thank you!


luciagolovin (approved):

src/zoid/venmo/component.jsx
```diff
@@ -0,0 +1,310 @@
...
+          type: "string",
+          required: false,
+          default: getClientMetadataID,
+          queryParam: "client-metadata-id",
+        },
+
+        onAuth: {
+          type: "function",
+          required: false,
+          trustedDomains: ["https://www.paypal.com", "https://venmo.com"],
```
nbierdeman, line 250: Out of curiosity, how's this going to work? I was originally assuming it would be something like `isTrustedDomain: isPayPalTrustedDomain()`.

mnicpt: @nbierdeman How it works is one answer but it does cause me to think about the allowed types here. The `isPayPalTrustedDomain()` is a convenience function that uses RegEx to compare the trusted domain regex and should be used here. It is very possible to call each regex individually as well. The issue I am seeing here though is that the `trustedDomains` prop should verify against a regex rather than a string: `prop.trustedDomains.includes(getDomain(window))`. This will give us flexibility across environments. I'll make that change in void. Thanks for catching!

wsbrunson (approved):

gregjopa:
It looks like the Percy screenshots are failing. @dtjones404 would you be able to take a look when you get a free moment?

dtjones404:
> It looks like the Percy screenshots are failing. @dtjones404 would you be able to take a look when you get a free moment?

Looks like something in `paypal-sdk-clientv4.0.173` is causing the tests to break.  Possibly the new global?  Still investigating.  In any case I don't think it's an issue with this PR, since `main` is having the same issue currently.

dtjones404:
Looks like no changes with `sdk-client` pinned to `4.0.172`

https://percy.io/bbbd4e9f/paypal-checkout-components/builds/28192291/unchanged/1567226914?browser=safari&viewLayout=overlay&viewMode=new&width=1280

gregjopa:
Thanks for taking a look @dtjones404! My initial guess is we also need to add the new global to this percy globals file: https://github.com/paypal/paypal-checkout-components/blob/main/test/percy/globals.js#L143

dtjones404:
> Thanks for taking a look @dtjones404! My initial guess is we also need to add the new global to this percy globals file: https://github.com/paypal/paypal-checkout-components/blob/main/test/percy/globals.js#L143

💯 Looks like that fixes the issue!

mnicpt:
@dtjones404 Thank a mil for unblocking! 🚀

src/zoid/content.js
```diff
@@ -0,0 +1,162 @@
+/* @flow */
```
mnicpt, line 1: Refactored out of checkout to keep size down and be reused by both PayPal and Venmo.

PR #2188, opened 2023-06-15
[DTPPSDK-1244] Add Flag to Disable Cookies for /smart/buttons

xuanwan05:
### Description

Add Flag to Disable Cookies for /smart/buttons

Screenshot 2023-06-15 at 11 21 41 AM

❤️ Thank you!

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "@krakenjs/zoid": "^10.0.0",
     "@paypal/common-components": "^1.0.35",
     "@paypal/funding-components": "^1.0.31",
-    "@paypal/sdk-client": "^4.0.173",
+    "@paypal/sdk-client": "^4.0.175",
     "@paypal/sdk-constants": "^1.0.128",
     "@paypal/sdk-logos": "^2.2.6"
   },
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -817,11 +817,7 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
         type: "boolean",
         queryParam: true,
         required: false,
-        value: () => {
-          if (getDisableSetCookie()) {
-            return true;
-          }
-        },
+        value: getDisableSetCookie,
       },
     },
   });


src/zoid/buttons/component.jsx
```diff
@@ -811,6 +812,17 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+
+      disableSetCookie: {
+        type: "boolean",
+        queryParam: true,
+        required: false,
+        value: () => {
+          if (getDisableSetCookie()) {
+            return true;
+          }
+        },
```
wsbrunson, line 824: could this be refactored as:
```suggestion
        value: () => getDisableSetCookie(),
```

gregjopa: From what I remember with zoid props that use `queryParam: true`, a boolean value will always be in the query params (`disableSetCookie=true` and `disableSetCookie=false`). If we return `undefined` it will not be in the query param (`disableSetCookie=true` or nothing). I think either solution works, but if we want to minimize the size of the query string, we should return `undefined` when false.

gregjopa: Actually I may be wrong about this behavior. Ignore the above comment. @xuanwan05 is going to test this out locally and we may be able to go with:
```js
disableSetCookie: {
    type: "boolean",
    queryParam: true,
    required: false,
    value: getDisableSetCookie
},
```

package.json
```diff
@@ -109,7 +109,7 @@
     "@krakenjs/zoid": "^10.0.0",
     "@paypal/common-components": "^1.0.35",
     "@paypal/funding-components": "^1.0.31",
-    "@paypal/sdk-client": "^4.0.166",
+    "@paypal/sdk-client": "^4.0.173",
```
gregjopa, line 112:
```suggestion
    "@paypal/sdk-client": "^4.0.174",
```

src/zoid/buttons/component.jsx
wsbrunson: oh i just assumed `getDisableSetCookie` returned true or false. If that's not the case, we could make sure it does.

wsbrunson (approved):

gregjopa (approved):

PR #2189, opened 2023-06-15
fix(venmo): web config

mnicpt:
### Description
https://paypal.atlassian.net/browse/DTCHKINT-1534

❤️ Thank you!


src/zoid/venmo/component.jsx
```diff
@@ -161,6 +166,19 @@ export function getVenmoCheckoutComponent(): VenmoCheckoutComponent {
           value: getLocale,
         },
 
+        channel: {
+          type: "string",
+          queryParam: true,
+          required: false,
+          value: () => (isDevice ? CHANNEL.MOBILE : CHANNEL.DESKTOP),
```
gregjopa, line 173: Is `isDevice` a function? If so, we need to execute it.
```suggestion
          value: () => (isDevice() ? CHANNEL.MOBILE : CHANNEL.DESKTOP),
```

mnicpt:
@gregjopa Thanks. Good catch

PR #2255, opened 2023-10-03
Fix Percy Screenshot CI Failure

jshawl:
### Description

This PR fixes two issues with `npm run percy-screenshot`:

1. `Error: browserType.launch: Executable doesn't exist at /home/runner/.cache/ms-playwright/chromium-1080/chrome-linux/chrome` [link](https://github.com/paypal/paypal-checkout-components/actions/runs/6397759802/job/17366408426#step:11:25)
2. `[percy] Heads up! The current version of @percy/cli is more than 10 releases behind! 1.18.0 -> 1.27.2` [link](https://github.com/paypal/paypal-checkout-components/actions/runs/6397635805/job/17365996503#step:11:16)

While working on an unrelated PR, I noticed the percy screenshot was failing. I pushed up an empty commit to this branch to verify the failure was on the `main` branch and not related to my changes. The error message told me to run `npx playwright install` <- I tried that, and it worked! I also tried finding a related change in either Percy or Playwright that would explain why this test started failing but was unable to locate mention of this error after 2022 🤔 

### Screenshots

Screenshot 2023-10-03 at 15 20 43

❤️ Thank you!


package.json
```diff
@@ -30,7 +30,7 @@
     "start": "npm run webpack -- --progress --watch",
     "test": "npm run format:check && npm run test:unit && npm run jest-ssr && npm run karma && npm run jest-screenshot",
     "test:unit": "vitest run",
-    "percy-screenshot": "babel-node ./test/percy/server/createButtonConfigs.js && percy exec -- playwright test --config=./test/percy/playwright.config.js --reporter=dot --pass-with-no-tests",
+    "percy-screenshot": "npx playwright install && babel-node ./test/percy/server/createButtonConfigs.js && percy exec -- playwright test --config=./test/percy/playwright.config.js --reporter=dot --pass-with-no-tests",
```
wsbrunson, line 33: do we still need this if we updated percy and playwright in the package.json?

jshawl: yes! I only know this because the second commit in this PR upgraded the dependencies and still failed: https://github.com/paypal/paypal-checkout-components/actions/runs/6397759802/job/17366408426 but the third commit added this line and passed 🥳 

Another option would be to back out the dependency upgrades and only keep `npx playwright install`.

PR #2262, opened 2023-10-27
Add a new sign_out_user prop to zoid checkout component

jshawl:
### Description

This PR adds a new prop `sign_out_user` that will be used in smart-payment-buttons to log out the buyer when they click on "Pay with a different account".

Clicking on "Pay with a different account" does not allow the buyer to change their account. This PR is a prerequisite for some upcoming SPB changes that will add a new query parameter `sign_out_user=true` when opening the popup, which will log out the buyer and allow them to log in with another PayPal account.

See the [sneak peek here](https://github.com/paypal/paypal-smart-payment-buttons/pull/705).

Vault the PayPal button. Click on the menu, and click "Pay with a different account". There is no option to change the account. 

Screenshot 2023-10-27 at 15 17 53

❤️ Thank you!

### DIFF
--- a/src/zoid/checkout/component.jsx
+++ b/src/zoid/checkout/component.jsx
@@ -288,6 +288,12 @@ export function getCheckoutComponent(): CheckoutComponent {
           required: false,
           queryParam: true,
         },
+
+        sign_out_user: {
+          type: "boolean",
+          queryParam: true,
+          required: false,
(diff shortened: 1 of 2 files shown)


nbierdeman (approved):
🚀

ravishekhar (approved):

PR #2272, opened 2023-11-03
feat: add support for PPCP AXO Loader

cgdibble:
### Description
Support loading AXO from PPCP

❤️ Thank you!


src/zoid/axo/component.jsx
```diff
@@ -0,0 +1,20 @@
...
+// eslint-disable-next-line flowtype/no-weak-types
+export type AXOComponent = any;
+
+export function getAxoComponent(): AXOComponent {
+  const cmid = getClientMetadataID();
+  const clientID = getClientID();
+  // this will change to whatever options we received
+  // from the merchant
+  try {
+    return loadConnectScript({ cmid, clientID });
```
cgdibble, line 16: We'd need to also retrieve the user Id token using the getter defined in sdk-client and pass it in. 

We want to `await` the script loading so we know it's done, then a window object is available for us to initialize. We then pass the appropriate params.
```suggestion
    await  loadConnectScript();
    window.braintree.connect.create({ cmid, clientID, userIdToken })
```

src/connect/component.jsx
```diff
@@ -0,0 +1,42 @@
...
+// -> Looks like it returns a function that accepts the props
+// How do we define the input of merchant params here?
+export const getConnectComponent = async (merchantProps) => {
+  const cmid = getClientMetadataID();
+  const clientID = getClientID();
+  const userIdToken = getUserIDToken();
+  // TODO: Sort out integration specifics for inputs
+  try {
+    const loadResult = await loadAxo({
+      client: { getVersion: () => "3.97.3-connect-alpha.6.1" }, // this is currently supported for BT
```
sunnypatel, line 20: Adding `sdkVersion` support in loader so can remove `client` from here

src/connect/component.jsx
```diff
@@ -0,0 +1,42 @@
...
+      sdkVersion: "", // PPCP proposal: string instead of function
+      minified: false,
```
sunnypatel, line 22:
```suggestion
      platform: "PPCP",
      sdkVersion: "3.97.3-connect-alpha.6.1",
      minified: false,
```

sunnypatel: @siddy2181 I updated the `sdkVersion` to -> `btSdkVersion` to be more specific since that's what we're attempting to load under the hood

src/connect/component.jsx
sunnypatel, line 20: @cgdibble @siddy2181 thoughts on making this a property the Merchant can pass in? 

Reasoning is for Partners who might have complex integrations where their loading Axo on PPCP but also loading other BT modules.  We should allow them to configure which BT SDK version to load Axo assuming it's at least the minimum version with the Connect module.

cgdibble:
Closing as we needed the feature prefix for CI and went straight to `main`: https://github.com/paypal/paypal-checkout-components/pull/2280

PR #2276, opened 2023-11-16
Hosted buttons component

jshawl:
### Description

This PR implements a new `paypal.HostedButtons()` component. Example usage:

```html
<div id="paypal-hosted-buttons"></div>
<script src="https://www.paypal.com/sdk/js?client-id=test&currency=USD&components=hosted-buttons"></script>
<script>
  paypal.HostedButtons({
    hostedButtonId: "B1234567890"
  }).render("#paypal-hosted-buttons")
</script>
```

This new component will allow merchants to configure styles, prices, and quantities through a hosted UI. These changes then fetch the details of that hosted button and render them using v5 of the JS SDK.

### Reproduction Steps

N/A - the Hosted Buttons UI is not publicly available 

❤️ Thank you!

### DIFF
--- a/src/hosted-buttons/index.js
+++ b/src/hosted-buttons/index.js
@@ -1,7 +1,5 @@
 /* @flow */
 
-import { getMerchantID } from "@paypal/sdk-client/src";
-
 import { getButtonsComponent } from "../zoid/buttons";
 
 import {
@@ -9,6 +7,7 @@ import {
   buildHostedButtonOnApprove,
   getHostedButtonDetails,
   renderForm,
+  getMerchantID,
 } from "./utils";
 import type {
   HostedButtonsComponent,
@@ -22,10 +21,7 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
   }: HostedButtonsComponentProps): HostedButtonsInstance {
     const Buttons = getButtonsComponent();
     const render = (selector) => {
-      // The SDK supports mutiple merchant IDs, but hosted buttons only
-      // have one merchant id as a query parameter to the SDK script.
-      // https://github.com/paypal/paypal-sdk-client/blob/c58e35f8f7adbab76523eb25b9c10543449d2d29/src/script.js#L144
-      const merchantId = getMerchantID()[0];
+      const merchantId = getMerchantID();
 
       getHostedButtonDetails({ hostedButtonId }).then(
         ({ html, htmlScript, style }) => {
--- a/src/hosted-buttons/index.js
+++ b/src/hosted-buttons/index.js
@@ -29,28 +29,19 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
 
       getHostedButtonDetails({ hostedButtonId }).then(
         ({ html, htmlScript, style }) => {
-          renderForm({ html, htmlScript, selector });
+          const { onInit, onClick } = renderForm({
+            hostedButtonId,
+            html,
+            htmlScript,
+            selector,
+          });
 
           // $FlowFixMe
           Buttons({
-            style,
             hostedButtonId,
-            onInit(data, actions) {
-              // disable the button, listen for input changes,
-              // and enable the button when the form is valid
-              // using actions.disable() and actions.enable()
-              window[`__pp_form_fields_${hostedButtonId}`]?.onInit?.(
-                data,
-                actions
-              );
-            },
-            onClick(data, actions) {
-              // render form errors, if present
-              window[`__pp_form_fields_${hostedButtonId}`]?.onClick?.(
-                data,
-                actions
-              );
-            },
+            style,
+            onInit,
+            onClick,
             createOrder: buildHostedButtonCreateOrder({
               hostedButtonId,
               merchantId,
--- a/src/hosted-buttons/index.js
+++ b/src/hosted-buttons/index.js
@@ -1,0 +1,71 @@
+/* @flow */
+
+import { getMerchantID } from "@paypal/sdk-client/src";
+
+import { getButtonsComponent } from "../zoid/buttons";
+
+import {
+  buildHostedButtonCreateOrder,
+  buildHostedButtonOnApprove,
+  getHostedButtonDetails,
+  renderForm,
+} from "./utils";
+import type {
+  HostedButtonsComponent,
+  HostedButtonsComponentProps,
+  HostedButtonsInstance,
+} from "./types";
+
+export const getHostedButtonsComponent = (): HostedButtonsComponent => {
+  function HostedButtons({
+    hostedButtonId,
+  }: HostedButtonsComponentProps): HostedButtonsInstance {
+    const Buttons = getButtonsComponent();
+    const render = (selector) => {
+      // The SDK supports mutiple merchant IDs, but hosted buttons only
+      // have one merchant id as a query parameter to the SDK script.
+      // https://github.com/paypal/paypal-sdk-client/blob/c58e35f8f7adbab76523eb25b9c10543449d2d29/src/script.js#L144
+      const merchantId = getMerchantID()[0];
+
+      getHostedButtonDetails({ hostedButtonId }).then(
+        ({ html, htmlScript, style }) => {
+          renderForm({ html, htmlScript, selector });
+
+          // $FlowFixMe
+          Buttons({
+            style,
+            hostedButtonId,
+            onInit(data, actions) {
+              // disable the button, listen for input changes,
+              // and enable the button when the form is valid
+              // using actions.disable() and actions.enable()
+              window[`__pp_form_fields_${hostedButtonId}`]?.onInit?.(
+                data,
+                actions
+              );
+            },
+            onClick(data, actions) {
+              // render form errors, if present
+              window[`__pp_form_fields_${hostedButtonId}`]?.onClick?.(
+                data,
+                actions
+              );
+            },
+            createOrder: buildHostedButtonCreateOrder({
+              hostedButtonId,
+              merchantId,
+            }),
+            onApprove: buildHostedButtonOnApprove({
+              hostedButtonId,
+              merchantId,
+            }),
+          }).render(selector);
+        }
+      );
+    };
+    return {
+      render,
+    };
+  }
+  return HostedButtons;
+};
--- a/src/hosted-buttons/utils.js
+++ b/src/hosted-buttons/utils.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { request, memoize, popup } from "@krakenjs/belter/src";
+import { request, memoize, popup, supportsPopups } from "@krakenjs/belter/src";
 import {
   getSDKHost,
   getClientID,
@@ -163,12 +163,16 @@ export const buildHostedButtonOnApprove = ({
         // The "Debit or Credit Card" button does not open a popup
         // so we need to open a new popup for buyers who complete
         // a checkout via "Debit or Credit Card".
-        if (data.paymentSource === "card") {
+        if (data.paymentSource === FUNDING.CREDIT) {
           const url = `${baseUrl}/ncp/payment/${hostedButtonId}/${data.orderID}`;
-          popup(url, {
-            width: DEFAULT_POPUP_SIZE.WIDTH,
-            height: DEFAULT_POPUP_SIZE.HEIGHT,
-          });
+          if (supportsPopups()) {
+            popup(url, {
+              width: DEFAULT_POPUP_SIZE.WIDTH,
+              height: DEFAULT_POPUP_SIZE.HEIGHT,
+            });
+          } else {
+            window.location = url;
+          }
         }
         return response;
       });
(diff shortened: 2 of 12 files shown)


nbierdeman (approved):
Niiiiiice! 💯

src/hosted-buttons/index.js
```diff
@@ -0,0 +1,209 @@
...
+              // disable the button, listen for input changes,
+              // and enable the button when the form is valid
+              // using actions.disable() and actions.enable()
+              window.__pp_form_fields?.onInit?.(data, actions);
+            },
+            onClick(data, actions) {
+              // render form errors, if present
+              window.__pp_form_fields?.onClick?.(data, actions);
+            },
+            createOrder: getHostedButtonCreateOrder({
```
jshawl, line 192: `buildCreateOrder` ?

src/hosted-buttons/index.js
```diff
@@ -0,0 +1,209 @@
...
+    url: `${baseUrl}/ncp/api/form-fields/${hostedButtonId}`,
+    headers: getHeaders(),
+  }).then(({ body }) => {
+    const variables = body.button_details.link_variables;
+    return {
+      style: {
+        layout: getButtonVariable(variables, "layout"),
+        shape: getButtonVariable(variables, "shape"),
+        color: getButtonVariable(variables, "color"),
+        label: getButtonVariable(variables, "button_text"),
```
jshawl, line 79: can the payload be formatted for us ahead of time?

src/hosted-buttons/index.js
```diff
@@ -0,0 +1,209 @@
...
+
+          // $FlowFixMe
+          Buttons({
+            style,
+            hostedButtonId,
+            onInit(data, actions) {
+              // disable the button, listen for input changes,
+              // and enable the button when the form is valid
+              // using actions.disable() and actions.enable()
+              window.__pp_form_fields?.onInit?.(data, actions);
```
jshawl, line 186: is there a visual cue that the button is disabled? aside from form errors?

src/hosted-buttons/index.js
```diff
@@ -0,0 +1,209 @@
...
+const getHeaders = (accessToken?: string) => ({
+  ...(accessToken && { Authorization: `Bearer ${accessToken}` }),
+  "Content-Type": "application/json",
+  "PayPal-Entry-Point": entryPoint,
+});
+
+const getButtonVariable = (variables: ButtonVariables, key: string): string =>
+  variables?.find((variable) => variable.name === key)?.value ?? "";
+
+const getFundingSource = (paymentSource) => {
```
jshawl, line 59: why are we reformatting the paymentsource?

src/hosted-buttons/index.js
```diff
@@ -0,0 +1,209 @@
...
+const createAccessToken = memoize<CreateAccessToken>((clientId) => {
+  return request({
+    url: `${baseUrl.replace("www", "api")}/v1/oauth2/token`,
+    method: "POST",
+    body: "grant_type=client_credentials",
+    headers: {
+      Authorization: `Basic ${btoa(clientId)}`,
+      "Content-Type": "application/json",
+    },
+  }).then((response) => response.body.access_token);
```
jshawl, line 96: what if this fails? what other errors should we handle?

nikrom17 (approved):

src/interface/hosted-buttons.js
```diff
@@ -0,0 +1,29 @@
...
+};
+
+export const CardForm: LazyProtectedExport<CardFormComponent> = {
+  __get__: () => protectedExport(getCardFormComponent()),
+};
+
+export function setup() {
+  getButtonsComponent();
+  getCheckoutComponent();
+}
```
wsbrunson, line 29: What does `setup` do?

src/zoid/checkout/component.jsx
```diff
@@ -317,6 +317,11 @@ export function getCheckoutComponent(): CheckoutComponent {
           queryParam: true,
           required: false,
         },
+
+        hostedButtonId: {
+          type: "string",
+          required: false,
+        },
```
wsbrunson, line 324: does the checkout ui need this?

src/hosted-buttons/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+          // $FlowFixMe
+          Buttons({
+            style,
+            hostedButtonId,
+            onInit(data, actions) {
+              // disable the button, listen for input changes,
+              // and enable the button when the form is valid
+              // using actions.disable() and actions.enable()
+              window[`__pp_form_fields_${hostedButtonId}`]?.onInit?.(
```
wsbrunson, line 42: could we return callbacks from `renderForm` and use them here instead of trying to find it on the window? does that make sense?

```
const formComponent = renderForm(...)
formComponent.onInit
```

src/interface/hosted-buttons.js
jshawl: I'm not sure other than without it, `window.xprops` is undefined - and this line throws: https://github.com/paypal/paypal-smart-payment-buttons/blob/c268be3d0b18ff0ecaa11df646e74c8787ec03ff/src/button/button.js#L130

will dm an internal link to where `import {setup}` is being called

jshawl: oh I can say this generally:

in csnw: `/server/sdk/bootstrap` line # 35

src/zoid/checkout/component.jsx
jshawl: yes but only for the pre-render button click https://github.com/paypal/paypal-checkout-components/pull/2276/files#diff-f15dc9fa288e403c9d09722ceae61adb63327f473aa7807fbf3cb49896732f7cR81-R86

src/interface/hosted-buttons.js
wsbrunson: oh we do have this in `interface/buttons.js` this is fine

src/zoid/checkout/component.jsx
wsbrunson: what is checkout doing with it?

jshawl: SPB destructures the `hostedButtonId` from `prerenderDetails` here - https://github.com/paypal/paypal-smart-payment-buttons/blob/c268be3d0b18ff0ecaa11df646e74c8787ec03ff/src/button/button.js#L393-L397

and then adds it into `payment` here - https://github.com/paypal/paypal-smart-payment-buttons/blob/c268be3d0b18ff0ecaa11df646e74c8787ec03ff/src/button/button.js#L413-L419

then initiate payment is called right below - https://github.com/paypal/paypal-smart-payment-buttons/blob/c268be3d0b18ff0ecaa11df646e74c8787ec03ff/src/button/button.js#L420

and then passed in here when `Checkout({` is invoked - https://github.com/paypal/paypal-smart-payment-buttons/blob/c268be3d0b18ff0ecaa11df646e74c8787ec03ff/src/payment-flows/checkout.js#L330

this took me a significant amount of time to figure out so I'm super open to removing it (because of the increased complexity) and/or identifying another solution - wdyt?

nhanthanhnguyen95 (approved):

src/hosted-buttons/index.js
jshawl: oo I like this idea - a254ae90

but I think they have to be called on `window` (or some other global on `window`) because the method is loaded dynamically after fetching the hosted button details.

src/hosted-buttons/index.js
```diff
@@ -0,0 +1,62 @@
...
+export const getHostedButtonsComponent = (): HostedButtonsComponent => {
+  function HostedButtons({
+    hostedButtonId,
+  }: HostedButtonsComponentProps): HostedButtonsInstance {
+    const Buttons = getButtonsComponent();
+    const render = (selector) => {
+      // The SDK supports mutiple merchant IDs, but hosted buttons only
+      // have one merchant id as a query parameter to the SDK script.
+      // https://github.com/paypal/paypal-sdk-client/blob/c58e35f8f7adbab76523eb25b9c10543449d2d29/src/script.js#L144
+      const merchantId = getMerchantID()[0];
```
ravishekhar, line 28: should we throw an error here if multiple merchant id's are provided instead of silently ignoring it

src/hosted-buttons/utils.js
```diff
@@ -0,0 +1,157 @@
...
+          merchant_id: merchantId,
+          context_id: data.orderID,
+        }),
+      }).then((response) => {
+        // The "Debit or Credit Card" button does not open a popup
+        // so we need to open a new popup for buyers who complete
+        // a checkout via "Debit or Credit Card".
+        if (data.paymentSource === "card") {
+          const url = `${baseUrl}/ncp/payment/${hostedButtonId}/${data.orderID}`;
+          popup(url, {
```
ravishekhar, line 148: If running in webview or non popup supported devices this logic would error out. Can smart buttons disable card-form payment flow instead when hosted button id is passed?

ravishekhar (approved):
Looks good. Left a few suggestions.

src/hosted-buttons/utils.js
```diff
@@ -0,0 +1,157 @@
...
+      },
+    }).then((response) => response.body.access_token);
+  }
+);
+
+const getButtonVariable = (variables: ButtonVariables, key: string): string =>
+  variables?.find((variable) => variable.name === key)?.value ?? "";
+
+const getFundingSource = (paymentSource) => {
+  if (paymentSource === "credit") {
```
nbierdeman, line 46: Maybe we could import these values from `FUNDING` in `@paypal/sdk-constants/src` instead of hardcoding `"credit"` and `'CARD'`?

src/hosted-buttons/utils.js
```diff
@@ -0,0 +1,157 @@
...
+  html,
+  htmlScript,
+  selector,
+}) => {
+  const elm =
+    typeof selector === "string" ? document.querySelector(selector) : selector;
+  if (elm) {
+    elm.innerHTML = html + htmlScript;
+    const newScriptEl = document.createElement("script");
+    const oldScriptEl = elm.querySelector("script");
```
nbierdeman, line 88: Will there only ever be one script per element? I'm assuming so, but thought I'd doublecheck.

src/hosted-buttons/utils.js
```diff
@@ -0,0 +1,157 @@
...
+    return createAccessToken(getClientID()).then((accessToken) => {
+      return request({
+        url: `${apiUrl}/v1/checkout/links/${hostedButtonId}/create-context`,
+        headers: getHeaders(accessToken),
+        method: "POST",
+        body: JSON.stringify({
+          entry_point: entryPoint,
+          funding_source: getFundingSource(data.paymentSource),
+          merchant_id: merchantId,
+          ...userInputs,
```
nbierdeman, line 118: Is this a possible vector for malicious JS that needs to be guarded against?

nbierdeman (commented):
Nice work, @jshawl! It's looking good! Just a couple minor questions/comments for you.

src/hosted-buttons/utils.js
jshawl: ah good catch - another option could be redirecting the `window` to this location if `!supportsPopups()` from https://github.com/krakenjs/belter/blob/d0c0868c996aea92d1ad9a6c8662da7a2c200b5b/src/device.js#L229 wdyt?

src/hosted-buttons/index.js
jshawl: yes! check it out 4e1cef71

src/hosted-buttons/utils.js
jshawl: yes! I like it 72e75b05

src/hosted-buttons/index.js
wsbrunson: we chatted offline about this, going to keep this as is

src/zoid/checkout/component.jsx
wsbrunson: we chatted offline about this, checkout is not actually using this value, going to remove it from zoid

src/hosted-buttons/utils.js
jshawl: that is right! this is only because it is sent as one property in the response here - https://github.com/paypal/paypal-checkout-components/pull/2276/files#diff-dc0ab5efddb0f4c5dc8c699ccafce1daa87906e1b839480c5be57048cee5482bR88

src/hosted-buttons/utils.js
jshawl: I'm having trouble thinking of an attack vector here but only because the user input is never put back into the DOM; it will be sent to the hosted buttons API though. I think we're good here but lmk if you had a sanitization strategy in mind and/or if you have an idea for how to exploit.

wsbrunson (approved):
ship it

src/hosted-buttons/utils.js
ravishekhar: This is also a good option, but effectively this would be duplicate of the logic SPB already has 
 https://github.com/paypal/paypal-smart-payment-buttons/blob/c268be3d0b18ff0ecaa11df646e74c8787ec03ff/src/payment-flows/checkout.js#L653

Also browsers may prevent from opening a popup here, as its not happening inside of a synchronous button onClick event.

ravishekhar: We chatted offline about this. Keeping it as is in the short term.

ravishekhar (approved):
🚀

PR #2280, opened 2023-11-28
feat(connect): add initial support for connect component

cgdibble:
### Description

The initial component for Connect to be available on the PPCP SDK. Still in development and gated at the Connect Portal layer. 

Ticket DTPPCPSDK-1367 for a variety of work. 

❤️ Thank you!


src/connect/component.jsx
```diff
@@ -0,0 +1,93 @@
...
+      dimensions: {
+        errorName: "connect_load_error",
+      },
+    });
+
+    throw new Error(error);
+  }
+
+  try {
+    const connect = await window.braintree.connect.create({
```
jshawl, line 65: I remember one issue where I was unable to rely on `window.braintree` because the npm module (if npm-installed) did not attach itself to the window object - just mentioning it in case the integration pattern should support `npm install braintree-web`

src/connect/component.jsx
```diff
@@ -0,0 +1,93 @@
+/* @flow */
+import { loadAxo } from "@paypal/connect-loader-component";
+import {
+  getClientID,
+  getClientMetadataID,
+  getUserIDToken,
+  getLogger,
+} from "@paypal/sdk-client/src";
+
+const sendCountMetric = ({
```
jshawl, line 10: I'm totally going to copy this for a different component! <- would it make sense to put `sendCountMetric` in a shared util file somewhere?

src/connect/component.jsx
```diff
@@ -0,0 +1,93 @@
...
+  try {
+    loadResult = await loadAxo({
+      platform: "PPCP",
+      btSdkVersion: "3.97.3-connect-alpha.6.1",
+      minified: false,
+      metadata,
+    });
+  } catch (error) {
+    sendCountMetric({
+      name: "pp.app.paypal_sdk.connect.init.error.count",
```
jshawl, line 54: should this be `pp.app.paypal_sdk.connect.load.error.count` ?

src/connect/component.jsx
```diff
@@ -0,0 +1,93 @@
...
+      platformOptions: {
+        platform: "PPCP",
+        userIdToken,
+        clientID,
+        clientMetadataID: cmid,
+      },
+    });
+
+    sendCountMetric({
+      name: "pp.app.paypal_sdk.connect.init.success.count",
```
jshawl, line 77: did you follow a guideline for this `name`? (I'm about to do the same with another component but want to standardize with what you've done here!)

jshawl (commented):
😍 this has given me a ton of inspiration for the HostedButtons component! Nice job, y'all!

src/connect/component.jsx
cgdibble: @jshawl we referenced the `card-fields` stuff we'd done before but otherwise no we didn't know of a guide. Does on exist?

src/connect/component.jsx
cgdibble: @jshawl yes! @siddy2181 is actually working on that and moving it to `sdk-client`. It's used in SPB a bunch, and we want it here. We copied it over as we navigate the timeline dance, but the intent is to use a shared util of `sendCountMetric` from `sdk-client`.

src/connect/component.jsx
```diff
@@ -0,0 +1,93 @@
...
+  const cmid = getClientMetadataID();
+  const clientID = getClientID();
+  const userIdToken = getUserIDToken();
+  const { metadata } = merchantProps;
+
+  let loadResult = {};
+  try {
+    loadResult = await loadAxo({
+      platform: "PPCP",
+      btSdkVersion: "3.97.3-connect-alpha.6.1",
```
cgdibble, line 48: So far this has to stay hardcoded. More to come, though.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paypal/checkout-components",
-  "version": "5.0.286",
+  "version": "5.0.288",
```
elizabethmv, line 3: Was this change meant to be included?
```suggestion
  "version": "5.0.286",
```

cgdibble: hmm, no, good catch let me fix that. Must be an artifact of feature-branch alpha versions?

src/connect/component.jsx
```diff
@@ -0,0 +1,93 @@
...
+    return connect;
+  } catch (error) {
+    sendCountMetric({
+      name: "pp.app.paypal_sdk.connect.init.error.count",
+      event: "error",
+      dimensions: {
+        errorName: "connect_init_error",
+      },
+    });
+    throw new Error(error);
```
wsbrunson, line 91: we should also log these errors `log.error()` so that we can investigate them in the future. I don't think we need logs for anything else, the metrics will be enough for the init and success cases. We'd want to see the errors themselves in cal though if theres an issue

wsbrunson (approved):

src/connect/component.jsx
siddy2181: We are keeping all errors under the same event name. Which then will be funneled based on `load`vs `create` based on the event dimension param

dist/button.js
cgdibble, line 1: Ugh whyyy. I'm going to remove these, I'm not ok with incorporating other changes like this for this scenario.

src/connect/component.jsx
```diff
@@ -0,0 +1,80 @@
...
+  const clientID = getClientID();
+  const userIdToken = getUserIDToken();
+  const { metadata } = merchantProps;
+
+  let loadResult = {};
+  try {
+    loadResult = await loadAxo({
+      platform: "PPCP",
+      btSdkVersion: "3.97.3-connect-alpha.6.1",
+      minified: false,
```
westeezy, line 31: Should this be true for when we merge it? Assuming minified was for testing locally

test/integration/tests/button/multiple.js
```diff
@@ -30,32 +30,32 @@ for (const flow of ["popup", "iframe"]) {
     });
 
     const cases = [
-      {
```
westeezy, line 33: Why were these commented out?

src/connect/component.jsx
cgdibble: @wsbrunson done!

src/connect/component.jsx
cgdibble: Good eye. We have a ticket to map the `debug` query param to this, but you're right that we should probably set it to `true`

nbierdeman (approved):
🚀

test/integration/tests/button/multiple.js
cgdibble: They shouldn't be. We ran the full test suite locally and this test was failing and we accidentally committed part of our debug process. TLDR it's test flake locally. They all pass in CI.

PR #2281, opened 2023-11-29
fix: bump connect loader to beta without polyfill

cgdibble:
### Description

Use the new Connect Loader beta version that removes the polyfill from the downstream asset loader.

Our build enforces flat deps and the polyfill is a dep required by old Hosted Fields.

❤️ Thank you!

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "@krakenjs/zoid": "^10.3.1",
     "@paypal/common-components": "^1.0.35",
     "@paypal/funding-components": "^1.0.31",
-    "@paypal/connect-loader-component": "^1.1.0",
+    "@paypal/connect-loader-component": "1.1.0-beta.1",
     "@paypal/sdk-client": "^4.0.176",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -110,7 +110,7 @@
     "@krakenjs/zoid": "^10.3.1",
     "@paypal/common-components": "^1.0.35",
     "@paypal/funding-components": "^1.0.31",
-    "@paypal/connect-loader-component": "^1.1.0",
+    "@paypal/connect-loader-component": "1.1.0-beta.1",
```
siddy2181, line 113: We need to update `package_lock.json` too?

SancSalix (approved):
🌟

package.json
cgdibble: @siddy2181 there isn't a lock file in checkout components.
We disable the lockfile generation here: https://github.com/paypal/paypal-checkout-components/blob/main/.npmrc#L3

siddy2181 (approved):
Let's ship it!!!!!

PR #2286, opened 2023-12-05
Add ShopperInsights component

elizabethmv:
This PR adds a new Shopper Insights component which exposes a `getRecommendedPaymentMethods({})` helper intended for combined use with the SPBs to present PayPal or Venmo upstream. It is also initially a protected export accessible on the PayPal domain until we open up the feature.

Right now we are using existing kraken libraries to support the requests (Belter, ZalgoPromise) but we will be creating a separate PR to use the fetch API. We wanted to break this out so that we can get more dedicated feedback on that separately.


src/constants/api.js
```diff
@@ -0,0 +1,35 @@
...
+export const HEADERS = {
+  AUTHORIZATION: "authorization",
+  CONTENT_TYPE: "content-type",
+  PARTNER_ATTRIBUTION_ID: "paypal-partner-attribution-id",
+};
+
+export const ELIGIBLE_PAYMENT_METHODS = "v2/payments/find-eligible-methods";
+
+// $FlowFixMe
+export const BASE_API_URL = {
```
wsbrunson, line 13: do we get these from csnw? https://github.com/paypal/paypal-sdk-client/blob/main/src/global.js#L46

that might be how we get the stage urls

src/lib/security.js
```diff
@@ -23,6 +23,15 @@ export function allowIframe(): boolean {
   return false;
 }
 
+// TODO: Incorporate an allow list for alpha merchants
+const isMerchantDomainAllowed = false;
```
wsbrunson, line 27: i think we are dropping this in favor of an allow list on the SI api

src/api/shopper-insights/validation.js
```diff
@@ -0,0 +1,198 @@
...
+    purchase_units: [
+      {
+        amount: {
+          currency_code: hasPurchaseUnits(merchantPayload)
+            ? // $FlowFixMe
+              merchantPayload?.purchaseUnits?.[0]?.amount?.currencyCode
+            : getCurrency(),
+        },
+      },
+    ],
```
wsbrunson, line 189: we can do a new build tomorrow but I think we should do:

```
purchase_units: hasPurchaseUnits(merchantPayload)
  ? merchantPayload.purchaseUnits
  : [...our purchase unit with the currency]
```

that way if for whatever reason they want to pass their whole purchase unit to the api they can and we'll stay out of the way

src/api/utils.js
```diff
@@ -0,0 +1,70 @@
...
+    headers: requestHeaders,
+    json: data,
+  }).then(({ status, body, headers: responseHeaders }) => {
+    if (status >= 300) {
+      const error = new Error(
+        `${url} returned status ${status}\n\n${JSON.stringify(body)}`
+      );
+
+      // $FlowFixMe
+      error.response = { status, headers: responseHeaders, body };
```
wsbrunson, line 63: error.status = status

wsbrunson (commented):
lookin' pretty dang slick

ravishekhar (approved):
💯 awesome work!!

PR #2287, opened 2023-12-05
feat: provide fraudnet preload functionality and wire up connect with it

cgdibble:

As the latest fraudnet option comes to sdk-client (https://github.com/paypal/paypal-sdk-client/pull/170) we want to use it and enable the Connet component to call it at the necessary point.

❤️ Thank you!


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paypal/checkout-components",
-  "version": "5.0.289",
+  "version": "5.0.290",
```
cgdibble, line 3: TODO: This must be removed before we merge. We have to increment for NPM alpha version testing

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paypal/checkout-components",
-  "version": "5.0.289",
```
cgdibble, line 3: TODO: This needs to be restored before we can merge

package.json
```diff
@@ -1,6 +1,6 @@
...
+  "version": "5.0.291",
```
cgdibble, line 3:
```suggestion
  "version": "5.0.289",
```

PR #2290, opened 2023-12-08
Update npm registry to npm.dev.paypalinc.com

sebastianfdz:
### Description

Update .npmrc file to include latest registry url.

❤️ Thank you!

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,4 +1,4 @@
-registry=https://registry.npmjs.org
+registry=https://npm.dev.paypalinc.com
 message="chore(release): %s 🎉"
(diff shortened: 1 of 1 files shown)


.npmrc
```diff
@@ -1,4 +1,4 @@
-registry=https://registry.npmjs.org
+registry=https://npm.dev.paypalinc.com
```
jshawl, line 1: I think this has to be the public npm registry because GitHub actions will fail to download dependencies from a non-public registry.

sebastianfdz: Got it! Thanks Jesse.

PR #2296, opened 2023-12-28
set referrerDomain in xprops

jshawl:
### Description

This PR adds a new prop to the `Buttons` component, which will be used to identify the referrer to a page that loaded the SDK.

The analytics team wants to be able to identify major channels (e.g. facebook / instagram) to SDK uses.

```
npm run karma
```

❤️ Thank you!

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -721,7 +721,13 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
 
       referrer_domain: {
         type: "string",
-        value: () => new URL(window.document.referrer).host,
+        value: () => {
+          const referrer = window.document.referrer;
+          const matches = referrer.match(/:\/\/([^\/\?]+)/);
+          if (matches) {
+            return matches[1];
+          }
+        },
       },
 
       userIDToken: {
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -719,16 +719,11 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
         },
       },
 
-      referrer_domain: {
+      referrerDomain: {
         type: "string",
         required: false,
-        value: () => {
-          const referrer = window.document.referrer;
-          const matches = referrer.match(/:\/\/([^/?]+)/);
-          if (matches) {
-            return matches[1];
-          }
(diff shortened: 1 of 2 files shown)


src/zoid/buttons/component.jsx
```diff
@@ -719,6 +719,18 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+      referrer_domain: {
+        type: "string",
+        required: false,
+        value: () => {
+          const referrer = window.document.referrer;
+          const matches = referrer.match(/:\/\/([^/?]+)/);
+          if (matches) {
+            return matches[1];
+          }
+        },
```
jshawl, line 731: A few notes about this implementation:

1. using `URL()` failed the eslint check because ie9 does not support `URL`
2. There are more comprehensive regexes we could use (https://stackoverflow.com/a/26987741) but I thought the simpler regex would be more readable and we actually don't _need_ a valid domain, just an identifier that is domain-like.
3. I thought about moving this to a shared util location but:
  - [belter](https://github.com/search?q=repo%3Akrakenjs%2Fbelter%20domain&type=code) doesn't have any domain-related utils
  - [cross-domain-utils](https://github.com/krakenjs/cross-domain-utils/blob/main/src/utils.js#L91) is coupled to `window.location` (not `document.referrer`)
  - [paypal-checkout-components](https://github.com/paypal/paypal-checkout-components/tree/main/src/lib) does not have any related URL parsing

I'm definitely open to moving it! wdyt?

src/zoid/buttons/component.jsx
```diff
@@ -719,6 +719,18 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
         },
       },
 
+      referrer_domain: {
+        type: "string",
+        required: false,
+        value: () => {
+          const referrer = window.document.referrer;
+          const matches = referrer.match(/:\/\/([^/?]+)/);
```
jshawl, line 727: this will match anything between a `://` and a `/`

src/zoid/buttons/component.jsx
wsbrunson: We don't support IE11 and earlier, you should always feel empowered to us a feature as long as its supported in most modern browsers

wsbrunson (approved):
great work!

PR #2302, opened 2024-01-05
Validate braintree version and default cmid to sessionId if null

sebastianfdz:
### Description

This PR contains 2 changes:

1. When we receive a null `cmid` from the sdk `data-client-metadata-id` data attribute, we will pass the sessionId to AXO as it is a required value.
2. We will validate that the version of braintree being used supports AXO. Otherwise, if an unsupported version is passed we will throw an error, and if no version is passed then it means the client is not using braintree and we can safely default it to the minimum supported version for AXO.

[cmid change ticket](https://paypal.atlassian.net/browse/DTPPCPSDK-1403)
[braintree validation ticket](https://paypal.atlassian.net/browse/DTPPCPSDK-1445) 

❤️ Thank you!

### DIFF
--- a/src/connect/component.jsx
+++ b/src/connect/component.jsx
@@ -20,7 +20,7 @@ const MIN_MAJOR_VERSION = 3;
 const MIN_MINOR_VERSION = 97;
 export const MIN_BT_VERSION = `${MIN_MAJOR_VERSION}.${MIN_MINOR_VERSION}.3-connect-alpha.6.1`; // Minimum for supporting AXO
 
-export function getSdkVersion(version: string | null): string | null {
+export function getSdkVersion(version: string | null): string {
   if (!version) {
     return MIN_BT_VERSION;
   }
--- a/src/connect/component.jsx
+++ b/src/connect/component.jsx
@@ -41,7 +41,7 @@ export function getSdkVersion(version: string | null): string {
     });
 
     throw new Error(
-      `The braintree version used does not support Accelerated Checkout (AXO). Please use version ${MIN_BT_VERSION} or above`
+      `The braintree version used does not support Connect. Please use version ${MIN_BT_VERSION} or above`
     );
   }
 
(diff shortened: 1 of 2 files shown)


src/connect/component.jsx
```diff
@@ -10,22 +10,55 @@ import {
...
+    sendCountMetric({
+      name: "pp.app.paypal_sdk.connect.init.error.count",
+      event: "error",
+      dimensions: {
+        errorName: "braintree_version_not_supported_error",
+      },
+    });
+
+    throw new Error(
+      `The braintree version used does not support Accelerated Checkout (AXO). Please use version ${MIN_BT_VERSION} or above`
```
cgdibble, line 44:
```suggestion
      `The braintree version used does not support Connect. Please use version ${MIN_BT_VERSION} or above`
```
`Connect` is the public-facing integration name so let's use that here

cgdibble (approved):

src/connect/component.jsx
sebastianfdz: Good catch!

src/connect/component.jsx
```diff
@@ -10,22 +10,55 @@ import {
...
+    );
+  }
+
+  return version;
+}
+
 // $FlowFixMe
 export const getConnectComponent = async (merchantProps = {}) => {
-  const cmid = getClientMetadataID();
+  const cmid = getClientMetadataID() || getSessionID();
```
sunnypatel, line 53: If I refresh the page does `getSessionID()` return a unique id each time?  What determines a "session"?

sunnypatel: Ideally we don't want a unique id.  So if a Buyer goes between checkout page and other pages it would return the same ID upon returning to the checkout page.  Is that how it would work?

sunnypatel: Gotcha that should be perfect ... sharing w/ our analytics team for visibility as they use this id to identity unique sessions.   Will tag them in our Slack thread [here](https://paypal.slack.com/archives/C04FBV3163E/p1704735353301459?thread_ts=1704383556.640879&cid=C04FBV3163E)

PR #2305, opened 2024-01-08
change jsx extension to js for file with no jsx

wsbrunson:
component.jsx does not have any jsx, it should have a `js` extension


jshawl (approved):
:shipit:

PR #2307, opened 2024-01-09
Memoize Shopper Insights API Call

ravishekhar:
### Description
* Refactor the existing shopper insights tests to mock the request method of external belter module instead of internal callRestAPI
* Memoize the Shopper Insights API Call to avoid duplicate calls with exact same set of parameters. 

❤️ Thank you!

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -1,7 +1,7 @@
 /* @flow */
 
 import { getPartnerAttributionID, getSessionID } from "@paypal/sdk-client/src";
-import { request } from "@krakenjs/belter/src";
+import { inlineMemoize, request } from "@krakenjs/belter/src";
 import { ZalgoPromise } from "@krakenjs/zalgo-promise/src";
 
 import { HEADERS } from "../constants/api";
@@ -51,4 +51,17 @@ export function callRestAPI({
 
     return body;
   });
+}
+
+export function callMemoizedRestAPI({
+  accessToken,
+  method,
+  url,
+  data,
+}: RestAPIParams): ZalgoPromise<Object> {
+  return inlineMemoize(
+    callMemoizedRestAPI,
+    () => callRestAPI({ accessToken, method, url, data }),
+    [accessToken, method, url, JSON.stringify(data)]
+  );
 }
--- a/src/api/shopper-insights/component.test.js
+++ b/src/api/shopper-insights/component.test.js
@@ -73,6 +73,69 @@ describe("shopper insights component - getRecommendedPaymentMethods()", () => {
     expect.assertions(2);
   });
 
+  test("should get recommended payment methods from memoized request for the exact same payload", async () => {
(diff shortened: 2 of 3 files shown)


src/api/api.js
```diff
@@ -1,7 +1,7 @@
 /* @flow */
 
 import { getPartnerAttributionID, getSessionID } from "@paypal/sdk-client/src";
-import { request } from "@krakenjs/belter/src";
+import { inlineMemoize, request } from "@krakenjs/belter/src";
```
jshawl, line 4: what is the difference between [`inlineMemoize`](https://github.com/krakenjs/belter/blob/d0c0868c996aea92d1ad9a6c8662da7a2c200b5b/src/util.js#L334) and [`memoize`](https://github.com/krakenjs/belter/blob/d0c0868c996aea92d1ad9a6c8662da7a2c200b5b/src/util.js#L188)

just curious..

jshawl: oh reading below it seems like it will cache based on the args passed

src/api/shopper-insights/component.test.js
```diff
@@ -67,20 +65,87 @@ describe("shopper insights component - getRecommendedPaymentMethods()", () => {
...
+    expect(response1).toEqual({
+      isPayPalRecommended: true,
+      isVenmoRecommended: false,
+    });
+    expect(response2).toEqual({
+      isPayPalRecommended: true,
+      isVenmoRecommended: false,
+    });
+    expect.assertions(6);
+  });
```
jshawl, line 106: awesome test coverage! The way you structured this test made it super easy to understand why the change is being made 🏅

src/api/shopper-insights/component.test.js
```diff
@@ -92,35 +157,39 @@ describe("shopper insights component - getRecommendedPaymentMethods()", () => {
...
         phone: {
           countryCode: "1",
-          nationalNumber: "2345678901",
+          nationalNumber: "2345678906",
         },
       },
     });
 
-    expect(callRestAPI).toHaveBeenCalledWith(
+    expect(request).toHaveBeenCalledWith(
```
jshawl, line 185: is there a reason you chose to expect `request` to be called instead of expecting `callMemoizedRestAPI` to be called? just curious (the change from `data` to `json` prompted this question for me)

jshawl (approved):
LGTM I had a few non-blocking questions just for learning purposes

wsbrunson (approved):

src/api/api.js
ravishekhar: Great question. You are 💯 right about args. Although memoize also creates cacheKey based on method arguments, but inlineMemoize generally provides more control on the cache key e.g. we can serialize complex objects as per the needs. 

Also with inlineMemoize the cache lives with method but memoize will end up creating a closure for the cache.

PR #2320, opened 2024-01-24
[ncp] handle create order errors

jshawl:

❤️ Thank you!


src/hosted-buttons/utils.js
```diff
@@ -126,16 +130,37 @@ export const buildHostedButtonCreateOrder = ({
...
+      div.classList.add(popupFallbackClassName);
+      div.innerHTML = popupFallback.replace("#", url);
+      const container =
+        typeof selector === "string"
+          ? document.querySelector(selector)
+          : selector;
+      container?.appendChild(div);
+    }
+  };
+};
```
jshawl, line 158: I'm not sure I like this implementation. I did it so that I didn't have to pass `popupFallback` and `selector` down into `onApprove`. e.g. the alternative to this would be:

```js
onApprove: buildHostedButtonOnApprove({
  merchantId,
  hostedButtonId,
  selector,
  popupFallback
})
```

and then inside of `buildHostedButtonOnApprove` would be the contents of this function. I did it this way so that I could easily unit-test it without managing passing around props but now that I'm looking at it in the GitHub diff it looks a little confusing. wdyt?

src/hosted-buttons/utils.js
```diff
@@ -149,19 +174,14 @@ export const buildHostedButtonOnApprove = ({
           context_id: data.orderID,
         }),
       }).then((response) => {
+        // remove the popup fallback message, if present
+        document.querySelector(`.${popupFallbackClassName}`)?.remove();
```
jshawl, line 178: this is another weird thing in this PR. I was really hoping to have all the dom manipulation contained within the popup handling code, but we need this here to handle a specific use case:
1. buyer uses inline guest to checkout, has popup blocked, error message is rendered in the dom
2. buyer uses a different payment method ("paypal") and when they click on the  button, there should no longer be an error message.

src/hosted-buttons/utils.js
nbierdeman: Ease of testing is a good reason in my book for going this route. Thanks for explaining this!

nbierdeman (approved):
Nice work! 💯

ravishekhar (approved):

PR #2327, opened 2024-02-05
feat: adds optional message prop to zoid interface

danzhaaspaypal:
### Description

Adds an optional message prop to zoid interface in anticipation of adding adjacent message support to checkout buttons.

"[buttons] Add message object option to zoid interface"

@Seavenly 

❤️ Thank you!

### DIFF
--- a/src/ui/buttons/buttons.jsx
+++ b/src/ui/buttons/buttons.jsx
@@ -179,6 +179,7 @@ export function Buttons(props: ButtonsProps): ElementNode {
     supportedNativeBrowser,
     showPayLabel,
     displayOnly,
+    // eslint-disable-next-line no-unused-vars
     message,
   } = normalizeButtonProps(props);
   const { layout, shape, tagline } = style;
@@ -286,7 +287,6 @@ export function Buttons(props: ButtonsProps): ElementNode {
           vault={vault}
           instrument={instruments[source]}
           showPayLabel={showPayLabel}
-          message={message}
         />
       ))}
 
--- a/src/ui/buttons/buttons.jsx
+++ b/src/ui/buttons/buttons.jsx
@@ -179,6 +179,7 @@ export function Buttons(props: ButtonsProps): ElementNode {
     supportedNativeBrowser,
     showPayLabel,
     displayOnly,
+    message,
   } = normalizeButtonProps(props);
   const { layout, shape, tagline } = style;
 
@@ -285,6 +286,7 @@ export function Buttons(props: ButtonsProps): ElementNode {
           vault={vault}
           instrument={instruments[source]}
           showPayLabel={showPayLabel}
+          message={message}
         />
       ))}
 
--- a/src/ui/buttons/props.js
+++ b/src/ui/buttons/props.js
@@ -746,23 +746,23 @@ export function normalizeButtonMessage(
         )}`
       );
     }
-    const invalidOffers = offer.filter((o) =>
-      values(MESSAGE_OFFER).includes(o)
+    const invalidOffers = offer.filter(
+      (o) => !values(MESSAGE_OFFER).includes(o)
     );
     if (invalidOffers.length > 0) {
       throw new Error(`Invalid offer(s): ${invalidOffers.join(",")}`);
     }
   }
 
-  if (color && values(MESSAGE_COLOR).includes(color)) {
+  if (color && !values(MESSAGE_COLOR).includes(color)) {
     throw new Error(`Invalid color: ${color}`);
   }
 
-  if (position && values(MESSAGE_POSITION).includes(position)) {
+  if (position && !values(MESSAGE_POSITION).includes(position)) {
     throw new Error(`Invalid position: ${position}`);
   }
 
-  if (align && values(MESSAGE_ALIGN).includes(align)) {
+  if (align && !values(MESSAGE_ALIGN).includes(align)) {
     throw new Error(`Invalid align: ${align}`);
   }
 
--- a/src/ui/buttons/props.js
+++ b/src/ui/buttons/props.js
@@ -435,7 +435,7 @@ export type ApplePaySessionConfigRequest = (
 
 export type ButtonMessage = {|
   amount?: number,
-  offer?: $ReadOnlyArray<?$Values<typeof MESSAGE_OFFER>>,
+  offer?: $ReadOnlyArray<$Values<typeof MESSAGE_OFFER>>,
   color?: $Values<typeof MESSAGE_COLOR>,
   position?: $Values<typeof MESSAGE_POSITION>,
   align?: $Values<typeof MESSAGE_ALIGN>,
@@ -443,7 +443,7 @@ export type ButtonMessage = {|
 
 export type ButtonMessageInputs = {|
   amount?: number | void,
-  offer?: $ReadOnlyArray<?$Values<typeof MESSAGE_OFFER>> | void,
+  offer?: $ReadOnlyArray<$Values<typeof MESSAGE_OFFER>> | void,
   color?: $Values<typeof MESSAGE_COLOR> | void,
   position?: $Values<typeof MESSAGE_POSITION> | void,
   align?: $Values<typeof MESSAGE_ALIGN> | void,
@@ -719,7 +719,7 @@ export function normalizeButtonStyle(
   };
 }
 
-export function validateButtonMessage(
+export function normalizeButtonMessage(
   props: ?ButtonPropsInputs,
   message: ButtonMessageInputs
 ): ButtonMessage {
@@ -727,19 +727,27 @@ export function validateButtonMessage(
     throw new Error(`Expected props.message to be set`);
   }
 
-  props = props || getDefaultButtonPropsInput();
-
   const { amount, offer, color, position, align } = message;
 
-  if (amount !== undefined) {
+  if (typeof amount !== "undefined") {
     if (typeof amount !== "number") {
       throw new TypeError(
-        `Expected style.height to be a number, got: ${amount}`
+        `Expected message.amount to be a number, got: ${amount}`
+      );
+    }
+    if (amount < 0) {
+      throw new TypeError(
+        `Expected message.amount to be a positive number, got: ${amount}`
       );
     }
   }
 
   if (offer) {
+    if (offer.constructor.name !== "Array") {
+      throw new TypeError(
+        `Expected message.offer to be an array of strings, got: ${offer}`
+      );
+    }
     const invalidOffers = offer.filter(
       (o) => values(MESSAGE_OFFER).indexOf(o) === -1
     );
@@ -882,7 +890,7 @@ export function normalizeButtonProps(
   }
 
   if (message) {
-    validateButtonMessage(props, message);
+    message = normalizeButtonMessage(props, message);
   }
 
   style = normalizeButtonStyle(props, style);
--- a/src/ui/buttons/props.js
+++ b/src/ui/buttons/props.js
@@ -40,6 +40,10 @@ import {
   BUTTON_SIZE,
   BUTTON_FLOW,
   MENU_PLACEMENT,
+  MESSAGE_OFFER,
+  MESSAGE_COLOR,
+  MESSAGE_POSITION,
+  MESSAGE_ALIGN,
 } from "../../constants";
 import { getFundingConfig, isFundingEligible } from "../../funding";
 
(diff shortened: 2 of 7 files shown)


src/ui/buttons/props.js
```diff
@@ -696,6 +719,56 @@ export function normalizeButtonStyle(
...
 
+export function validateButtonMessage(
+  props: ?ButtonPropsInputs,
+  message: ButtonMessageInputs
+): ButtonMessage {
+  if (!message) {
+    throw new Error(`Expected props.message to be set`);
+  }
+
+  props = props || getDefaultButtonPropsInput();
```
Seavenly, line 730: Is this being used anywhere or can it be removed?

src/zoid/buttons/component.jsx
```diff
@@ -877,6 +881,17 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+
+      message: {
+        type: "object",
+        queryParam: true,
+        required: false,
+        validate: ({ props, value = {} }) => {
+          // $FlowFixMe
+          validateButtonMessage(props, value);
+        },
+        default: () => ({}),
```
Seavenly, line 893: Will a default value of `{}` always turn on messages? We want this to be an opt-in feature so if so we likely want to remove that.

src/zoid/buttons/component.jsx
```diff
@@ -877,6 +881,17 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
           return props?.displayOnly || [];
         },
       },
+
+      message: {
+        type: "object",
+        queryParam: true,
+        required: false,
+        validate: ({ props, value = {} }) => {
```
Seavenly, line 889: Should we consider using the `decorate` zoid prop option as well similar to `style` which will ensure that the `message` object is formatted correctly and only contains the expected keys?

src/ui/buttons/props.js
```diff
@@ -696,6 +719,56 @@ export function normalizeButtonStyle(
...
+    }
+  }
+
+  if (offer) {
+    const invalidOffers = offer.filter(
+      (o) => values(MESSAGE_OFFER).indexOf(o) === -1
+    );
+    if (invalidOffers.length > 0) {
+      throw new Error(`Invalid offer(s): ${invalidOffers.join(",")}`);
+    }
```
Seavenly, line 748: Since we only need one invalid value to be in a failed state, instead of iterating over each value with filter could this be simplified with the following?
``` js
for (const o of offer) {
    if (values(MESSAGE_OFFER).indexOf(o) === -1) {
        throw new Error(`Invalid offer(s): ${invalidOffers.join(",")}`);
    }
}
```

src/ui/buttons/props.js
```diff
@@ -696,6 +719,56 @@ export function normalizeButtonStyle(
...
+
+  if (amount !== undefined) {
+    if (typeof amount !== "number") {
+      throw new TypeError(
+        `Expected style.height to be a number, got: ${amount}`
+      );
+    }
+  }
+
+  if (offer) {
```
Seavenly, line 742: Should we consider checking to ensure that this is actually an array before we try to iterate over it below?

src/ui/buttons/props.js
```diff
@@ -696,6 +719,56 @@ export function normalizeButtonStyle(
...
+): ButtonMessage {
+  if (!message) {
+    throw new Error(`Expected props.message to be set`);
+  }
+
+  props = props || getDefaultButtonPropsInput();
+
+  const { amount, offer, color, position, align } = message;
+
+  if (amount !== undefined) {
```
Seavenly, line 734: It's generally considered best practice to use `typeof amount !== "undefined"` due to old JS shenanigans.

src/ui/buttons/props.js
```diff
@@ -696,6 +719,56 @@ export function normalizeButtonStyle(
...
+    if (typeof amount !== "number") {
+      throw new TypeError(
+        `Expected style.height to be a number, got: ${amount}`
+      );
+    }
```
Seavenly, line 739: Can we add an additional check to ensure that the number is positive?

src/ui/buttons/props.js
```diff
@@ -696,6 +719,56 @@ export function normalizeButtonStyle(
   };
 }
 
+export function validateButtonMessage(
```
Seavenly, line 722: Consider `normalizeButtonMessage` to follow the naming convention of some of the other functions in the file.

src/ui/buttons/props.js
```diff
@@ -807,6 +881,10 @@ export function normalizeButtonProps(
     }
   }
 
+  if (message) {
+    validateButtonMessage(props, message);
+  }
```
Seavenly, line 886: Should we follow the same pattern as the `style` below?
`message = validateButtonMessage(props, message);`

src/zoid/buttons/component.jsx
Seavenly: We also likely don't want a default value of `{}` as mentioned in the comment about `default`.

src/ui/buttons/buttons.jsx
```diff
@@ -285,6 +286,7 @@ export function Buttons(props: ButtonsProps): ElementNode {
           vault={vault}
           instrument={instruments[source]}
           showPayLabel={showPayLabel}
+          message={message}
```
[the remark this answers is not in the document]
danzhaaspaypal: good callout, I'll remove that

src/ui/buttons/props.js
danzhaaspaypal: oop, that's an artifact from copying the `style` validation.  I'll remove that, good callout

src/zoid/buttons/component.jsx
danzhaaspaypal: good call, will remove

src/zoid/buttons/component.jsx
danzhaaspaypal: noted, will add decorate and remove `{}`

src/ui/buttons/props.js
danzhaaspaypal: @Seavenly while we're on the topic of `offer`'s type, should we consider converting a string value here to an array? Might give users a helping hand if they fail to notice we're looking for an array and not just the offer name.

src/ui/buttons/props.js
danzhaaspaypal: noted, will fix

src/ui/buttons/props.js
danzhaaspaypal: will do

src/ui/buttons/props.js
danzhaaspaypal: can do

src/ui/buttons/props.js
```diff
@@ -429,6 +433,22 @@ export type ApplePaySessionConfigRequest = (
   request: Object
 ) => ApplePaySessionConfig;
 
+export type ButtonMessage = {|
+  amount?: number,
+  offer?: $ReadOnlyArray<?$Values<typeof MESSAGE_OFFER>>,
```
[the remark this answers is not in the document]
danzhaaspaypal: Good callout, it seems we don't want the `?` since we don't want our array to accept null or undefined values. Will remove the `?`

src/ui/buttons/props.js
```diff
@@ -522,6 +543,7 @@ export type ButtonProps = {|
   createVaultSetupToken: CreateVaultSetupToken,
   displayOnly?: $ReadOnlyArray<$Values<typeof DISPLAY_ONLY_VALUES>>,
   hostedButtonId?: string,
+  message: ButtonMessage,
```
Seavenly, line 546: Should this be optional?

src/ui/buttons/props.js
Seavenly: I'm not sure there's precedence for that already for the buttons component so we may want to avoid that in favor of the errors pointing the user to the correct integration.

src/ui/buttons/props.js
```diff
@@ -696,6 +719,66 @@ export function normalizeButtonStyle(
...
+  if (typeof amount !== "undefined") {
+    if (typeof amount !== "number") {
+      throw new TypeError(
+        `Expected message.amount to be a number, got: ${amount}`
+      );
+    }
+    if (amount < 0) {
+      throw new TypeError(
+        `Expected message.amount to be a positive number, got: ${amount}`
+      );
```
Seavenly, line 741: This should likely be an `Error` instead of `TypeError` since we're validating the value.

src/ui/buttons/props.js
```diff
@@ -696,6 +719,66 @@ export function normalizeButtonStyle(
...
+    }
+  }
+
+  if (offer) {
+    if (offer.constructor.name !== "Array") {
```
Seavenly, line 746: Consider `Array.isArray(offer)` as best practice to type check arrays.

src/ui/buttons/props.js
danzhaaspaypal: @Seavenly I imagined this multiple invalid offer logging feature could be useful in some cases to save merchants a rerender to discover they have multiple invalid values instead of one.  For example, if there was a case where the merchant had a misspelling in the two offers they entered, they could learn they have to look at both values from a single error instead of having to rerender to learn about the second misspelling.  Satisfying use case?

Seavenly: That seems fair. Thanks for explaining the use case.

src/ui/buttons/props.js
```diff
@@ -696,6 +719,62 @@ export function normalizeButtonStyle(
...
+  if (offer) {
+    if (!Array.isArray(offer)) {
+      throw new TypeError(
+        `Expected message.offer to be an array of strings, got: ${String(
+          offer
+        )}`
+      );
+    }
+    const invalidOffers = offer.filter(
+      (o) => values(MESSAGE_OFFER).indexOf(o) === -1
```
wsbrunson, line 750:
```suggestion
      (o) => !values(MESSAGE_OFFER).includes(o)
```

src/ui/buttons/props.js
```diff
@@ -696,6 +719,62 @@ export function normalizeButtonStyle(
...
+    );
+    if (invalidOffers.length > 0) {
+      throw new Error(`Invalid offer(s): ${invalidOffers.join(",")}`);
+    }
+  }
+
+  if (color && values(MESSAGE_COLOR).indexOf(color) === -1) {
```
wsbrunson, line 757: we can change all of these `indexOf` to `includes`

src/ui/buttons/props.js
```diff
@@ -841,5 +925,6 @@ export function normalizeButtonProps(
     supportedNativeBrowser,
     showPayLabel,
     displayOnly,
+    message,
```
wsbrunson, line 928:
```suggestion
    message: message ? normalizeButtonMessage(props, message) : undefined,
```

and then remove the `if (message)` statement. I think its better if we don't override params/options from objects. you'll find it all over the place in this codebase but it makes it harder to understand when values are changing

src/zoid/buttons/component.jsx
```diff
@@ -877,6 +881,20 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+        queryParam: true,
+        required: false,
+        decorate: ({ props, value }) => {
+          // $FlowFixMe
+          return normalizeButtonMessage(props, value);
+        },
+        validate: ({ props, value }) => {
+          // $FlowFixMe
+          normalizeButtonMessage(props, value);
+        },
```
wsbrunson, line 896: totally fine to keep this here if you want it but just wanted to make sure you know: in production (and maybe every environment?) this will not throw an error and stop the integration. It will log to the console but it will continue to render

src/ui/buttons/buttons.jsx
```diff
@@ -179,6 +179,8 @@ export function Buttons(props: ButtonsProps): ElementNode {
     supportedNativeBrowser,
     showPayLabel,
     displayOnly,
+    // eslint-disable-next-line no-unused-vars
+    message,
```
wsbrunson, line 183: curious, why do we need this here if we aren't using it?

danzhaaspaypal: we aren't using it _yet_, but it will be the jumping-off point for use in another story in this feature, to be added soon.

src/zoid/buttons/component.jsx
Seavenly: Briefly synced internally to get more context around how `zoid` is handling this. Thanks!

wsbrunson (approved):
🏎️

nhanthanhnguyen95 (approved):

danzhaaspaypal:
@wsbrunson @nhanthanhnguyen95 @Seavenly Heads up, I had to make one last commit: the `message` value validation logic got flipped when I changed `indexOf (enum) === -1` to `includes`, and I had neglected to add `!` to account for that.  This commit makes that fix.  The PR now works as intended, as evidenced by passing all its karma tests.

wsbrunson (approved):

Seavenly (approved):

PR #2333, opened 2024-02-13
prepare to move Connect name to Fastlane

wsbrunson:
The Connect feature is being renamed to Fastlane

we can't remove Connect until all merchants have moved away from Connect. There will be a future PR to remove all usage of Connect once we have the go-ahead


elizabethmv (approved):
lgtm!

elizabethmv (approved):

PR #2329, opened 2024-02-09
Log SDK Integration Source

nbierdeman:
### Description

Herein lies part of the effort charged by [DTPPCPSDK-1482](https://paypal.atlassian.net/browse/DTPPCPSDK-1482).

### Why are we making these changes?

We need to pass `sdkIntegrationSource` as a query param to SPB for logging. Also, just for fun, I alphabetized the button component props, but am happy to back that out if anyone hates it.

### Screenshot

Screenshot 2024-02-09 at 12 05 11 PM

### Dependent Changes (if applicable)

This needs to be released before the server side changes will work.

PR #2334, opened 2024-02-13
Rename sdkIntegrationSource buttons zoid prop

nbierdeman:
### Description

We decided on a new name for tracking library usage, thus this PR. Alphabetization is under its own commit. 😁 Original PR: https://github.com/paypal/paypal-checkout-components/pull/2329


jshawl (approved):

nbierdeman:
Dear BundleMon, why?

src/zoid/buttons/component.jsx
```diff
@@ -610,6 +610,13 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
         value: getIntent,
       },
 
+      jsSdkLibrary: {
+        type: "string",
+        queryParam: true,
+        required: false,
```
jshawl, line 616: is `required: false` the default value? just curious. but also the `type: "string"` could come after `required`

jshawl (approved):
:shipit: one small non-blocking comment but otherwise LGTM!

PR #2335, opened 2024-02-14
feat: reserves space for button message and renders messageMarkup

danzhaaspaypal:
### Description

For the Buttons dynamic messaging feature:
In the Buttons component, reserves space for button message during 1st render and renders messageMarkup on second render.

[Dynamic Messaging with Buttons Epic](https://paypal.atlassian.net/browse/DTCRCMERC-2772) and [Update button render to reserve message space Story](https://paypal.atlassian.net/jira/software/c/projects/DTCRCMERC/boards/27252?issueParent=6748936&selectedIssue=DTCRCMERC-2804)

Pass in `message` prop to button component, optionally specifying its own child property `position` with 'top' or 'bottom'.  Conditionally, a 36 px space will populate above or below the group of buttons.  

Screenshot 2024-02-14 at 1 14 04 PM
Screenshot 2024-02-14 at 1 14 15 PM
Screenshot 2024-02-14 at 1 14 21 PM

n/a

Upstream Messaging
XO components

❤️ Thank you!


src/ui/buttons/buttons.jsx
```diff
@@ -150,6 +152,47 @@ export function validateButtonProps(props: ButtonPropsInputs) {
   normalizeButtonProps(props);
 }
 
+function calculateMsgPosition({
```
Seavenly, line 155: Consider `calculateMessagePosition` so that we aren't using abbreviations

src/ui/buttons/buttons.jsx
```diff
@@ -150,6 +152,47 @@ export function validateButtonProps(props: ButtonPropsInputs) {
...
+  layout,
+  fundingSources,
+  fundingSource,
+}): string {
+  if (!message) {
+    return null;
+  }
+  const { position } = message;
+  // Cases for throwing an error
+  if (tagline && layout === BUTTON_LAYOUT.HORIZONTAL && !fundingSource) {
```
Seavenly, line 167: Would it be worthwhile to turn this into a variable such as `showTagline` in the `Buttons` function that could be passed into this function so that `showTagline` could be reused also where the `<Tagline />` component is being considered. This could help introducing an accidental bug in the future if the conditional here to determine whether or not to display the tagline is updated in one place but not the other.

src/ui/buttons/buttons.jsx
```diff
@@ -150,6 +152,47 @@ export function validateButtonProps(props: ButtonPropsInputs) {
...
+    throw new ValidationError("Button cannot show both Message and Tagline.");
+  }
+  if (
+    fundingSources.indexOf(FUNDING.CARD) !== -1 &&
+    position === MESSAGE_POSITION.BOTTOM
```
Seavenly, line 172: Similar comment as above but for something like `showPoweredBy`. Also consider updating `fundingSources.indexOf(FUNDING.CARD) !== -1` to `fundingSources.includes(FUNDING.CARD)` to align with best practices.

src/ui/buttons/buttons.jsx
```diff
@@ -150,6 +152,47 @@ export function validateButtonProps(props: ButtonPropsInputs) {
...
+    return MESSAGE_POSITION.TOP;
+  } else if (
+    (layout === BUTTON_LAYOUT.VERTICAL &&
+      position === MESSAGE_POSITION.BOTTOM) ||
+    position === MESSAGE_POSITION.BOTTOM ||
+    layout === BUTTON_LAYOUT.HORIZONTAL ||
+    layout
+  ) {
+    return MESSAGE_POSITION.BOTTOM;
+  }
```
Seavenly, line 193: Could this entire `else if` block be removed in favor of just having `return MESSAGE_POSTION.BOTTOM`? From the way this function appears to be structured, the top logic is responsible for determining if a message can be displayed, and this bottom section determines if it's top or bottom. If there's other scenarios where we'd prefer to return `null`, we should consider including them above. e.g. `if (...) { return null; }`

src/ui/buttons/buttons.jsx
```diff
@@ -242,6 +285,14 @@ export function Buttons(props: ButtonsProps): ElementNode {
     return i;
   };
 
+  const calculatedMsgPosition = calculateMsgPosition({
```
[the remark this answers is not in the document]
danzhaaspaypal: I think it could use a distinct name from `message.position` because while it's represented differently in the code vs `messagePosition`, they can be confused when talking about them unless the devs write the names perfectly every time.  Otherwise, "message position" could be interpreted 2 ways.  

I don't mind losing the abbreviation though to `calculatedMessagePosition` or i am open to other names

src/ui/buttons/buttons.jsx
danzhaaspaypal: very good call

src/ui/buttons/buttons.jsx
danzhaaspaypal: this is true, I may have just included the bottom block so i could see that I factored in every condition while I was working, but now it's unnecessary. I'll remove that

src/ui/buttons/buttons.jsx
danzhaaspaypal: on second look, i specified `fundingSources.includes(FUNDING.CARD)` here intentionally to cover both PoweredBy and this condition from the story's notes:

- Revert to top position if merchant puts bottom when the Debit or Credit card button is shown (this button already displays a message below it)

Seavenly: Should we be accounting for the `layout === BUTTON_LAYOUT.VERTICAL` of the `showPoweredBy`? Would we want `showPoweredBy && position === MESSAGE_POSITION.BOTTOM`?

src/ui/buttons/message.jsx
```diff
@@ -0,0 +1,32 @@
...
+const INITIAL_RESERVED_HEIGHT = "36px";
+
+type MessageProps = {|
+  messageMarkup: ?string,
+  calculatedMsgPosition: string,
+|};
+
+export function Message({
+  messageMarkup,
+  calculatedMessagePosition,
```
Seavenly, line 17: Since these are in the limited context of a `<Message />` component could they be shortened to just `markup` and `position`?

src/ui/buttons/message.jsx
```diff
@@ -0,0 +1,32 @@
...
+}: MessageProps): ChildType {
+  const reservationDiv = `<div class="${CLASS.BUTTON_MESSAGE_RESERVE}" style="height:${INITIAL_RESERVED_HEIGHT}" ><p>test</p></div>`;
```
Seavenly, line 19: Note to remove `<p>test</p>`

src/ui/buttons/buttons.jsx
danzhaaspaypal: you were right after all - fixed

test/integration/tests/button/message.js
```diff
@@ -0,0 +1,408 @@
...
+                done(new Error("message space was reserved incorrectly"));
+              }
+              done();
+            },
+          },
+        })
+        .render("#testContainer");
+    });
+
+    it("should not reserve space for a message when messageMarkup is truthy", (done) => {
```
Seavenly, line 89: Should this be `...when messageMarkup is string` since we have a `typeof` string check as the decision?

src/ui/buttons/buttons.jsx
```diff
@@ -150,6 +152,28 @@ export function validateButtonProps(props: ButtonPropsInputs) {
...
 
+function calculateMessagePosition({ message, showPoweredBy, layout }): string {
+  if (!message) {
+    return "no message";
+  }
+  const { position } = message;
+
+  if (showPoweredBy && position === MESSAGE_POSITION.BOTTOM) {
+    throw new ValidationError(
+      "Message position must be 'top' when Debit and/or Credit Card is a funding source"
```
Seavenly, line 163:
```suggestion
      "Message position must be 'top' when Debit and/or Credit Card button is is present"
```

test/integration/tests/button/message.js
```diff
@@ -0,0 +1,408 @@
...
+            },
+            test: {
+              onRender() {
+                assert.ok(getElementRecursive(".paypal-button-message-bottom"));
+                done();
+              },
+            },
+          })
+          .render("#testContainer");
+      });
```
Seavenly, line 250: Is this just testing the same scenario as L172 since the horizontal layout will only show 2 buttons?

test/integration/tests/button/message.js
```diff
@@ -0,0 +1,408 @@
...
+            },
+            test: {
+              onRender() {
+                assert.ok(getElementRecursive(".paypal-button-message-top"));
+                done();
+              },
+            },
+          })
+          .render("#testContainer");
+      });
```
Seavenly, line 343: Similar comment as above being similar to the scenario on L253.

test/integration/tests/button/message.js
danzhaaspaypal: I wanted to have tests that differentiate these cases:

1. messageMarkup is `""`
2. messageMarkup is a string with length > 0.

Would those labels be better?

test/integration/tests/button/message.js
danzhaaspaypal: now that you mention it, yes; I'll nix this test then

test/integration/tests/button/message.js
danzhaaspaypal: true. will nix

src/ui/buttons/buttons.jsx
```diff
@@ -261,6 +299,13 @@ export function Buttons(props: ButtonsProps): ElementNode {
         fundingEligibility={fundingEligibility}
       />
 
+      {message && calculatedMessagePosition === MESSAGE_POSITION.TOP ? (
+        <Message
+          markup={messageMarkup}
+          position={(calculatedMessagePosition: string)}
```
Seavenly, line 305: Does this need to be cast to a string now?

Seavenly (approved):

src/ui/buttons/buttons.jsx
```diff
@@ -150,6 +152,28 @@ export function validateButtonProps(props: ButtonPropsInputs) {
...
+
+  if (
+    showPoweredBy ||
+    position === MESSAGE_POSITION.TOP ||
+    (layout === BUTTON_LAYOUT.VERTICAL && !position)
+  ) {
+    return MESSAGE_POSITION.TOP;
+  }
+  return MESSAGE_POSITION.BOTTOM;
+}
```
jshawl, line 175: option to move this function `ui/buttons/message.jsx` (I did not see it used elsewhere and/or reference variables in the parent scope)

src/ui/buttons/buttons.jsx
```diff
@@ -150,6 +152,28 @@ export function validateButtonProps(props: ButtonPropsInputs) {
   normalizeButtonProps(props);
 }
 
+function calculateMessagePosition({ message, showPoweredBy, layout }): string {
+  if (!message) {
+    return "none";
```
jshawl, line 157: option to add a `NONE` constant here - https://github.com/paypal/paypal-checkout-components/blob/fbe73816fafc65d6fde81bb43d4107548e6effd7/src/constants/button.js#L77-L80 (for consistency) and change the return type to `$Keys<MESSAGE_POSITION>;`  (just kind of cool)

src/ui/buttons/message.jsx
```diff
@@ -0,0 +1,26 @@
...
+  const reservationDiv = `<div class="${CLASS.BUTTON_MESSAGE_RESERVE}" style="height:${INITIAL_RESERVED_HEIGHT}" ></div>`;
+
+  return (
+    <div
+      class={[CLASS.BUTTON_MESSAGE, `${CLASS.BUTTON_MESSAGE}-${position}`].join(
+        " "
+      )}
+      innerHTML={typeof markup === "string" ? markup : reservationDiv}
+    />
+  );
```
wsbrunson, line 25: Github UI probably won't format this well but I think something like this would be clearer in intention:

```suggestion
  const messageClassNames = [CLASS.BUTTON_MESSAGE, `${CLASS.BUTTON_MESSAGE}-${position}`].join(" ")

  if (!typeof markup === "string") {
    return <div className={messageClassNames}>
      <div className={CLASS.BUTTON_MESSAGE_RESERVE} style={`height:${INITIAL_RESERVED_HEIGHT}`} />
     </div>
  }

  return (
    <div
      className={messageClassNames}
      innerHTML={markup}
    />
  );
```

wsbrunson: or you could do something like this:
```suggestion
  const messageClassNames = [CLASS.BUTTON_MESSAGE, `${CLASS.BUTTON_MESSAGE}-${position}`].join(" ")
  
  if (!typeof markup === "string") {
    return <div className={`${messageClassNames} $CLASS.BUTTON_MESSAGE_RESERVE}`} style={`height:${INITIAL_RESERVED_HEIGHT}`} />
  }
  
  return (
    <div
      className={messageClassNames}
      innerHTML={markup}
    />
  );
```

wsbrunson (approved):

src/ui/buttons/buttons.jsx
```diff
@@ -242,6 +266,20 @@ export function Buttons(props: ButtonsProps): ElementNode {
     return i;
   };
 
+  const showTagline =
+    tagline &&
+    layout === BUTTON_LAYOUT.HORIZONTAL &&
+    !fundingSource &&
+    !message;
+  const showPoweredBy =
+    layout === BUTTON_LAYOUT.VERTICAL && fundingSources.includes(FUNDING.CARD);
```
jshawl, line 275: I think `fundingSources` will always include `FUNDING.CARD` - https://github.com/paypal/paypal-checkout-components/blob/fbe73816fafc65d6fde81bb43d4107548e6effd7/src/ui/buttons/buttons.jsx#L240-L245 (so can remove the `&& fundingSources.includes`)

test/integration/tests/button/message.js
```diff
@@ -0,0 +1,381 @@
...
+                assert.ok(getElementRecursive(".paypal-button-message-bottom"));
+                done();
+              },
+            },
+          })
+          .render("#testContainer");
+      });
+    });
+  });
+});
```
jshawl, line 381: awesome test coverage!! 💯 🏅

jshawl (approved):
I added a few non-blocking comments but LGTM 🚀

src/ui/buttons/buttons.jsx
wsbrunson: I actually don't think it would. Its confusing, but the standalone button case uses `Buttons` component but with `fundingSource=paypal|venmo|card|etc`. I think this check is precisely for those instances where we are in Standalone and its not the Card button

jshawl: ah good catch! thank you for clarifying that!

wsbrunson: Not a problem, actually that would be a good candidate for a refactor. It would probably simplify things if we created a `StandaloneButton` component and rendered that instead of `Buttons` if `fundingSource` was passed

src/ui/buttons/message.jsx
wsbrunson: also also `className -> class` in all my examples i forgot i'm not writing React 😂

src/ui/buttons/buttons.jsx
Seavenly: We do not want `NONE` to be considered a valid input option which `MESSAGE_POSITION` is currently used for as a type. It's purely an internal computed value. Would it be beneficial to have a separate constant to capture the possible internal values?

jshawl: ah that's a good point! I think if it were me I would leave it as-is. one other idea just for fun would be to make the return type `string | void` and the early return would return nothing/undefined. But I think I like what you have here better than that

danzhaaspaypal: @jshawl Nate and I tried a few different approaches being mindful of type returned here; the `null` return approach was our first attempted strategy, and if Flow would accommodate 'null', we would gladly go with that option.  But it didn't exactly mind our logic correctly: it tried to suggest `null` should be a possible value in the `Message` component where it cannot logically be passed in due to logic on [lines 302 and 359](https://github.com/paypal/paypal-checkout-components/pull/2335/files/fbe73816fafc65d6fde81bb43d4107548e6effd7#diff-3fba3feb650bae3f6bfc93d232bac1d8cfde682d5bfffbdd40d8cdfeda7fcd8aR302).  And it seemed wrong to abide that workaround by accommodating `null` inside Message where it could not appear.

We tried a few other typing workarounds, and finally settled on this being the least of the evils.

test/integration/tests/button/message.js
danzhaaspaypal: @jshawl thank you!

src/ui/buttons/buttons.jsx
jshawl: this is flow's fault 😂

src/ui/buttons/buttons.jsx
danzhaaspaypal: I like the idea of decluttering `buttons.jsx`; I'd like to suggest `ui/buttons/util.js` as another good place to consider moving that function.  Thoughts, preference?

jshawl: I think `ui/buttons/util.js` would work well! 🚀

PR #2337, opened 2024-02-20
hosted buttons: use Promise instead of ZalgoPromise

jshawl:
### Description

This PR is a light refactor to use `Promise` instead of `ZalgoPromise` in preparation for some [incoming changes](https://github.com/paypal/paypal-sdk-client/pull/184) from `@paypal/sdk-client`.

I started working on consuming [some](https://github.com/paypal/paypal-sdk-client/pull/186) [changes](https://github.com/paypal/paypal-sdk-client/pull/187) from `@paypal/sdk-client` but realized the diff was quite large. I wanted to separate out the refactor from the new functionality to make the diffs easier to read.

because of the change from `.then(() => {...` to `await` I recommend reviewing this PR with the [whitespace ignored](https://github.com/paypal/paypal-checkout-components/pull/2337/files?diff=unified&w=1)

❤️ Thank you!

### DIFF
--- a/src/hosted-buttons/index.js
+++ b/src/hosted-buttons/index.js
@@ -20,35 +20,34 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
     hostedButtonId,
   }: HostedButtonsComponentProps): HostedButtonsInstance {
     const Buttons = getButtonsComponent();
-    const render = (selector) => {
+    const render = async (selector) => {
       const merchantId = getMerchantID();
+      const { html, htmlScript, style } = await getHostedButtonDetails({
+        hostedButtonId,
+      });
 
-      getHostedButtonDetails({ hostedButtonId }).then(
-        ({ html, htmlScript, style }) => {
-          const { onInit, onClick } = renderForm({
-            hostedButtonId,
-            html,
-            htmlScript,
-            selector,
-          });
+      const { onInit, onClick } = renderForm({
+        hostedButtonId,
+        html,
+        htmlScript,
+        selector,
+      });
 
-          // $FlowFixMe
-          Buttons({
-            hostedButtonId,
-            style,
-            onInit,
-            onClick,
-            createOrder: buildHostedButtonCreateOrder({
-              hostedButtonId,
-              merchantId,
-            }),
-            onApprove: buildHostedButtonOnApprove({
-              hostedButtonId,
-              merchantId,
-            }),
-          }).render(selector);
-        }
-      );
+      // $FlowFixMe
+      Buttons({
+        hostedButtonId,
+        style,
+        onInit,
+        onClick,
+        createOrder: buildHostedButtonCreateOrder({
+          hostedButtonId,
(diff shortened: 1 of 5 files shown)


src/api/shopper-insights/validation.js
```diff
@@ -1,5 +1,6 @@
 /* @flow */
 
+// eslint-disable-next-line import/no-deprecated
```
jshawl, line 3: this is fixed in https://github.com/paypal/paypal-checkout-components/pull/2332

src/hosted-buttons/index.js
```diff
@@ -20,35 +20,34 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
     hostedButtonId,
   }: HostedButtonsComponentProps): HostedButtonsInstance {
     const Buttons = getButtonsComponent();
-    const render = (selector) => {
+    const render = async (selector) => {
       const merchantId = getMerchantID();
+      const { html, htmlScript, style } = await getHostedButtonDetails({
```
nikrom17, line 25: I've been meaning to ask why most of the async code we write uses the .then syntax instead of the async/await syntax. 

I'm all about the async/await syntax but was curious if there were any team norms around when to use either approach

nikrom17 (approved):

src/hosted-buttons/index.js
wsbrunson: if you are working inside of `spb` its best to use `.then` and assume you are working with `ZalgoPromise`. Also if you are working in the `buttons` part of `checkout-components`

There is a very specific part of SPB that has to be synchronous or else the popup will fail. there's also a lot of code that expects zalgo promises to be returned.

If you are making new components though, its usually safe to adopt async/await and native Promises.

jshawl: 2 motivations for me to use `ZalgoPromise` instead of `async/await`:
- a lot of the kraken utility functions return `ZalgoPromise` 
  - https://github.com/krakenjs/belter/blob/449034e63bca445708b62d9e08f7190a44ba4169/src/http.js#L48
  - https://github.com/krakenjs/belter/blob/449034e63bca445708b62d9e08f7190a44ba4169/src/dom.js#L216
  - https://github.com/krakenjs/post-robot/blob/30ab862e90a023531cd46d5fe8a7342c887b521b/src/drivers/receive/index.js#L147 
  - which means that:
- any consumers of those utility functions have to return a `ZalgoPromise` _or_ explicitly:
  - eslint-disable the usage of `Promise` https://github.com/krakenjs/grumbler-scripts/blob/de4b7a3490bff44cc9ebb7dc88eea87f64b96a36/packages/eslint-config-grumbler/eslintrc-browser.js#L9
  - `$FlowIssue` lines that don't use `ZalgoPromise` 

I think it's definitely worth changing the `eslint` rule to allow promises and/or rewriting (or providing alternatives to) some of the utility functions to return native promises

spencersablan (approved):

PR #2338, opened 2024-02-20
fix Error: Expected client-id parameter in sdk url

jshawl:
### Description

This PR fixes an issue that was introduced with the latest version of `@paypal/sdk-client` 

CI is failing for new PRs to paypal-checkout-components.

run `npm run karma` with v4.0.181 of `@paypal/sdk-client`

Screenshot 2024-02-20 at 15 01 15

related - https://github.com/paypal/paypal-checkout-components/pull/2332

❤️ Thank you!

### DIFF
--- a/src/api/shopper-insights/validation.js
+++ b/src/api/shopper-insights/validation.js
@@ -1,5 +1,6 @@
 /* @flow */
 
+// eslint-disable-next-line import/no-deprecated
 import { sendCountMetric } from "@paypal/sdk-client/src";
 
 import {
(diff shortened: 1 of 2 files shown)


src/api/shopper-insights/validation.js
```diff
@@ -1,5 +1,6 @@
 /* @flow */
 
+// eslint-disable-next-line import/no-deprecated
```
jshawl, line 3: this is fixed in https://github.com/paypal/paypal-checkout-components/pull/2332 but that can't be merged until the client-id error is fixed (in this PR)

test/paypal.js
```diff
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { setupSDK, insertMockSDKScript } from "@paypal/sdk-client/src";
+import { setupSDK } from "@paypal/sdk-client/src";
```
jshawl, line 3: there are a few other repos that will need a similar change - https://github.com/search?q=org%3Apaypal%20insertMockSDKScript&type=code (I think) so a follow-up PR could fix the `insertMockSDKScript` implementation and revert this current PR (just FYI)

wsbrunson (approved):

PR #2340, opened 2024-02-21
Add `enableDPoP` flag for Hosted Buttons

jshawl:
### Description

This PR adds a new prop `enableDPoP` to the hosted buttons component to feature-flag adding DPoP headers to `/v1/oauth2/token` and protected resource requests.

```js
paypal.HostedButtons({
  hostedButtonId: "H1234567890",
  enableDPoP: true // 🆕
})
```

When `enableDPoP` is `true`, requests to create an access token will include a `DPoP` header (see https://datatracker.ietf.org/doc/html/rfc9449#name-dpop-access-token-request) that includes information about the request (request method and uri) and information about the device (the in-memory public key generated from the buyer's browser) and signed with an in-memory and non-extractable private key.

When `enableDPoP` is `true`, requests to protected resources will replace the existing `Authorization: Bearer <token>` with the new authentication scheme: `Authorization: DPoP <token>` and also include a `DPoP` header. (see https://datatracker.ietf.org/doc/html/rfc9449#name-protected-resource-access).

When `enableDPoP` is `false` (or `undefined`), no existing functionality is changed. This property is used as a feature flag only before general availability.

The hosted buttons component creates access tokens cross-origin from the merchant's site. DPoP enforces a policy where access tokens can only be used from the device that requested the access token. 

From https://datatracker.ietf.org/doc/html/rfc9449#name-objectives:
>The primary aim of DPoP is to prevent unauthorized or illegitimate parties from using leaked or stolen access tokens, by binding a token to a public key upon issuance and requiring that the client proves possession of the corresponding private key when using the token.

### Reproduction Steps

add `enableDPoP: true` to `paypal.HostedButtons({})`

### Screenshots

See an example DPoP proof (and its contents) validated in the [jwt.io debugger](https://jwt.io/#debugger-io?token=eyJhbGciOiJFUzI1NiIsInR5cCI6ImRwb3Arand0IiwiandrIjp7ImNydiI6IlAtMjU2IiwiZXh0Ijp0cnVlLCJrZXlfb3BzIjpbInZlcmlmeSJdLCJrdHkiOiJFQyIsIngiOiIwazNxYWVKTnhMc1lHeFE5Y2NaeFpSa015VEI0VzhZclVPWk5sQlFvbWFrIiwieSI6IlRJVVQwZ2xqb2ZVcFl2ajh0NGtOeV9sWmJzc3ExREJ2cXZOa3lZeGtIajgifX0.eyJjbmYiOnsiamt0IjoicDJfLTNXOHpQZjZxUi1mYXl3bjA1M3U3ajdTM3JTTkJIeU9Jd2VhWWZPUSJ9LCJodG0iOiJQT1NUIiwiaHR1IjoiaHR0cHM6Ly9sb2NhbGhvc3QucGF5cGFsLmNvbTo4NDQzL3YxL29hdXRoMi90b2tlbiIsImlhdCI6MTcwODYwMzY0MSwianRpIjoiZjRmNWNiMjEtMWJlMy00OTcyLTlmMWUtNDA2NTdlMGVkYzYyIn0.PT_pwpe20pj71hH66KSSGcprpxXmzAnIHO8J1jCBis9X76Bi5N-J_KGc62wQ4j6VecD_cLwBvyTzDtRB_-0gnw&publicKey=%7B%0A%20%20%22crv%22%3A%20%22P-256%22%2C%0A%20%20%22ext%22%3A%20true%2C%0A%20%20%22key_ops%22%3A%20%5B%0A%20%20%20%20%22verify%22%0A%20%20%5D%2C%0A%20%20%22kty%22%3A%20%22EC%22%2C%0A%20%20%22x%22%3A%20%220k3qaeJNxLsYGxQ9ccZxZRkMyTB4W8YrUOZNlBQomak%22%2C%0A%20%20%22y%22%3A%20%22TIUT0gljofUpYvj8t4kNy_lZbssq1DBvqvNkyYxkHj8%22%0A%7D) .

I also wanted to include the above link to show that no sensitive information is included in the `DPoP` proof. Even when an access token is passed to `buildDPoPHeaders`, it is hashed ([ath](https://datatracker.ietf.org/doc/html/rfc9449#section-4.2-6.2)).

### Dependent Changes

Everything related to JWT creation, key pair generation, and signatures is implemented in `@paypal/sdk-client`:

- https://github.com/paypal/paypal-sdk-client/pull/184
- https://github.com/paypal/paypal-sdk-client/pull/186
- https://github.com/paypal/paypal-sdk-client/pull/187

Future PRs:

- this feature flag can be removed (and default to DPoP requests) when e2e testing has been completed.

❤️ Thank you!

### DIFF
--- a/src/hosted-buttons/types.js
+++ b/src/hosted-buttons/types.js
@@ -59,21 +59,4 @@ export type RenderForm = ({|
   onClick: (data: mixed, actions: mixed) => void,
 |};
 
-type Request = {|
-  url: string,
-  headers: {|
-    [key: string]: string,
-    Authorization?: string,
-  |},
-  method: string,
-  body: string,
-|};
-
-type Response = {|
-  status: number,
-  headers: { [string]: string },
-  body: Object,
-|};
-
-export type RequestWithDPoP = (Request) => Response;
 /* eslint-enable no-restricted-globals, promise/no-native */
--- a/src/hosted-buttons/types.js
+++ b/src/hosted-buttons/types.js
@@ -6,6 +6,7 @@ export type HostedButtonsComponentProps = {|
 |};
 
 export type GetCallbackProps = {|
+  enableDPoP?: boolean,
   hostedButtonId: string,
   merchantId?: string,
 |};
@@ -40,7 +41,10 @@ export type OnApprove = (data: {|
   paymentSource: string,
 |}) => Promise<mixed>;
 
-export type CreateAccessToken = (clientID: string) => Promise<string>;
+export type CreateAccessToken = ({|
+  clientId: string,
+  enableDPoP?: boolean,
+|}) => Promise<{| accessToken: string, nonce: string |}>;
 
 export type HostedButtonsComponent =
   (HostedButtonsComponentProps) => HostedButtonsInstance;
@@ -55,4 +59,21 @@ export type RenderForm = ({|
   onClick: (data: mixed, actions: mixed) => void,
 |};
 
+type Request = {|
+  url: string,
+  headers: {|
+    [key: string]: string,
+    Authorization?: string,
+  |},
+  method: string,
+  body: string,
+|};
+
+type Response = {|
+  status: number,
+  headers: { [string]: string },
+  body: Object,
+|};
+
+export type RequestWithDPoP = (Request) => Response;
 /* eslint-enable no-restricted-globals, promise/no-native */
--- a/src/hosted-buttons/utils.js
+++ b/src/hosted-buttons/utils.js
@@ -43,18 +43,22 @@ export const createAccessToken: CreateAccessToken = memoize<CreateAccessToken>(
   async ({ clientId, enableDPoP }) => {
     const url = `${apiUrl}/v1/oauth2/token`;
     const method = "POST";
+    const DPoPHeaders = enableDPoP
+      ? await buildDPoPHeaders({
+          uri: url,
+          method,
+        })
+      : {};
     const response = await request({
       url,
       method,
       body: "grant_type=client_credentials",
+      // $FlowIssue optional properties are not compatible with [key: string]: string
       headers: {
         Authorization: `Basic ${btoa(clientId)}`,
         "Content-Type": "application/json",
-        ...(enableDPoP &&
-          (await buildDPoPHeaders({
-            uri: url,
-            method,
-          }))),
+        // $FlowIssue exponential-spread
+        ...DPoPHeaders,
       },
     });
     // $FlowIssue request returns ZalgoPromise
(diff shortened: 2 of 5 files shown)


src/hosted-buttons/utils.js
```diff
@@ -38,9 +40,43 @@ export const getMerchantID = (): string | void => {
...
 };
 
+export const requestWithDPoP: RequestWithDPoP = async ({
+  url,
+  headers,
+  method,
+  body,
+}) => {
+  let accessToken, nonce;
+  if (!headers.Authorization?.match(/Basic/)) {
```
jshawl, line 50: this seems a little odd - thinking out loud here:
- I wanted to avoid rebuilding the headers in the three functions that call apis (`createAccessToken` `buildHostedButtonCreateOrder`, `buildHostedButtonCaptureOrder`) because:
- I would have to update the test cases for each of those (doubling the number of tests) but also I like how the existing tests are not updated to show that this new feature flag does not break existing functionality.
- this is feature flagged code that will be removed in the near(ish) future
- given that^, we can't pass `accessToken` to `requestWithDPoP` because `request` doesn't accept an `accessToken` parameter
- I thought parsing out the access token from the headers was less hacky than having if/else conditions in each of the api-calling methods

src/hosted-buttons/types.js
```diff
@@ -55,4 +59,21 @@ export type RenderForm = ({|
...
+  |},
+  method: string,
+  body: string,
+|};
+
+type Response = {|
+  status: number,
+  headers: { [string]: string },
+  body: Object,
+|};
```
jshawl, line 76: the two above types are modified from belter, which are not exported - https://github.com/krakenjs/belter/blob/449034e63bca445708b62d9e08f7190a44ba4169/src/http.js#L6-L21

alternatively, I can add `export` to those two types, publish a new version of belter and consume here (but also see `Authorization?` comment above ^^

src/hosted-buttons/types.js
```diff
@@ -55,4 +59,21 @@ export type RenderForm = ({|
   onClick: (data: mixed, actions: mixed) => void,
 |};
 
+type Request = {|
+  url: string,
+  headers: {|
+    [key: string]: string,
+    Authorization?: string,
```
jshawl, line 66: I couldn't figure out a way to do this with flow:

Screenshot 2024-02-22 at 14 56 52

so I settled for adding the optional `Authorization` property (even though we know it won't be in this object if undefined)

src/hosted-buttons/utils.js
ravishekhar: Thinking out loud here, Can we encapsulate the `request` method inside another method `request_WithOut_DPoP` (we need a better name though) .
and `requestFn` can be either this or the other method. 
e.g. `const requestFn = enableDPoP ? requestWithDPoP : requestWithOutDPoP;` 

Under the hood both methods can invoke a memoized version of `createAccessToken` and then call `request`

jshawl: I like this suggestion! I tried it out but ran into the opposite issue, where `createAccessToken` doesn't have enough information to use `Authorization: Basic`. I have another idea (move this logic to `getHeaders` that I think will improve the readability and not require matching on the headers - stay tuned!

jshawl: I like this a lot better! a9299614 wdyt? (I abandoned the abstracted `requestWithDPoP` in favor of moving all the complexity to different test cases)

ravishekhar (approved):
🎉

PR #2348, opened 2024-02-28
Skip screenshot tests in CI

nbierdeman:
### Description

This PR is a compromise, but a reasonable one, in my opinion. Jest screenshot tests pass locally, but fail in CI. Valiant attempts were made at fixing them (please see the [commit history](https://github.com/paypal/paypal-checkout-components/pull/2348/commits) for excruciating detail), but it was ultimately decided by **Product** that it was time to move on. Instead, screenshot tests run locally in a pre-push hook, but are skipped in CI, including when `git push` is run in CI. Another way of thinking of these changes is every step in `npm run test` happens between the local pre-push hook and the GitHub Actions tasks. If all goes well, this should unblock publishing `v4` using GitHub Actions.

DTPPCPSDK-1978

### Screenshots

Here you can see the pre-push hook running screenshot tests locally:

Screenshot 2024-03-06 at 2 02 31 PM


.husky/pre-push
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+# Check if not running in GitHub Actions
+if [ -z "$CI" ]; then
+    echo "Running pre-push hook in local environment"
+    npm run buttondiff
+    exit 0
```
jshawl, line 8: this can be `exit $?` to reuse the exit code from `npm run buttondiff` (which I think will exit 1 for a failure) but wanted to mention this will still allow the push even if the `npm run buttondiff` fails (I think) (which I'm also ok with)

.husky/pre-push
```diff
@@ -0,0 +1,12 @@
...
+else
+    echo "Skipping pre-push hook in CI environment"
+    exit 0
+fi
```
jshawl, line 12: I thought this compromise was clever! nice!

package.json
```diff
@@ -10,11 +10,11 @@
     "flow:build": "flow gen-flow-files ./src/index.js --out-dir ./dist/module",
     "karma": "cross-env NODE_ENV=test babel-node --plugins=transform-es2015-modules-commonjs ./node_modules/.bin/karma start",
     "babel": "babel src/ --out-dir dist/module",
-    "webpack": "babel-node ./node_modules/.bin/webpack --progress",
+    "webpack": "babel-node ./node_modules/.bin/webpack",
     "webpack-base": "babel-node ./node_modules/.bin/webpack --progress --config webpack.base.config.js",
     "webpack-button": "babel-node ./node_modules/.bin/webpack --progress --config webpack.button.config.js",
-    "test": "if [ ! $SKIP_TEST ]; then npm run lint && npm run flow-typed && npm run flow && npm run test-ssr && npm run karma && npm run buttondiff; fi;",
-    "build": "npm run test && npm run webpack",
+    "test": "npm run lint && npm run flow-typed && npm run flow && npm run test-ssr && npm run karma && npm run buttondiff",
```
jshawl, line 16: should `npm run buttondiff` also be removed from `test`? (will CI continue to hang?)

jshawl (approved):
a commendable effort and a clever compromise! :shipit:

package.json
nbierdeman: `npm run test` doesn't get run by CI. It's really only there as a convenience script for local development (to be run manually).

.husky/pre-push
nbierdeman: Nice catch! 🙏

nbierdeman: It looks like `git push` doesn't complete when `npm run buttondiff` returns `exit 1`:

Screenshot 2024-03-06 at 2 52 31 PM

jshawl: yes! I thought this was the desired behavior (if buttondiff fails - don't push)

PR #2352, opened 2024-03-04
feat: message hover and click behavior and modal

danzhaaspaypal:
### Description

- Ensure on message click a modal is opened up on the parent page
- Modal should display the product and amount for the message that was clicked
- Modal should lazy load modal.js upon click
- If the messages component exists on the page, leverage the MessagesModal component instead
- Ensure appropriate analytics are fired for the click event into the XO endpoint

❤️ Thank you!


src/zoid/buttons/component.jsx
```diff
@@ -681,6 +682,84 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+                        amount,
+                        offer: offerType,
+                        account: clientID,
+                        merchantId: merchantID,
+                      });
+                    },
+                  })
+                  .render("body");
+
+                state.isAwaitingShow = false;
```
danzhaaspaypal, line 712: this block is written for an edge case where a user has hovered, thus starting the modal download, and then clicked before the download has finished.  To do: make sure that `state` updates as I believe it does, and make sure click logging occurs in this case.

src/zoid/buttons/util.js
```diff
@@ -357,3 +358,45 @@ export function getButtonSize(
...
+    local:
+      "https://www.paypalobjects.com/upstream/bizcomponents/stage/modal.js",
+    test: "https://www.paypalobjects.com/upstream/bizcomponents/stage/modal.js",
+    stage:
+      "https://www.paypalobjects.com/upstream/bizcomponents/stage/modal.js",
+    sandbox:
+      "https://www.paypalobjects.com/upstream/bizcomponents/sandbox/modal.js",
+    production:
+      "https://www.paypalobjects.com/upstream/bizcomponents/js/modal.js",
+  };
```
Seavenly, line 374: Instead of hardcoding the URL string multiple times increasing the bundle size, could we just fill in the `/js/` / `/sandbox/` / `/stage/` piece as is appropriate based on the environment?

src/zoid/buttons/component.jsx
danzhaaspaypal: removed this for a better solution

src/zoid/buttons/component.jsx
```diff
@@ -681,6 +682,62 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+            getLogger()
+              .info("button_message_clicked")
+              .track({
+                [FPTI_KEY.EVENT_NAME]: "message_click",
+                // [FPTI_KEY.BUTTON_MESSAGE_OFFER_TYPE]: offerType,
+                // [FPTI_KEY.BUTTON_MESSAGE_TYPE]: messageType,
+                // [FPTI_KEY.BUTTON_MESSAGE_POSITION]: message.position,
+                // [FPTI_KEY.BUTTON_MESSAGE_ALIGN]: message.align,
+                // [FPTI_KEY.BUTTON_MESSAGE_COLOR]: message.color,
+                // [FPTI_KEY.AMOUNT]: amount,
```
danzhaaspaypal, line 734: @paypal-rosman I entered the FPTI_KEYs for you, and they are commented out until you get to add them to the sdk-constants

paypal-rosman: @danzhaaspaypal I we also want `MESSAGE_TYPE` here as well? @Seavenly thoughts ?

danzhaaspaypal: @paypal-rosman I think I've got you covered on line 730 there, unless you're requesting a name change?

paypal-rosman: my eyes are still not working lol thank you

src/zoid/buttons/util.js
```diff
@@ -389,10 +399,18 @@ export const getModal: () => Object = memoize(() => {
       return new Promise((resolve) => {
         const script = document.createElement("script");
         script.src = modalBundleUrl();
+        script.setAttribute("data-namespace", namespace);
```
Seavenly, line 402: I believe `modal.js` only uses `data-pp-namespace` so is this safe to remove?

src/zoid/buttons/util.js
```diff
@@ -389,10 +399,18 @@ export const getModal: () => Object = memoize(() => {
...
+        script.setAttribute("data-pp-namespace", namespace);
         document.body?.appendChild(script);
         script.addEventListener("load", () => {
```
Seavenly, line 405: I believe we can leverage an `"error"` event to capture failures from the script (e.g. network error) [according to MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/error_event) so we can use that to round out this promise based return value.

src/zoid/buttons/util.js
```diff
@@ -362,9 +363,18 @@ export function getButtonSize(
...
+  merchantID: $ReadOnlyArray<string> | void
+) => Object = memoize((clientID, merchantID) => {
+  const namespace = getNamespace();
+
+  if (window[namespace]?.MessagesModal) {
+    const modal = window[namespace].MessagesModal;
+    return modal({
+      account: `client-id:${clientID}`,
+      merchantId: merchantID?.join(",") || undefined,
+    });
```
Seavenly, line 377: Are we able to reorder this function to reduce some nesting and duplication?
``` ts
if (!window[namespace].MessagesModal)
    await new Promise(
        // Load the modal script
    )
}
// This now captures both scenarios and we know the modal exists at this point
const modal = window[namespace].MessagesModal;
return modal(...)
```
This helps keep a singular linear flow that doesn't have 2 spots where the modal can be created where we may forget to update one over the other in the future.

src/zoid/buttons/util.js
danzhaaspaypal: It seems that in practice data-namespace works and data-pp-namespace does not. data-pp-namespace fails to make a request to smart/modal

src/zoid/buttons/util.js
```diff
@@ -357,3 +362,59 @@ export function getButtonSize(
...
+        case "production":
+        default:
+          envPiece = "js";
+      }
+      return `https://www.paypalobjects.com/upstream/bizcomponents/${envPiece}/modal.js`;
+    };
+
+    try {
+      // eslint-disable-next-line no-restricted-globals
+      return new Promise(() => {
```
Seavenly, line 393: If we return the Promise then the modal creation code below will never execute since the function is memoized. We will need to `await` the promise instead so that it becomes just another step.

src/zoid/buttons/util.js
```diff
@@ -357,3 +362,59 @@ export function getButtonSize(
...
+        const script = document.createElement("script");
+        script.src = modalBundleUrl();
+        script.setAttribute("data-pp-namespace", namespace);
+        script.addEventListener("error", (err: Event) => {
+          throw err;
```
Seavenly, line 398: Since we're inside a promise we should be using `reject(err)` instead of throwing it.

src/zoid/buttons/util.js
```diff
@@ -357,3 +362,59 @@ export function getButtonSize(
...
+        });
+        document.body?.appendChild(script);
+        script.addEventListener("load", () => {
+          document.body?.removeChild(script);
```
Seavenly, line 402: We're no longer calling `resolve()` so this promise will never complete and just hang indefinitely.

src/zoid/buttons/util.js
```diff
@@ -357,3 +362,59 @@ export function getButtonSize(
...
+        });
+      });
+    } catch (err) {
```
Seavenly, line 405: For good measure we should consider encompassing the whole function within the try/catch.

src/zoid/buttons/util.js
danzhaaspaypal: this was discussed and fixed externally; for posterity, `data-pp-namespace` is the correct attribute.

src/zoid/buttons/util.js
```diff
@@ -357,3 +362,60 @@ export function getButtonSize(
...
   }
 }
+
+export const getModal: (
+  clientID: string,
+  merchantID: $ReadOnlyArray<string> | void
+) => Object = memoize(async (clientID, merchantID) => {
+  try {
+    const namespace = getNamespace();
+    if (!window[namespace]?.MessagesModal) {
```
Seavenly, line 372: I think we can be confident that `window[namespace]` will exist otherwise the SDK would error out while initializing. Are we able to remove the optional chaining to reduce doubt of its existence?

src/zoid/buttons/util.js
```diff
@@ -357,3 +362,60 @@ export function getButtonSize(
...
+        script.addEventListener("load", () => {
+          document.body?.removeChild(script);
+          resolve();
+        });
+        document.body?.appendChild(script);
+      });
+    }
+
+    const modal = window[namespace].MessagesModal;
+    return modal({
```
Seavenly, line 408: Since this variable isn't really being reused are we able to combine these to `window[namespace].MessagesModal({ ... })` since that the common form we're used to seeing when creating SDK components?

src/zoid/buttons/util.js
```diff
@@ -357,3 +360,60 @@ export function getButtonSize(
...
+        });
+        document.body?.appendChild(script);
+      });
+    }
+
+    return window[namespace].MessagesModal({
+      account: `client-id:${clientID}`,
+      merchantId: merchantID?.join(",") || undefined,
+    });
+  } catch (err) {
```
Seavenly, line 410: If there's an error scenario (e.g. the `modal.js` URL failed for some reason) we likely don't want that to be cached via the `memoize` wrapper. We'd like a retry to occur when the user clicks on the message again. It looks like the `memoize` util has a [reset capability](https://github.com/krakenjs/belter/blob/449034e63bca445708b62d9e08f7190a44ba4169/src/util.js#L295) we may be able to call to reset the memoized cache in the error scenario.

``` js
} catch (err) {
    getModal.reset();
    // ...
}
```

test/integration/windows/button/modal.js
```diff
@@ -0,0 +1,22 @@
+/* @flow */
+const namespace = document.currentScript?.getAttribute("data-pp-namespace");
+
+// function createMockFn(callback = () => {}) {
+//   function mockFn(...args) {
+//     mockFn.calls.push(args);
+//     callback(...args)
+//   }
+// }
```
Seavenly, line 9: Can this be removed?

src/zoid/buttons/util.js
```diff
@@ -357,3 +360,60 @@ export function getButtonSize(
...
+            break;
+          case "sandbox":
+            envPiece = "sandbox";
+            break;
+          case "production":
+          default:
+            envPiece = "js";
+        }
+        return `https://www.paypalobjects.com/upstream/bizcomponents/${envPiece}/modal.js`;
+      };
```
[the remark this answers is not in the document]
Seavenly: Ah yeah that's a good callout for the test case. All of the different SDK components have different `globals.js` that get merged together, so introducing another top level namespace could potentially cause a collision. We may want to stay inside `__PAYPAL_CHECKOUT__` since that's what's already reserved here. If we want to include the full URL though that may likely be fine. We can make that change and then see how the rest of the team feels.

src/zoid/buttons/util.js
danzhaaspaypal: VSCode forsees a problem with that code as identified by flow:
> Cannot call `getModal.reset` because property `reset` is missing in  statics of function type [1].Flow(prop-missing)

I've been searching in [PayPal's Github for examples of other implementations of that reset method](https://github.paypal.com/search?l=JavaScript&p=6&q=memoize+reset&type=Code), but haven't had any luck.  Closest I could find was [an implementation](https://github.paypal.com/Checkout/xo-buttonjs/blob/4495d4ae6d7aa28488379408f00d2dd5e54f357a/public/js/inlineGuest/creditCardForm.js) on a [handwritten memoize function](https://github.paypal.com/Checkout/xo-buttonjs/blob/4495d4ae6d7aa28488379408f00d2dd5e54f357a/public/js/button/util.js#L24).

src/ui/buttons/props.js
```diff
@@ -776,15 +776,18 @@ export function normalizeButtonMessage(
     }
   }
 
-  if (color && !values(MESSAGE_COLOR).includes(color)) {
+  if (typeof color !== "undefined" && !values(MESSAGE_COLOR).includes(color)) {
```
wsbrunson, line 779: what prompted the increased strictness on these if statements?

src/zoid/buttons/component.jsx
```diff
@@ -681,6 +682,51 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+      },
+
+      onMessageClick: {
+        type: "function",
+        required: false,
+        value: ({ props }) => {
+          return async ({ offerType, messageType }) => {
+            const { message, clientID, merchantID, currency, buttonSessionID } =
+              props;
+            const amount = message?.amount || undefined;
```
wsbrunson, line 704: i think if `message?.amount` didn't exist it would be `undefined` anyway. You either don't need this part `|| undefined` or we should talk about how to make it more explicit if you do.

src/zoid/buttons/component.jsx
```diff
@@ -681,6 +682,51 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+            getLogger()
+              .info("button_message_clicked")
+              .track({
+                [FPTI_KEY.EVENT_NAME]: "message_click",
+                // [FPTI_KEY.BUTTON_MESSAGE_OFFER_TYPE]: offerType,
+                "button message type fpti key placeholder": messageType,
+                // [FPTI_KEY.BUTTON_MESSAGE_POSITION]: message.position,
+                // [FPTI_KEY.BUTTON_MESSAGE_ALIGN]: message.align,
+                // [FPTI_KEY.BUTTON_MESSAGE_COLOR]: message.color,
+                // [FPTI_KEY.AMOUNT]: amount,
```
wsbrunson, line 717: do we still need the commented out keys?

src/zoid/buttons/component.jsx
```diff
@@ -681,6 +682,51 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+                [FPTI_KEY.BUTTON_SESSION_UID]: buttonSessionID,
+              });
+
+            return modalInstance.show({
+              amount,
+              offer: offerType?.join(",") || undefined,
```
wsbrunson, line 723: same comment here about `|| undefined`

src/zoid/buttons/util.js
```diff
@@ -357,3 +360,60 @@ export function getButtonSize(
...
 }
+
+export const getModal: (
+  clientID: string,
+  merchantID: $ReadOnlyArray<string> | void
+) => Object = memoize(async (clientID, merchantID) => {
+  try {
+    const namespace = getNamespace();
+    if (!window[namespace].MessagesModal) {
+      const modalBundleUrl = () => {
```
wsbrunson, line 371: can we move this up to a top-level function declaration, called something like `buildModalBundleUrl`

src/zoid/buttons/util.js
```diff
@@ -357,3 +360,60 @@ export function getButtonSize(
...
+      });
+    }
+
+    return window[namespace].MessagesModal({
+      account: `client-id:${clientID}`,
+      merchantId: merchantID?.join(",") || undefined,
+    });
+  } catch (err) {
+    getLogger()
+      .info("button_message_modal_fetch_error")
```
wsbrunson, line 412:
```suggestion
      .error("button_message_modal_fetch_error", {err})
```

src/zoid/buttons/util.js
```diff
@@ -357,3 +360,60 @@ export function getButtonSize(
...
+      .track({
+        err: err.message || "BUTTON_MESSAGE_MODAL_FETCH_ERROR",
+        details: err.details,
+        stack: JSON.stringify(err.stack || err),
```
wsbrunson, line 416: are these fpti keys that the credit team uses for errors?

danzhaaspaypal: No, these aren't credit team fpti keys; I took this catch block's getLogger statement from [an existing catch block in the button component](https://github.com/paypal/paypal-checkout-components/blob/main/src/zoid/buttons/component.jsx#L150-L156) that I thought might be representative of how the team wants their errors logged, and just renamed the error message accordingly.  I can handle it differently if you prefer.

src/zoid/buttons/component.jsx
danzhaaspaypal: Sorry, I jumped the gun a bit on hitting the 'ready for review' button after I finished the feature and tests but before merging the other PR.  These commented-out keys will be replaced very soon with updated, more precise keys from [a more dedicated tracking PR](https://github.com/paypal/paypal-checkout-components/pull/2360/files#diff-30a215bcfe3877b1f55af82ba98b729149bab885b3c5d2921d3bdc7e1c9ac799R748-R768).  Will let you know again when that PR is merged.

src/ui/buttons/props.js
danzhaaspaypal: we needed this normalization code block to run in the case of the merchant passing in an empty string

src/zoid/buttons/util.js
wsbrunson: i will always get out of the way of precedent 😂

src/zoid/buttons/component.jsx
```diff
@@ -681,6 +682,108 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+        required: false,
+        value: ({ props }) => {
+          // eslint-disable-next-line no-unused-vars
+          return ({
+            offerType,
+            messageType,
+            offerCountryCode,
+            creditProductIdentifier,
+          }) => {
+            // offerType, messageType, offerCountryCode, and creditProductIdentifier may be used in an upcoming message hover logging feature
```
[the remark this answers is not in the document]
danzhaaspaypal: Sure. I'll just leave the comment on line 741 then as an artifact of that info, and will remove the object being passed into the function.

src/zoid/buttons/component.jsx
```diff
@@ -681,6 +682,102 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+                button_message_type: messageType,
+                button_message_position: message?.position,
+                button_message_align: message?.align,
+                button_message_color: message?.color,
+                button_message_offer_country: offerCountryCode,
+                button_message_amount: amount,
+                [FPTI_KEY.BUTTON_SESSION_UID]: buttonSessionID,
+              });
+
+            const modalInstance = await getModal(clientID, merchantID);
```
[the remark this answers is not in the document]
danzhaaspaypal: what do we do if `getModal` is undefined, just not return anything?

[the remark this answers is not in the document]
Seavenly: Yeah. Something as simple as `return modalInstance?.show()` would work.

Seavenly (approved):

danzhaaspaypal:
@wsbrunson we're ready for you to take another look at this PR, at your convenience

wsbrunson (approved):

PR #2355, opened 2024-03-06
update axo loader library

wsbrunson:
Reverts the revert. This is re-adding the updated axo library


src/connect/component.jsx
```diff
@@ -1,5 +1,5 @@
 /* @flow */
-import { loadAxo } from "@paypal/connect-loader-component";
+import { loadAxo } from "@paypal/accelerated-checkout-loader/dist/loader.esm";
```
ravishekhar, line 2: Can we make the loader a top level export from this package? Bit worried that importing nested dependencies can create issues with future refactorings in this component.

ravishekhar (approved):
Left a minor comment, otherwise looks good.

PR #2356, opened 2024-03-07
ncps - expand inline guest

jshawl:

❤️ Thank you!

### DIFF
--- a/src/hosted-buttons/utils.js
+++ b/src/hosted-buttons/utils.js
@@ -1,14 +1,18 @@
 /* @flow */
 
-import { request, memoize } from "@krakenjs/belter/src";
+import { request, memoize, popup } from "@krakenjs/belter/src";
 import {
   buildDPoPHeaders,
   getSDKHost,
   getClientID,
   getMerchantID as getSDKMerchantID,
 } from "@paypal/sdk-client/src";
+import { FUNDING } from "@paypal/sdk-constants/src";
+
+import { DEFAULT_POPUP_SIZE } from "../zoid/checkout";
 
 import type {
+  BuildOpenPopup,
   ButtonVariables,
   CreateAccessToken,
   CreateOrder,
@@ -21,6 +25,7 @@ import type {
 const entryPoint = "SDK";
 const baseUrl = `https://${getSDKHost()}`;
 const apiUrl = baseUrl.replace("www", "api");
+export const popupFallbackClassName = "paypal-popup-fallback";
 
 const getHeaders = (accessToken?: string) => ({
   ...(accessToken && { Authorization: `Bearer ${accessToken}` }),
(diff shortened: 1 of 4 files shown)


src/hosted-buttons/utils.js
```diff
@@ -208,6 +209,14 @@ export const buildHostedButtonOnApprove = ({
         merchant_id: merchantId,
         context_id: data.orderID,
       }),
+    }).then((response) => {
+      // The "Debit or Credit Card" button does not open a popup
```
wsbrunson, line 213: It does sometimes open a popup though right? or are we saying in the ncps case it ALWAYS does the expand?

jshawl: it does sometimes open a popup - e.g. clicking on the debit or credit card button before second render. I tested this scenario and made sure a _second_ popup isn't opened though it's not obvious to me just yet why that isn't a bug 🤔

wsbrunson: does it redirect inside the popup instead of the parent page?

jshawl: no it closes the popup and then full page redirects from the merchant site to the thank you page

wsbrunson: oh wow, so this handles both cases. That's great!

jshawl: yes! though I think what's happening under the hood is that this `window.location = ` is forcing the popup to close in the `beforeUnload` event. Similar to:

1. Click on paypal button (popup opens)
2. open dev console and `window.location = 'https://example.com/'`
3. Observe popup close

I think this is an OK tradeoff but wanted to be transparent about why this "works". An alternative would be only invoking `window.location=` if the popup window wasn't opened because of a prerender click (not sure what's involved in that but I bet it will be challenging / complex)

wsbrunson: would there be enough time to close it through onApprove?

jshawl: Definitely! But I think the correct way would be to _not_ close the popup  - e.g. redirect from within the popup to the thank you page (identical to the PayPal button functionality)

jshawl: e.g. I don't think there's a way to know that `onApprove` was invoked from a prerender-click context: I'll timebox it early next week and see if it's a quick win but otherwise I think we should go with what's in this PR - wdyt? I'm def open to other ideas

wsbrunson: i'm down for the bare minimum that matches what product is asking for. Whatever that is lol but feel free to do as much exploration as you want

nikrom17 (approved):

spencersablan (approved):

PR #2361, opened 2024-03-19
WIP: New Feature disableMaxHeight

nikrom17:

❤️ Thank you!

### DIFF
--- a/src/ui/buttons/props.js
+++ b/src/ui/buttons/props.js
@@ -320,6 +320,7 @@ export type ButtonStyle = {|
   period?: number,
   height?: number,
   disableMaxWidth?: boolean,
+  disableMaxHeight?: boolean,
   borderRadius?: number,
 |};
 
@@ -332,6 +333,7 @@ export type ButtonStyleInputs = {|
   period?: number | void,
   height?: number | void,
   disableMaxWidth?: boolean | void,
+  disableMaxHeight?: boolean | void,
   borderRadius?: number | void,
 |};
 
@@ -624,6 +626,7 @@ export function normalizeButtonStyle(
     period,
     menuPlacement = MENU_PLACEMENT.BELOW,
     disableMaxWidth,
+    disableMaxHeight,
     borderRadius,
   } = style;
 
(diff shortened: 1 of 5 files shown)


test/integration/tests/button/validation.js
```diff
@@ -852,17 +852,27 @@ const buttonConfigs = [
...
+      },
+      {
+        disableMaxHeight: true,
+        height: 60,
+        valid: true,
+      },
 
       // $FlowFixMe
-    ].map(({ height, size, valid }) => ({
+    ].map(({ height, valid, disableMaxHeight }) => ({
```
nikrom17, line 867: It didn't seem like size was used at all in this sections so I removed it and used the sentence structure to indicate if the `disbaleMaxHeight` feature was enabled.

wsbrunson (approved):
looking great!

src/ui/buttons/props.js
```diff
@@ -694,13 +697,43 @@ export function normalizeButtonStyle(
...
 
+  if (disableMaxHeight !== undefined) {
+    if (typeof disableMaxHeight !== "boolean") {
+      throw new TypeError(
+        `Expected style.disableMaxHeight to be a boolean, got: ${disableMaxHeight}`
+      );
+    }
+
+    const disableMaxHeightInvalidFundingSources = [FUNDING.CARD, undefined];
+    const disableMaxHeightValidFundingSources = [
```
wsbrunson, line 721: why are these the only ones supported for disabling max height?

wsbrunson (approved):

src/ui/buttons/props.js
nikrom17: The only 2 buttons that aren't supported are the smart stack and card. The main reason these aren't supported was the challenge of getting them to render correctly and not being the specific use case that Stripe was looking for.

It seemed like they were rendering the standalone paypal button so we focused on enabling that only.

The card button is not supported because the inline guest experience gets rendered in the same container as the card button. If the merchant but a height limitation on the container for the card button, then it was also going to apply to the inline guest experience.

I know we have a bunch of other funding sources, but I believe we only have buttons for the funding sources below. The other funding sources are for the card fields, correct?

Screenshot 2024-05-21 at 9 24 12 AM

wsbrunson: We also have all the APMs although maybe we don't support them as standalone buttons? If so, I would think we would have a config value somewhere that disallows those buttons as standalone

https://github.com/paypal/paypal-sdk-constants/blob/main/src/funding.js#L3

spencersablan:
Thanks to `disableMaxHeight` & `disableMaxWidth`, I can finally fulfill my dream of making an app that is just a full screen paypal button. 

image

spencersablan (approved):
Nice bro!  🚀

src/ui/buttons/props.js
nikrom17: Looks like all those funding sources can be rendered as a standalone button. I've updated the error message to include all of those as well.

wsbrunson: I would consider the following tweaks, might be easier to maintain:

1. Early return if `fundingSource === undefined`. Error message could be something like `disableMaxHeight is only supported for standalone buttons. You can render a standalone button by passing `fundingSource` to the `Button` config: paypal.Buttons({fundingSource: paypal.FUNDING.PAYPAL})

2. Then, down below, throw the error for invalid buttons and just say "This funding source, {fundingSource}, does not support `disableMaxHeight`"

That's some very rough copy, feel free to change it if you decide to go in this direction

PR #2363, opened 2024-03-21
fix: supports empty message object

danzhaaspaypal:
### Description

Ensures that an empty message object input (e.g. `message: {}`) results in a message render by populating message props `align`, `position`, and `color` with computed default values.

❤️ Thank you!


src/ui/buttons/props.js
```diff
@@ -914,6 +925,29 @@ export function normalizeButtonProps(
...
+    components,
+    onShippingChange,
+    flow,
+    wallet,
+    applePaySupport,
+    supportsPopups,
+    supportedNativeBrowser,
+    experiment,
+    displayOnly,
+  });
```
Seavenly, line 946: Can we leverage the same `renderedButtons` prop instead of recalculating this?

src/ui/buttons/buttons.jsx
```diff
@@ -278,8 +272,8 @@ export function Buttons(props: ButtonsProps): ElementNode {
         fundingEligibility={fundingEligibility}
       />
 
-      {message && calculatedMessagePosition === MESSAGE_POSITION.TOP ? (
-        <Message markup={messageMarkup} position={calculatedMessagePosition} />
+      {message && message.position === MESSAGE_POSITION.TOP ? (
+        <Message markup={messageMarkup} position={message.position} />
```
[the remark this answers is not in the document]
danzhaaspaypal: good thought - the effect is that flow complains about the `position = {message.position}`:

> Cannot get `message.position` because property `position` is missing in  undefined [1].Flow(incompatible-use)

src/ui/buttons/props.js
danzhaaspaypal: just tried it - it appears to work, but flow notes 
> property `renderedButtons` is missing in  `ButtonPropsInputs` [1].Flow(prop-missing)

what would be the courteous way to handle this? Add `renderedButtons` to the type?

src/ui/buttons/buttons.jsx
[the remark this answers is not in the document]
Seavenly: Oh right. Thanks for the reminder hah

src/ui/buttons/props.js
Seavenly: Yeah that sounds appropriate

danzhaaspaypal:
closed in favor of the rebased PR [here](https://github.com/paypal/paypal-checkout-components/pull/2369)

PR #2367, opened 2024-03-27
feat: dynamic messaging - spelling fix add merchant id add currency

paypal-rosman:
### Description

add currency, fix spelling mistake, add merchantID (seller id in sdk constants)

❤️ Thank you!


src/zoid/buttons/component.jsx
```diff
@@ -714,6 +714,9 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
                 button_message_color: message?.color,
                 button_message_offer_country: offerCountryCode,
                 button_message_amount: amount,
+                button_message_currency: currency,
+                // comes back as an array
+                [FPTI_KEY.SELLER_ID]: merchantID[0],
```
ravishekhar, line 719: There can be more then one payee in a transaction. Do you want to log only one payee in fpti?

src/zoid/buttons/component.jsx
```diff
@@ -766,11 +769,14 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
                 button_message_credit_product_identifier:
                   creditProductIdentifier,
                 button_message_type: messageType,
-                button_message_posiiton: message?.position,
+                button_message_position: message?.position,
```
ravishekhar, line 772: good catch.

src/zoid/buttons/component.jsx
paypal-rosman: good question, I am not sure if we need both. Is there an order of how this array is generated? I found a couple examples in scnw only logging the first one so i assumed that might be the more "important" one. Are you able to clarify that?

PR #2370, opened 2024-04-02
feat: update message object to accept array offer param and  send offer as string

perco12:
### Description

Ensures that the message object `offer` input remains an array but output results in message having `offer` property as a comma-delimited string

❤️ Thank you!


src/ui/buttons/props.js
```diff
@@ -762,18 +761,22 @@ export function normalizeButtonMessage(
...
+      }
+      offer = [offer];
     }
     const invalidOffers = offer.filter(
       (o) => !values(MESSAGE_OFFER).includes(o)
     );
     if (invalidOffers.length > 0) {
       throw new Error(`Invalid offer(s): ${invalidOffers.join(",")}`);
     }
+    offer = offer.join(",");
```
ravishekhar, line 779: Does this change breaks any existing clients of this method?

perco12: yes, clients that are expecting offer as an array should now expect it as a joined string. They will also have to be updated to receive a string.

src/ui/buttons/props.js
```diff
@@ -762,18 +761,22 @@ export function normalizeButtonMessage(
 
   if (typeof offer !== "undefined") {
     if (!Array.isArray(offer)) {
```
danzhaaspaypal, line 763: this type check is now unnecessary thanks to the new type check below

test/integration/tests/button/validation.js
```diff
@@ -948,18 +948,6 @@ const buttonConfigs = [
...
-
-      {
-        message: {
-          amount: 100,
-          offer: "pay_later_long_term", // invalid: should be in an array
-          color: "black",
-          position: "top",
-          align: "left",
-        },
-        valid: false,
```
danzhaaspaypal, line 960: Alternately, we could change `valid` to `true` here

src/ui/buttons/props.js
```diff
@@ -430,7 +430,7 @@ export type ApplePaySessionConfigRequest = (
 
 export type ButtonMessage = {|
   amount?: number,
-  offer?: $ReadOnlyArray<$Values<typeof MESSAGE_OFFER>>,
+  offer?: $Values<typeof MESSAGE_OFFER> | string,
```
danzhaaspaypal, line 433:
```suggestion
  offer?: $Values<typeof MESSAGE_OFFER>
```

The string type is implicit by the enum values of `MESSAGE_OFFER`

danzhaaspaypal (approved):

src/ui/buttons/props.js
Seavenly, line 433: Since the user input will be an array, should this be
``` js
offer?: $ReadOnlyArray<$Values<typeof MESSAGE_OFFER>> | string,
```

src/ui/buttons/props.js
```diff
@@ -762,18 +761,22 @@ export function normalizeButtonMessage(
...
-        )}`
-      );
+      if (typeof offer !== "string") {
+        throw new TypeError(
+          `Expected message.offer to be an array of strings, got: ${String(
+            offer
+          )}`
+        );
+      }
+      offer = [offer];
```
Seavenly, line 771: If the input to this function on its second call is a string of multiple offers (e.g. `"pay_later_long_term,pay_later_short_term"`), is this then going to result in one large single string entry `["pay_later_long_term,pay_later_short_term"]` which will then cause the invalid offer error to be thrown?

test/integration/tests/button/validation.js
```diff
@@ -952,12 +952,12 @@ const buttonConfigs = [
       {
         message: {
           amount: 100,
-          offer: "pay_later_long_term", // invalid: should be in an array
+          offer: "pay_later_long_term",
```
Seavenly, line 955: We should include a test that includes multiple offers in string form

src/ui/buttons/props.js
perco12: It will throw an invalid offer error before it even run the offer check as the offer input looks for `$ReadOnlyArray<$Values<typeof MESSAGE_OFFER>>` which can only be `pay_later_long_term` or `pay_later_short_term`, no multi offer. If we did want multiple offer, then we will have to add those cases to the MESSAGE_OFFER enum.

Seavenly: `$ReadOnlyArray<$Values<typeof MESSAGE_OFFER>>` is inclusive of an input of `["pay_later_short_term", "pay_later_long_term"]`. If I pass that in I get the following error:
```
Uncaught Error: Error: Invalid offer(s): pay_later_short_term,pay_later_long_term
```

src/ui/buttons/props.js
perco12: that is output, offer in `ButtonMessageinput` is the array

src/ui/buttons/props.js
perco12: I just realized the actual error and it has been fixed!!!

perco12:
closing this PR in favor of as this is showing 22 old commit into main. NEW PR is #2392

PR #2371, opened 2024-04-04
prevent popup bridge xss

jshawl:
### Description

This PR will `reject` if `actions.redirect` attempts to redirect to a javascript URI.

LI-38558 - prevent xss

Specific details are in the ticket, but I was able to test this locally by using a redirect extension to redirect requests from https://www.paypalobjects.com/api/checkout.js to the local built js https://127.0.0.1:8080/dist/checkout.js

Screenshot 2024-04-05 at 07 08 45

❤️ Thank you!

### DIFF
--- a/src/lib/dom.js
+++ b/src/lib/dom.js
@@ -237,11 +237,14 @@ export function extendUrl(url : string, params : { [key : string] : string } = {
 }
 
 export function redirect(win : CrossDomainWindowType = window, url : string) : ZalgoPromise<void> {
-    return new ZalgoPromise(resolve => {
+    return new ZalgoPromise((resolve, reject) => {
 
         info(`redirect`, { url });
 
         setTimeout(() => {
+            if (url.toLowerCase().replace(/[^a-z:]+/g, '').match(/^javascript:/)) {
+                return reject(new Error('Invalid redirect URL'));
+            }
             win.location = url;
             if (!urlWillRedirectPage(url)) {
                 resolve();
--- a/src/lib/dom.js
+++ b/src/lib/dom.js
@@ -241,10 +241,11 @@ export function redirect(win : CrossDomainWindowType = window, url : string) : Z
 
         info(`redirect`, { url });
 
+        if (url.toLowerCase().replace(/[^a-z:]+/g, '').match(/^javascript:/)) {
+            return reject(new Error('Invalid redirect URL'));
+        }
+
         setTimeout(() => {
-            if (url.toLowerCase().replace(/[^a-z:]+/g, '').match(/^javascript:/)) {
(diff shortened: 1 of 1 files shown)


src/lib/dom.js
```diff
@@ -237,11 +237,14 @@ export function extendUrl(url : string, params : { [key : string] : string } = {
...
 export function redirect(win : CrossDomainWindowType = window, url : string) : ZalgoPromise<void> {
-    return new ZalgoPromise(resolve => {
+    return new ZalgoPromise((resolve, reject) => {
 
         info(`redirect`, { url });
 
         setTimeout(() => {
+            if (url.toLowerCase().replace(/[^a-z:]+/g, '').match(/^javascript:/)) {
+                return reject(new Error('Invalid redirect URL'));
+            }
```
jshawl, line 247: I spent an unreasonable amount of time trying to write a test for this, but left it out because:

- all of the tests have a custom `setupPopupBridge` implementation: https://github.com/paypal/paypal-checkout-components/blob/6cb970b6edbb5ac091e46444885ed5f1c99006ab/test/tests/button/popupBridge.js#L19
- that hardcode the `return_uri`: https://github.com/paypal/paypal-checkout-components/blob/6cb970b6edbb5ac091e46444885ed5f1c99006ab/test/tests/common.js#L623
- and depend on a `hashchange` event for assertions: https://github.com/paypal/paypal-checkout-components/blob/6cb970b6edbb5ac091e46444885ed5f1c99006ab/test/tests/button/popupBridge.js#L119-L121

wsbrunson (approved):

gregjopa (approved):

src/lib/dom.js
```diff
@@ -237,11 +237,14 @@ export function extendUrl(url : string, params : { [key : string] : string } = {
 }
 
 export function redirect(win : CrossDomainWindowType = window, url : string) : ZalgoPromise<void> {
-    return new ZalgoPromise(resolve => {
+    return new ZalgoPromise((resolve, reject) => {
 
         info(`redirect`, { url });
 
         setTimeout(() => {
+            if (url.toLowerCase().replace(/[^a-z:]+/g, '').match(/^javascript:/)) {
```
ravishekhar, line 245: Just a nit pick, can we move this logic outside of this setTimeout?

ravishekhar (approved):

src/lib/dom.js
jshawl: I also re-tested this and it worked correctly! https://github.com/paypal/paypal-checkout-components/pull/2371/commits/5f7433272bb44aa9246516d0c11e525e92d4b69b

PR #2375, opened 2024-04-17
Use Standalone buttons for V2 of Hosted Buttons

nikrom17:
### Description

Instead of using the smart stack, we want to render standalone buttons. We will introduce a new prop to the hosted buttons `fundingSources` and this will render the v2 experience.  

https://paypal.atlassian.net/browse/DTPPCPSDK-2109?atlOrigin=eyJpIjoiODQzYzczYWM2MzQ1NGQ5NjgyYjVlMGVlZWZlNjE4MGQiLCJwIjoiaiJ9

V1 experience
Screenshot 2024-04-17 at 9 01 02 AM

V2 Experience
Screenshot 2024-04-17 at 9 01 20 AM

❤️ Thank you!


src/hosted-buttons/index.js
```diff
@@ -19,6 +23,7 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
   function HostedButtons({
     enableDPoP = false,
     hostedButtonId,
+    fundingSources = [],
```
nikrom17, line 26: After our latest meeting with NCP team, they will provide a new variable `version` in our function call `getHostedButtonDetails`. I left this feature flag in for now since that is what was outlined in the ticket when I started.

I can update it to match what we will expect from the above response or we can update it once we actually have the updated response.

src/hosted-buttons/index.js
```diff
@@ -34,24 +39,59 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
...
-        }),
-        onApprove: buildHostedButtonOnApprove({
-          enableDPoP,
+      if (shouldRenderSDKButtons(fundingSources)) {
+        const { flexDirection } = getFlexDirection({ ...style });
+
+        buildButtonContainer({ flexDirection, selector });
+
+        // Only render 2 buttons max
+        fundingSources?.slice(0, 2).forEach((fundingSource, index) => {
```
nikrom17, line 48: This will need to be refactored once we get the updated response from the NCP team.

src/hosted-buttons/types.js
```diff
@@ -1,8 +1,27 @@
...
+}
+
+export interface BuildButtonContainerArgs {
+  flexDirection: FlexDirection;
+  selector: string | HTMLElement;
+}
+
 export type HostedButtonsComponentProps = {|
   hostedButtonId: string,
+  fundingSources?: $ReadOnlyArray<FundingSources>,
```
wsbrunson, line 24: Since we default this to `[]` should it be not optional?
```suggestion
  fundingSources: $ReadOnlyArray<FundingSources>,
```

src/hosted-buttons/types.js
```diff
@@ -1,8 +1,27 @@
 /* @flow */
 /* eslint-disable no-restricted-globals, promise/no-native */
 
+export type FlexDirection = string;
```
nikrom17, line 4: I would love to update all the strings types to their actual string values, but I'm thinking that could be a separate ticket once we have the updated response from NCP team

src/hosted-buttons/index.js
```diff
@@ -34,24 +39,59 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
...
+              enableDPoP,
+              hostedButtonId,
+              merchantId,
+            }),
+            onApprove: buildHostedButtonOnApprove({
+              enableDPoP,
+              hostedButtonId,
+              merchantId,
+            }),
+          }).render(
```
wsbrunson, line 69: we'd still want to check `Button.isEligible()` before rendering. Venmo for instance has a lot of client-only eligibility checks

src/hosted-buttons/utils.js
```diff
@@ -220,3 +225,86 @@ export const buildHostedButtonOnApprove = ({
...
+      paypal: "silver",
+      venmo: "blue",
+      paylater: "silver",
+    },
+  };
+
+  return colorMap[color][fundingSource];
+}
+
+export function buildButtonContainer({
```
nikrom17, line 276: If we get the html structure from NCP with a div to add the buttons to, this function will need to be refactored.

src/hosted-buttons/utils.js
```diff
@@ -220,3 +225,86 @@ export const buildHostedButtonOnApprove = ({
...
+  flexDirection,
+  selector,
+}: BuildButtonContainerArgs) {
+  if (typeof selector !== "string") {
```
nikrom17, line 280: I started looking at how Zoid was handling accepting a sting and an html element, but I didn't find anything useful here. Since NCP team is generating the piece of code that contains the selector and it is currently a string referring to an id, I thought it would be simpler to only accept that for the hosted button experience.

src/hosted-buttons/utils.js
```diff
@@ -30,7 +35,7 @@ const getHeaders = (accessToken?: string) => ({
 });
 
 export const getMerchantID = (): string | void => {
-  // The SDK supports mutiple merchant IDs, but hosted buttons only
+  // The SDK supports multiple merchant IDs, but hosted buttons only
   // have one merchant id as a query parameter to the SDK script.
   // https://github.com/paypal/paypal-sdk-client/blob/c58e35f8f7adbab76523eb25b9c10543449d2d29/src/script.js#L144
   const merchantIds = getSDKMerchantID();
```
wsbrunson, line 41: we should update this and the error message, this is a feature called Multi-Seller Payments (MSP) https://developer.paypal.com/docs/multiparty/checkout/multiseller-payments/

src/hosted-buttons/utils.js
```diff
@@ -220,3 +225,86 @@ export const buildHostedButtonOnApprove = ({
...
+export function getFlexDirection({
+  layout,
+}: GetFlexDirectionArgs): GetFlexDirection {
+  return { flexDirection: layout === "horizontal" ? "row" : "column" };
+}
+
+export function shouldRenderSDKButtons(
+  fundingSources: $ReadOnlyArray<FundingSources>
+): boolean {
+  return Boolean(fundingSources?.length);
```
wsbrunson:
```suggestion
  return Boolean(fundingSources.length);
```

src/hosted-buttons/index.js
```diff
@@ -34,24 +39,59 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
...
-          enableDPoP,
-          hostedButtonId,
-          merchantId,
-        }),
-        onApprove: buildHostedButtonOnApprove({
-          enableDPoP,
+      if (shouldRenderSDKButtons(fundingSources)) {
+        const { flexDirection } = getFlexDirection({ ...style });
+
+        buildButtonContainer({ flexDirection, selector });
```
jshawl, line 45: it took me a second to understand that `buildButtonContainer` was appending the newly created elements to the DOM - maybe this could be named `appendButtonContainer`? Totally optional suggestion I'm not that great at naming things

src/hosted-buttons/index.js
```diff
@@ -34,24 +39,59 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
...
+              ...style,
+              color: getButtonColor(style.color, fundingSource),
+            },
+            onInit,
+            onClick,
+            createOrder: buildHostedButtonCreateOrder({
+              enableDPoP,
+              hostedButtonId,
+              merchantId,
+            }),
```
jshawl, line 63: option to move these functions above, like:

```js
const createOrder = buildHostedButtonCreateOrder({
    enableDPoP,
    hostedButtonId,
    merchantId,
});
```

and then here it can be:

```js
Buttons({
    ...,
    onInit,
    onClick,
    createOrder,
    onApprove,
})
```

src/hosted-buttons/index.test.js
```diff
@@ -79,7 +79,36 @@ describe("HostedButtons", () => {
...
+      hostedButtonId: "B1234567890",
+      fundingSources: ["paypal", "venmo"],
+    }).render("#example");
+    expect(Buttons).toHaveBeenCalledWith(
+      expect.objectContaining({
+        hostedButtonId: "B1234567890",
+      })
+    );
+    expect(Buttons).toHaveBeenCalledTimes(2);
+    expect.assertions(2);
```
jshawl, line 111: awesome test coverage!

src/hosted-buttons/utils.js
```diff
@@ -220,3 +225,84 @@ export const buildHostedButtonOnApprove = ({
...
+  const secondaryButton = document.createElement("div");
+  secondaryButton.setAttribute("id", `ncp-secondary-button`);
+  secondaryButton.setAttribute("style", "flex-grow: 1");
+
+  buttonContainer.appendChild(primaryButton);
+  buttonContainer.appendChild(secondaryButton);
+
+  const ncpButtonContainer = window.document.getElementById(selector.slice(1));
+  ncpButtonContainer.appendChild(buttonContainer);
+}
```
jshawl, line 308: I think this code might be a good fit for a new `styles.js` file - totally up to you / I trust your judgment here

src/hosted-buttons/utils.test.js
```diff
@@ -288,4 +292,81 @@ describe("buildHostedButtonOnApprove", () => {
...
+        colorMap[color][fundingSource]
+      );
+    });
+  });
+});
+
+test("shouldRenderSDKButtons", () => {
+  expect(shouldRenderSDKButtons([])).toBe(false);
+  expect(shouldRenderSDKButtons(["paypal"])).toBe(true);
+  expect(shouldRenderSDKButtons(["paypal", "venmo"])).toBe(true);
```
jshawl, line 347: optional naming suggestion: `shouldRenderStandaloneButtons`

src/hosted-buttons/utils.test.js
```diff
@@ -288,4 +292,81 @@ describe("buildHostedButtonOnApprove", () => {
...
+  expect(mock).toHaveBeenCalledTimes(1);
+  expect(mock).toHaveBeenCalledWith(containerId);
+
+  expect(() =>
+    buildButtonContainer({ flexDirection: "row", selector: `.${containerId}` })
+  ).toThrow("Selector must be referring to an id");
+  expect(() =>
+    buildButtonContainer({ flexDirection: "row", selector })
+  ).toThrow("Selector must be a string");
+});
```
jshawl, line 370: awesome test coverage!

src/hosted-buttons/utils.js
```diff
@@ -220,3 +225,86 @@ export const buildHostedButtonOnApprove = ({
...
+  flexDirection,
+  selector,
+}: BuildButtonContainerArgs) {
+  if (typeof selector !== "string") {
+    throw new TypeError("Selector must be a string");
+  }
+
+  if (selector.charAt(0) !== "#") {
+    throw new TypeError("Selector must be referring to an id");
+  }
```
wsbrunson, line 286: we should consider doing something like this index.js:
```
const parsedSelector = parseSelector(selector)
...
func parsedSelector (selector: ?string) {
  ...validation checks on selector
}
```

That way we return very early if we weren't given a proper selector and any code that gets the `parsedSelector` knows its good

src/hosted-buttons/utils.js
```diff
@@ -220,3 +225,86 @@ export const buildHostedButtonOnApprove = ({
...
+  primaryButton.setAttribute("style", "flex-grow: 1");
+
+  const secondaryButton = document.createElement("div");
+  secondaryButton.setAttribute("id", `ncp-secondary-button`);
+  secondaryButton.setAttribute("style", "flex-grow: 1");
+
+  buttonContainer.appendChild(primaryButton);
+  buttonContainer.appendChild(secondaryButton);
+
+  // is this guaranteed to be an id?
```
wsbrunson: possibly delete

src/hosted-buttons/utils.js
```diff
@@ -220,3 +225,86 @@ export const buildHostedButtonOnApprove = ({
...
+
+  const buttonContainer = document.createElement("div");
+
+  buttonContainer.setAttribute(
+    "style",
+    `display: flex; flex-wrap: nowrap; gap: 16px; flex-direction: ${flexDirection}`
+  );
+
+  const primaryButton = document.createElement("div");
+  primaryButton.setAttribute("id", `ncp-primary-button`);
```
wsbrunson, line 296: are we using these ids anywhere else?

jshawl (commented):
I would like to pull this down but wanted to get an initial review back to you. So far looks great! nice job!

src/hosted-buttons/utils.js
jshawl: I think this is actually a limitation of the backend - it only accepts one merchant ID in the request payload

src/hosted-buttons/index.js
nikrom17: `appendButtonContainer` works for me and will make even more sense if NCP teams is giving us an id to append to

src/hosted-buttons/index.js
jshawl: edit: this could go one level higher too, like:

```js
const buttonOptions = {
  onInit,
  onClick,
  createOrder,
  onApprove
}
```

and then

```js
Buttons(buttonOptions)
```

totally optional though!

src/hosted-buttons/index.js
nikrom17: Updated in this [commit](https://github.com/paypal/paypal-checkout-components/pull/2375/commits/ac9f189a7d1d6e145742691a1707958bd7908fa1)

src/hosted-buttons/types.js
nikrom17: I originally had it as an optional prop but added the default empty array to avoid some type errors. I will update.

src/hosted-buttons/index.js
```diff
@@ -34,24 +41,60 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
...
+          } else {
+            getLogger().error(`${fundingSource} is not eligible.`);
+          }
+        });
+      } else {
+        // V1 Experience
+        // $FlowFixMe
+        Buttons({
           hostedButtonId,
-          merchantId,
```
jshawl, line 51: I think `merchantId` was accidentally removed. And I think it needs to be in the standalone buttons, too for analytics purposes (would be good to double-check why it's needed there, if at all!)

src/hosted-buttons/utils.js
nikrom17: I totally forgot about `document.querySelector`. I will update function to support.

Changes can be found in this [commit](https://github.com/paypal/paypal-checkout-components/pull/2375/commits/ba8f4f64569a3b4a2276d08c108d51660e7eddab)

src/hosted-buttons/utils.js
```diff
@@ -30,7 +35,7 @@ const getHeaders = (accessToken?: string) => ({
 });
 
 export const getMerchantID = (): string | void => {
-  // The SDK supports mutiple merchant IDs, but hosted buttons only
+  // The SDK supports multiple merchant IDs, but hosted buttons only
```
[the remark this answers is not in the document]
nikrom17: This is my favorite VS code extension: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

src/hosted-buttons/utils.test.js
nikrom17: I called it `shouldRenderSDKButtons` thinking we could reuse it and update the logic when we get the `checkoutButtonType` variable from the updated response.

We wont want to run any of the code in that if statement when `checkoutButtonType=single`

Screenshot 2024-04-17 at 11 44 45 AM

src/hosted-buttons/utils.js
nikrom17: I refactored this function to accept html elements. I've created a new check to see if the element was found and throw and error if not.

Let me know if you think it still should be its own function

https://github.com/paypal/paypal-checkout-components/pull/2375/files#diff-dc0ab5efddb0f4c5dc8c699ccafce1daa87906e1b839480c5be57048cee5482bR285

src/hosted-buttons/utils.js
nikrom17: No , just here. This will also get refactored since NCP team will:

1. provide us with an div#id in their html response to add the buttons to
2. they won't be providing a `fundingSoruces` array. This container building logic will be refactored once we get the updated response from them.

src/hosted-buttons/index.js
nikrom17: sounds good. I added it back to `buttonOptions`

src/hosted-buttons/utils.js
```diff
@@ -225,3 +230,86 @@ export const buildHostedButtonOnApprove = ({
...
+): boolean {
+  return Boolean(fundingSources.length);
+}
+
+export function appendButtonContainer({
+  flexDirection,
+  selector,
+}: BuildButtonContainerArgs) {
+  const elm =
+    typeof selector === "string" ? document.querySelector(selector) : selector;
```
spencersablan, line 286: I noticed this becoming a pattern for getting the selector.  Can we maybe move it into its own function like `getSelector()`? 😄

src/hosted-buttons/utils.js
```diff
@@ -225,3 +230,86 @@ export const buildHostedButtonOnApprove = ({
...
+
+  if (!elm) {
+    throw new Error("PayPal button container selector was not found");
+  }
+
+  // if (selector.charAt(0) !== "#") {
+  //   throw new TypeError("Selector must be referring to an id");
+  // }
```
spencersablan, line 294: Noticed this commented code.. Do we need this validation?

src/hosted-buttons/index.js
```diff
@@ -34,24 +42,58 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
...
+      if (shouldRenderSDKButtons(fundingSources)) {
+        const { flexDirection } = getFlexDirection({ ...style });
+
+        appendButtonContainer({ flexDirection, selector });
+
+        // Only render 2 buttons max
+        fundingSources.slice(0, 2).forEach((fundingSource, index) => {
+          // $FlowFixMe
+          const standaloneButton = Buttons({
+            ...buttonOptions,
```
jshawl, line 75: I think `fundingSource` needs to be in here so that the standalone button(s) render

src/hosted-buttons/utils.js
spencersablan: ⬆️  I like that idea.

src/hosted-buttons/utils.js
```diff
@@ -225,3 +230,86 @@ export const buildHostedButtonOnApprove = ({
...
+    `display: flex; flex-wrap: nowrap; gap: 16px; flex-direction: ${flexDirection}`
+  );
+
+  const primaryButton = document.createElement("div");
+  primaryButton.setAttribute("id", `ncp-primary-button`);
+  primaryButton.setAttribute("style", "flex-grow: 1");
+
+  const secondaryButton = document.createElement("div");
+  secondaryButton.setAttribute("id", `ncp-secondary-button`);
+  secondaryButton.setAttribute("style", "flex-grow: 1");
```
spencersablan, line 309: Just for my learning... is there any benefit of using `.setAttribute()` rather than just using `secondaryButton.id` or `secondaryButton.style`? 😄.  No need to change, I'm just curious.

src/hosted-buttons/utils.js
```diff
@@ -225,3 +230,86 @@ export const buildHostedButtonOnApprove = ({
...
+
+  // if (selector.charAt(0) !== "#") {
+  //   throw new TypeError("Selector must be referring to an id");
+  // }
+
+  const buttonContainer = document.createElement("div");
+
+  buttonContainer.setAttribute(
+    "style",
+    `display: flex; flex-wrap: nowrap; gap: 16px; flex-direction: ${flexDirection}`
```
jshawl, line 300: I noticed that the horizontal layout will take up 100% of the parent container's width but the v1 buttons had a max-width set:

v1:


Screenshot 2024-04-17 at 12 37 43

v2: 
Screenshot 2024-04-17 at 12 33 28

I think one quick win could be adding `max-width: 750px` here.

There will be a small difference from v1 because the horizontal buttons will hide the second button if there is not enough room (< 300px) but I could see handling that in the future (if ever)

src/hosted-buttons/index.js
```diff
@@ -19,12 +25,14 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
...
     enableDPoP = false,
     hostedButtonId,
+    fundingSources = [],
   }: HostedButtonsComponentProps): HostedButtonsInstance {
     const Buttons = getButtonsComponent();
     const render = async (selector) => {
       const merchantId = getMerchantID();
       const { html, htmlScript, style } = await getHostedButtonDetails({
         hostedButtonId,
+        fundingSources,
```
spencersablan, line 35: I see you're passing `fundingSources` to `getHostedButtonDetails()`, but `getHostedButtonDetails()` doesn't seem to be doing anything with it. 🤔  https://github.com/paypal/paypal-checkout-components/pull/2375/files#diff-dc0ab5efddb0f4c5dc8c699ccafce1daa87906e1b839480c5be57048cee5482bR82-R84

src/hosted-buttons/utils.js
wsbrunson: oh yeah, I agree, I mean we should update our comment and our error message to call out that the MSP feature is unsupported

src/hosted-buttons/index.js
```diff
@@ -34,24 +42,58 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
...
+
+          if (standaloneButton.isEligible()) {
+            standaloneButton.render(
+              index === 0 ? "#ncp-primary-button" : "#ncp-secondary-button"
+            );
+          } else {
+            getLogger().error(`${fundingSource} is not eligible.`);
+          }
+        });
+      } else {
```
spencersablan, line 90: I may be misunderstanding this logic, but should we check the eligible buttons before we slice the first two buttons?  

For example, if they pass `fundingSources: ["paypal", "venmo"]`, and Venmo is not eligible for a buyer, we would want to show PayPal and PayLater (like we do for the smart stack).  

I was thinking we'd use [`determineEligibleFunding()`](https://github.com/paypal/paypal-checkout-components/blob/main/src/funding/funding.js#L186) like we do for the smart stack but if this logic falls back correctly, I'm good with it 😄

src/hosted-buttons/utils.js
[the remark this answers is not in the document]
spencersablan: Woah I need that

spencersablan (commented):
A few comments/questions! 🚀

src/hosted-buttons/index.js
nikrom17: Flow was complaining about it being missing when I removed it as an option prop from `HostedButtonsComponentProps`.

src/hosted-buttons/utils.js
nikrom17: Not that I'm away of. I'm not a vanilla JS expert. I saw `setAttribute` somewhere and just ran with it.

src/hosted-buttons/utils.js
nikrom17: Nope, I forgot to remove this when I refactored to support html elements as a selector

src/hosted-buttons/utils.js
nikrom17: I will definitely add the max-with property. Hopefully NCP team can include these styles on html the provide us in the updated response 🤞

src/hosted-buttons/utils.js
nikrom17: sounds good, I've updated the comment block here

src/hosted-buttons/index.js
nikrom17: All of this logic will need to be refactored when we get the updated response from NCP team.

They won't be passing us the `fundingSources` array so we will need entirely new logic on how to handle rendering the second button.

I think we will be better able to update this logic once we have the updated response from NCP team

spencersablan: Ahh, sounds good 👍  Thanks!

src/hosted-buttons/utils.js
```diff
@@ -225,3 +238,81 @@ export const buildHostedButtonOnApprove = ({
...
+
+  if (!elm) {
+    throw new Error("PayPal button container selector was not found");
+  }
+
+  const buttonContainer = document.createElement("div");
+
+  buttonContainer.setAttribute(
+    "style",
+    `display: flex; flex-wrap: nowrap; gap: 16px; max-width: 750px; flex-direction: ${flexDirection}`
```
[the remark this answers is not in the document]
jshawl: but in vertical mode, 16px looks right to me as-is

src/hosted-buttons/utils.test.js
```diff
@@ -309,4 +315,99 @@ describe("buildHostedButtonOnApprove", () => {
...
+  });
+  expect(getFlexDirection({ layout: "vertical" })).toStrictEqual({
+    flexDirection: "column",
+  });
+});
+
+test("getButtonColor", () => {
+  const colors = ["gold", "blue", "silver", "white", "black"];
+  const fundingSources = ["paypal", "venmo", "paylater"];
+  const colorMap = {
```
jshawl, line 330: option to export this object from `utils.js` and then reuse it here

jshawl (approved):
Looks great! Nice job on this one 🚀 I left a few non-blocking comments

src/hosted-buttons/utils.js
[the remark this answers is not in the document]
nikrom17: I got the gap of 16px from the NCP figma files. We can let them know the gap defaults to 8px in our smart stack and see if they want to update their designs.

Screenshot 2024-04-17 at 3 23 37 PM

src/hosted-buttons/utils.test.js
nikrom17: I thought about reusing that color map, but I was concerned about someone accidentally changing a value. If the tests were using the same object, then they would automatically pass.

Happy to still do make the change, but that was my thought process.

jshawl: clever!

src/hosted-buttons/index.js
```diff
@@ -34,24 +42,59 @@ export const getHostedButtonsComponent = (): HostedButtonsComponent => {
...
+        style,
+      };
+
+      if (shouldRenderSDKButtons(fundingSources)) {
+        const { flexDirection } = getFlexDirection({ ...style });
+
+        appendButtonContainer({ flexDirection, selector });
+
+        // Only render 2 buttons max
+        fundingSources.slice(0, 2).forEach((fundingSource, index) => {
```
spencersablan, line 72: You don't have to do this, but imo it might be helpful to leave a comment above this to say it needs to be refactored.  I'm just scared they'll start passing funding sources in and it won't work as expected 😅  

But up to you! If you feel like its unnecessary, I'm down for that as well. 🚀

PR #2376, opened 2024-04-17
feat: adds button message popup handling for purchase protection

danzhaaspaypal:
### Description

Adds feature to have button messages open a popup (or new tab, depending on settings) on click instead of always a modal.

Supporting Purchase Protection messages, which will open a popup in cases where showing credit product presentment is inappropriate.

[Test on this demo page](https://young-anchorage-87182.herokuapp.com/_personal/nate/stage/buttons-dynamic-messaging.html?stage=te-dzh-purchase-protection.qa.paypal.com&client-id=B_Aksh3K8UpGwAs_Ngi0ahNHJja2bsxqtuqmhRAm944xzM-p6Qkq9JbqFc2yTz2Fg5WYtcW1QEu_tHDVuI&currency=GBP&amount=200.00&standard-vertical-stack=true&standard-horizontal-stack=true&paypal-standalone=true&pay-later-standalone=true&paypal-credit-standalone=true&all-standalone=true)

Screenshot 2024-05-15 at 10 38 54 AM

❤️ Thank you!

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -91,7 +91,7 @@ import {
   getButtonSize,
   getButtonExperiments,
   getModal,
-  getURIPopup,
+  openPopupAtURI,
 } from "./util";
 
 export type ButtonsComponent = ZoidComponent<ButtonProps>;
@@ -752,7 +752,7 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
               });
 
             if (clickUrlOverride) {
-              return getURIPopup(
+              return openPopupAtURI(
                 clickUrlOverride,
                 merchantRequestedPopupsDisabled
               );
@@ -775,7 +775,7 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
           return ({ clickUrlOverride }) => {
             // messageType, offerType, offerCountryCode, and creditProductIdentifier are also passed in, and may be used in an upcoming message hover logging feature
 
-            // no preloading the modal for purchase protection message: that opens a popup instead
+            // messages with clickUrlOverride will open a popup instead of the modal, so don't preload the modal
             if (!clickUrlOverride) {
               // lazy loads the modal, to be memoized and executed onMessageClick
               const { clientID, merchantID } = props;
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -91,6 +91,7 @@ import {
   getButtonSize,
   getButtonExperiments,
   getModal,
+  getURIPopup,
 } from "./util";
 
 export type ButtonsComponent = ZoidComponent<ButtonProps>;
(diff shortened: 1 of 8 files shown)


src/zoid/buttons/component.jsx
```diff
@@ -757,12 +771,15 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
-          return () => {
-            // offerType, messageType, offerCountryCode, and creditProductIdentifier are passed in and may be used in an upcoming message hover logging feature
-
-            // lazy loads the modal, to be memoized and executed onMessageClick
-            const { clientID, merchantID } = props;
-            return getModal(clientID, merchantID);
+          return (messageType) => {
+            // offerType, offerCountryCode, and creditProductIdentifier are also passed in, and may be used in an upcoming message hover logging feature
+
+            // no preloading the modal for purchase protection message: that opens a popup instead
```
danzhaaspaypal, line 777: should we consider preloading the popup content?

src/zoid/buttons/component.jsx
```diff
@@ -716,9 +717,16 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
           }) => {
-            const { message, clientID, merchantID, currency, buttonSessionID } =
-              props;
+            const {
+              message,
+              clientID,
+              merchantID,
+              currency,
+              buttonSessionID,
+              merchantRequestedPopupsDisabled,
```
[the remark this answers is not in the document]
danzhaaspaypal: This implementation was inspired by an [existing popup implementation](https://github.com/paypal/paypal-checkout-components/blob/04cca79a5ebcd7893488bc77fca214897452de78/src/zoid/buttons/prerender.jsx#L64) and thought it would be best to add its' considerations to this feature as well.  Glad you found utility for it elsewhere!

yeul (approved):

wsbrunson (approved):
I don't see any issues with this but wanted to ask a question about webviews

is this completely configurable by the merchant? One issue we see with `window.open` / popups is that they don't work on mobile in-app webviews. Our checkout popup actually renders inline on the page inside of an iframe if it detects that it is in a webview.

not sure if we need to go that far for this popup or if we suggest to merchants to not use that setting if they are using mobile in-app webviews

src/zoid/buttons/util.js
```diff
@@ -410,3 +413,29 @@ export const getModal: (
...
+        popup(uri, {
+          width: DEFAULT_POPUP_SIZE.WIDTH,
+          height: DEFAULT_POPUP_SIZE.HEIGHT,
+        })
+      );
+    } else {
+      return window.open(uri, "_blank");
+    }
+  } catch (err) {
+    getLogger()
```
ravishekhar, line 433: I would suggest adding a fallback logic here to fallback to modal experience if popup fails.

ravishekhar (approved):

test/integration/tests/button/message.js
```diff
@@ -451,14 +452,18 @@ describe(`paypal button message`, () => {
     });
   });
 
-  describe("modal", () => {
+  describe.only("modal", () => {
```
spencersablan, line 455: Was this `.only` left in accidentally? 😅

spencersablan (commented):
Quick find but otherwise lgtm 😄 👾

test/integration/tests/button/message.js
danzhaaspaypal: 😅 indeed, good eye. let me fix that

src/zoid/buttons/util.js
danzhaaspaypal: Thanks for the suggestion @ravishekhar. Sounds like we will be doing some dev work to make a purchase protection modal to address this issue that you and @wsbrunson called out.

Seavenly:
Changed back to a draft PR for the time being while the details are worked out internally to address the issue raised about webview popups.

PR #2380, opened 2024-05-01
feat: dynamic messages - modal instrumentation

paypal-rosman:

❤️ Thank you!


src/zoid/buttons/component.jsx
```diff
@@ -743,10 +743,15 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
-            const modalInstance = await getModal(clientID, merchantID);
+            const modalInstance = await getModal(
+              clientID,
+              merchantID,
+              buttonSessionID
+            );
             return modalInstance?.show({
               amount,
               offer: offerType,
+              buttonSessionId: buttonSessionID,
```
Seavenly, line 754: We should only need to pass the `buttonSessionID` into either `getModal` or `show`. I don't think we need to do both. Since we need it in `getModal` for the events, we can probably omit it from `show` since the `buttonSessionID` value shouldn't change.

paypal-rosman: okay i will remove one from the show right now

danzhaaspaypal (approved):

wsbrunson (approved):

perco12 (approved):

PR #2381, opened 2024-05-01
fix: dynamic messages - pass merchant id from cpnw callback, flush logger

paypal-rosman:
### Description

we are passing merchantID from a callback handled in an internal service. We are also flushing the logger that way on page navigation we can send up logger events 

❤️ Thank you!

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -716,25 +716,10 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
             messageType,
             offerCountryCode,
             creditProductIdentifier,
-            merchantID: serverMerchantId,
           }) => {
-            const {
-              message,
-              clientID,
-              currency,
-              buttonSessionID,
-              merchantID: partnerMerchantId,
-            } = props;
+            const { message, clientID, currency, buttonSessionID, merchantID } =
+              props;
             const amount = message?.amount;
-            const isPartnerMerchantIdEmpty =
-              !partnerMerchantId ||
-              partnerMerchantId.length === 0 ||
-              partnerMerchantId[0] === "";
-            // check to see if a partner merchant id integration exists
-            // if not grab the server merchant id
-            const merchantID = isPartnerMerchantIdEmpty
-              ? serverMerchantId
-              : partnerMerchantId?.join();
 
             getLogger()
               .info("button_message_click")
@@ -745,7 +730,6 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
                 [FPTI_KEY.CONTEXT_ID]: buttonSessionID,
                 [FPTI_KEY.CONTEXT_TYPE]: "button_session_id",
                 [FPTI_KEY.EVENT_NAME]: "message_click",
-                [FPTI_KEY.SELLER_ID]: merchantID,
                 [FPTI_KEY.BUTTON_MESSAGE_OFFER_TYPE]: offerType,
                 [FPTI_KEY.BUTTON_MESSAGE_CREDIT_PRODUCT_IDENTIFIER]:
                   creditProductIdentifier,
@@ -773,17 +757,8 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
         type: "function",
         required: false,
         value: ({ props }) => {
-          return ({ merchantID: serverMerchantId }) => {
-            const { clientID, merchantID: partnerMerchantId } = props;
-            const isPartnerMerchantIdEmpty =
-              !partnerMerchantId ||
-              partnerMerchantId.length === 0 ||
-              partnerMerchantId[0] === "";
-            // check to see if a partner merchant id integration exists
-            // if not grab the server merchant id
-            const merchantID = isPartnerMerchantIdEmpty
-              ? serverMerchantId
-              : partnerMerchantId?.join();
+          return () => {
+            const { clientID, merchantID } = props;
             // offerType, messageType, offerCountryCode, and creditProductIdentifier are passed in and may be used in an upcoming message hover logging feature
 
             // lazy loads the modal, to be memoized and executed onMessageClick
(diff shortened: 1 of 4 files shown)


src/zoid/buttons/component.jsx
```diff
@@ -716,11 +716,17 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
-            const { message, clientID, merchantID, currency, buttonSessionID } =
-              props;
+            const { message, clientID, currency, buttonSessionID } = props;
             const amount = message?.amount;
 
+            // since we are adding our own merchantID
+            // this should override the default coming from sdk client
+            getLogger().addTrackingBuilder(() => ({
+              [FPTI_KEY.SELLER_ID]: merchantID,
+            }));
```
Seavenly, line 728: What happens when an actual `merchantID` is passed in as part of the SDK script (i.e. partner integration)? Will this unintentionally override that or are we still good there? Also will this impact other non-messaging events that are fired from outside the IFrame since `addTrackingBuilder` changes the shared logger instance?

src/zoid/buttons/util.js
```diff
@@ -396,7 +396,7 @@ export const getModal: (
 
     return window[namespace].MessagesModal({
       account: `client-id:${clientID}`,
-      merchantId: merchantID?.join(",") || undefined,
+      merchantId: merchantID,
```
Seavenly, line 399: For the modal input here we want this value to be the SDK script `merchantID` value because we don't want to pass in a `merchantId` value unless it is a partner integration, otherwise it may impact authentication.

src/zoid/buttons/component.jsx
```diff
@@ -730,7 +736,7 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
                 [FPTI_KEY.CONTEXT_ID]: buttonSessionID,
                 [FPTI_KEY.CONTEXT_TYPE]: "button_session_id",
                 [FPTI_KEY.EVENT_NAME]: "message_click",
-                [FPTI_KEY.SELLER_ID]: merchantID?.join(","),
+                [FPTI_KEY.SELLER_ID]: merchantID,
```
Seavenly, line 739: If we're including this above, do we also need it here?

src/zoid/buttons/util.js
paypal-rosman: good callout, ill fix this one

src/zoid/buttons/component.jsx
paypal-rosman: let me attempt that now

src/zoid/buttons/component.jsx
paypal-rosman: going to test this to see if that is the case

Seavenly: Do we need this?

src/zoid/buttons/component.jsx
```diff
@@ -776,8 +801,29 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+              partnerMerchantId[0] === "";
+            // check to see if a partner merchant id integration exists
+            // if not grab the server merchant id
+            const merchantID = isPartnerMerchantIdEmpty
+              ? serverMerchantId
+              : partnerMerchantId?.join();
+
+            // override with server id if partner does not exist
+            getLogger().addTrackingBuilder(() => ({
+              [FPTI_KEY.SELLER_ID]: merchantID,
```
Seavenly, line 825: Would something similar to `props.merchantID.toString() || serverMerchantId` be the same result without the need for some of the code above?

src/zoid/buttons/component.jsx
```diff
@@ -757,11 +773,20 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+            const { clientID, merchantID: partnerMerchantId } = props;
+            const isPartnerMerchantIdEmpty =
+              !partnerMerchantId ||
+              partnerMerchantId.length === 0 ||
+              partnerMerchantId[0] === "";
+            // check to see if a partner merchant id integration exists
+            // if not grab the server merchant id
+            const merchantID = isPartnerMerchantIdEmpty
+              ? serverMerchantId
+              : partnerMerchantId?.join();
```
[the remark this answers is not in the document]
paypal-rosman: i will modify this, i thought we wanted both server + partner going to it. let me change this

src/zoid/buttons/component.jsx
paypal-rosman: yea that should work ill update

test/integration/windows/button/index.jsx
```diff
@@ -236,8 +236,8 @@ if (onRender) {
     click() {
       getElement(".paypal-button", document).click();
     },
-    hoverMessage(): ZalgoPromise<void> | void {
-      return window.xprops.onMessageHover();
+    hoverMessage({ merchantID }): ZalgoPromise<void> | void {
+      return window.xprops.onMessageHover({ merchantID });
```
Seavenly, line 240: It looks like we're no longer consuming the `merchantID` from the hover callback. Are these changes able to be reverted?

Seavenly (approved):

danzhaaspaypal (approved):

danzhaaspaypal (approved):

src/zoid/buttons/component.jsx
```diff
@@ -776,9 +776,16 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
             offerCountryCode,
             creditProductIdentifier,
+            merchantID: serverMerchantId,
           }) => {
+            // merchantID that comes from props is an array
             const { message, buttonSessionID, currency, merchantID } = props;
 
+            // override with server id if partner does not exist
+            getLogger().addTrackingBuilder(() => ({
+              [FPTI_KEY.SELLER_ID]: merchantID?.toString() || serverMerchantId,
```
wsbrunson, line 786: this already exists in `sdk-client`: https://github.com/paypal/paypal-sdk-client/blob/main/src/tracking.js#L138

I would consider changing this to
```
if (serverMerchantId) {
  getLogger().addTrackingBuilder(() => ({
    [FPTI_KEY.SELLER_ID]: serverMerchantId,
  }))
}
```

src/zoid/buttons/util.js
```diff
@@ -396,7 +396,7 @@ export const getModal: (
 
     return window[namespace].MessagesModal({
       account: `client-id:${clientID}`,
-      merchantId: merchantID?.join(",") || undefined,
+      merchantId: merchantID?.join() || undefined,
```
wsbrunson, line 399: did you mean to drop the `","` in the join?

paypal-rosman: i think by default it makes it comma delimited but i can add it back in for clarity

src/zoid/buttons/component.jsx
paypal-rosman: let me get this added

src/zoid/buttons/util.js
wsbrunson: wow i had no idea: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join

paypal-rosman: i went ahead and added it back in just in case

paypal-rosman: i addressed the comments

wsbrunson (approved):

wsbrunson (approved):

PR #2384, opened 2024-05-01
fix: handle tagline field on ncps hosted buttons

imbrian:
### Description

Hosted Buttons are not able to disable the `tagline` on horizontal layout buttons. This change uses the `tagline` value from `/ncp/api/form-fields`. Defaulting to `false` if the field is not provided.

Bug fix: LI-45315

### Screenshots
Before:
tagline-before-screenshot

With fix:
tagline-screenshot


spencersablan (approved):
Nice! 🚀

src/hosted-buttons/index.test.js
```diff
@@ -150,5 +155,66 @@ describe("HostedButtons", () => {
...
+      render: renderMock,
+      isEligible: vi.fn(() => false),
+    }));
+    // $FlowIssue
+    getButtonsComponent.mockImplementationOnce(() => Buttons);
+    const HostedButtons = getHostedButtonsComponent();
+    // $FlowIssue
+    request.mockImplementationOnce(() =>
+      // eslint-disable-next-line compat/compat
+      Promise.resolve({
```
jshawl, line 172: can this reuse the previous response from https://github.com/paypal/paypal-checkout-components/blob/71026ef07f05bd150e29a9e0463d5ea5553b2768/src/hosted-buttons/index.test.js#L34 ?

src/hosted-buttons/index.test.js
```diff
@@ -150,5 +155,66 @@ describe("HostedButtons", () => {
...
+                value: "false",
+              },
+            ],
+          },
+        },
+      })
+    );
+    await HostedButtons({
+      hostedButtonId: "B1234567890",
+      fundingSources: ["paypal", "venmo"],
```
jshawl, line 207: I think `fundingSources` can be removed here

src/hosted-buttons/index.test.js
```diff
@@ -150,5 +155,66 @@ describe("HostedButtons", () => {
...
+    }).render("#example");
+    expect(Buttons).toHaveBeenCalledWith(
+      expect.objectContaining({
+        hostedButtonId: "B1234567890",
+        style: expect.objectContaining({ tagline: false }),
+      })
+    );
+    expect(Buttons).toHaveBeenCalledTimes(2);
+    expect(renderMock).toHaveBeenCalledTimes(0);
+    expect.assertions(3);
```
jshawl, line 217: awesome test coverage!

src/hosted-buttons/utils.test.js
```diff
@@ -89,6 +93,107 @@ test("getHostedButtonDetails", async () => {
...
+              value: "vertical",
+            },
+            {
+              name: "color",
+              value: "gold",
+            },
+            {
+              name: "button_text",
+              value: "paypal",
+            },
```
jshawl, line 181: ^ similar comment about reusing the previous response variables

src/hosted-buttons/utils.test.js
```diff
@@ -89,6 +93,107 @@ test("getHostedButtonDetails", async () => {
...
+          ],
+        },
+      },
+    })
+  );
+  await getHostedButtonDetails({
+    hostedButtonId,
+    fundingSources: [],
```
jshawl, line 189: I think `fundingSources` can be removed here

jshawl (approved):
Looks good! A few small non-blocking suggestions

src/hosted-buttons/index.test.js
imbrian: Yeah, thought about it, but the tests are changing the value for `tagline` (true, false, undefined). Felt a new object was cleaner than nested spreads and mapping over the array to change one value.

src/hosted-buttons/utils.test.js
nikrom17: Yeah, I would remove these. This is how we are temporarily rendering the v2 experience but will remove the `fundingSources` in this [ticket](https://paypal.atlassian.net/browse/DTPPCPSDK-2229) that's currently in the sprint.

The tagline only gets rendered in the horizontal smart stack and the v2 experience is rendering 2 standalone buttons.

imbrian: It's currently required on the type. Are we removing it completely in that ticket? Might make it optional for now, and @nbierdeman can address in that ticket?

PR #2385, opened 2024-05-01
DTPPCPSDK-2227: Return V2 Variables from getHostedButtonDetails 👾

spencersablan:
### Description

- 🥖 **Prepares us for the NCPS 2.0 changes** - All we need to do for this PR is extract those variables from the response to prepare for future changes.
- 📝  **Added tests for v2 response** -  Since we need to support both V1 & V2 responses from `/ncp/api/form-fields/`, I added some quick unit tests to make sure `getHostedButtonDetails` can handle either response appropriately.

**JIRA:** [DTPPCPSDK-2227](https://paypal.atlassian.net/browse/DTPPCPSDK-2227)

❤️ Thank you!

### DIFF
--- a/src/hosted-buttons/types.js
+++ b/src/hosted-buttons/types.js
@@ -34,10 +34,10 @@ export type HostedButtonsInstance = {|
   render: (string | HTMLElement) => Promise<void>,
 |};
 
-export type HostedButtonPreferences = {
+export type HostedButtonPreferences = {|
   buttonPreferences: $ReadOnlyArray<string>,
   eligibleFundingMethods: $ReadOnlyArray<string>,
-};
+|};
 
 export type HostedButtonDetailsParams =
   (HostedButtonsComponentProps) => Promise<{|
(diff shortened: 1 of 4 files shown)


jshawl (approved):
:shipit:

src/hosted-buttons/types.js
```diff
@@ -34,6 +34,11 @@ export type HostedButtonsInstance = {|
   render: (string | HTMLElement) => Promise<void>,
 |};
 
+export type HostedButtonPreferences = {|
+  buttonPreferences: $ReadOnlyArray<string>,
```
imbrian, line 38: Possible to narrow these `string` types to known preferences and sources? Could do that in follow-up implementation PR too

imbrian (approved):
small comment re: preference types

src/hosted-buttons/types.js
spencersablan: I like that!

spencersablan: Donezo funzo

nbierdeman (approved):
Nice! 🤘 🚀 Should we add `tagline` to the tests?

imbrian:
> Nice! 🤘 🚀 Should we add `tagline` to the tests?

`tagline` isn't implemented; fix is [here](https://github.com/paypal/paypal-checkout-components/pull/2384), but waiting for @spencersablan to merge this to deal with conflicts on that branch.

PR #2386, opened 2024-05-02
Fix for NCP height string to number conversion 👾

spencersablan:
### Description

JIRA: [DTPPCPSDK-2227](https://paypal.atlassian.net/browse/DTPPCPSDK-2227)

- 🔨 _**Height should be a number, not a string**_ - This fixes the error that yells at you in your dreams as you try to sleep but you open your eyes and height is in front of your face punishing you for attempting to make him a string, when he is in fact, a number.  He then threatens you but you cannot move.  He repeats himself, "height should be a number. height should be a number."  You wake up from the nightmare just for it to all be true. This actually happened.

❤️ Thank you!

### DIFF
--- a/src/hosted-buttons/utils.js
+++ b/src/hosted-buttons/utils.js
@@ -98,7 +98,7 @@ export const getHostedButtonDetails: HostedButtonDetailsParams = async ({
       shape: getButtonVariable(variables, "shape"),
       color: getButtonVariable(variables, "color"),
       label: getButtonVariable(variables, "button_text"),
-      height: parseInt(getButtonVariable(variables, "height")) || undefined,
+      height: parseInt(getButtonVariable(variables, "height"), 10) || undefined,
     },
     version: body.version,
     buttonContainerId: body.button_container_id,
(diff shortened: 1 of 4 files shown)


src/hosted-buttons/utils.js
```diff
@@ -98,7 +98,7 @@ export const getHostedButtonDetails: HostedButtonDetailsParams = async ({
       shape: getButtonVariable(variables, "shape"),
       color: getButtonVariable(variables, "color"),
       label: getButtonVariable(variables, "button_text"),
-      height: getButtonVariable(variables, "height"),
+      height: parseInt(getButtonVariable(variables, "height")) || undefined,
```
imbrian, line 101: Maybe this isn't possible from API, but `Boolean(parseInt('0'))` is `false`, so would end up `undefined` here

spencersablan: Since `height` needs to be between 22-55 anyway, I don't think we have to worry about this 😄

spencersablan: Though, I do have to add the radix

PR #2392, opened 2024-05-16
refactor: update message object to accept array offer param and  send offer as string

perco12:
### Description

Ensures that the message object `offer` input remains an array but output results in message having `offer` property as a comma-delimited string

❤️ Thank you!


src/ui/buttons/props.js
```diff
@@ -765,18 +764,22 @@ export function normalizeButtonMessage(
...
-        )}`
-      );
+      if (typeof offer !== "string") {
+        throw new TypeError(
+          `Expected message.offer to be an array of strings, got: ${String(
+            offer
+          )}`
+        );
+      }
+      offer = offer.split(",");
```
Seavenly, line 774: Would the following help improve clarity of these different `if` blocks by removing a level of nesting and tying the conversion from string to array code closer to its respective `if` code?

``` js
if (typeof offer !== "undefined") {
    if (typeof offer === "string") {
        offer = offer.split(",");
    }
    if (!Array.isArray(offer)) {
        throw new TypeError(/* ... */)
    }

    // ...
}
```

danzhaaspaypal (approved):

src/ui/buttons/props.js
```diff
@@ -430,7 +430,7 @@ export type ApplePaySessionConfigRequest = (
 
 export type ButtonMessage = {|
   amount?: number,
-  offer?: $ReadOnlyArray<$Values<typeof MESSAGE_OFFER>>,
+  offer?: $Values<typeof MESSAGE_OFFER> | string,
```
wsbrunson, line 433: if this is going to be a comma separated string, should the type just be `string`?

wsbrunson (approved):

src/ui/buttons/props.js
perco12: that is a good call, as the `join` statement makes the output a generic string, instead of our strict flow type.

Seavenly: I think this may be meant to be `$ReadOnlyArray<$Values<typeof MESSAGE_OFFER>> | string` since we want to accept the input as both and array or comma-delimited string. The merchant input will be `["a", "b"]` vs `"a,b"` for better ergonomics. We convert it internally to a comma-delimited string so that we pass it to the endpoint as a string to get around the `zoid` `dotify` encoding issue.

Seavenly: Oh apologies. I misspoke and got my wires crossed. `ButtonMessage` is the output of the function and not the input. :sweat_smile: My comment above can be disregarded.

wsbrunson (approved):

wsbrunson (approved):

PR #2393, opened 2024-05-17
feat(venmo): relax eligibility

mnicpt:
### Description
Relaxing venmo first render eligibility now that Venmo web experience is live and can better manage fallback from native app switch.

image

### PR
https://paypal.atlassian.net/browse/DTVP-75

**Note:**The last requirement, 1d, will be implemented at a later date.

❤️ Thank you!


src/funding/venmo/config.jsx
```diff
@@ -42,7 +42,6 @@ export function getVenmoConfig(): FundingSourceConfig {
         experiment.venmoWebEnabled === false
       ) {
         return {
-          native: true,
           popup: true,
```
wsbrunson, line 45: If device is Mobile & Venmo Web is disabled, then require `popup`? Am I reading that right? From this little snippet it seems like native should be true and popup should be false but I am likely missing context of the bigger picture

I just read your description though, is this saying:

we may still app-switch, but if we can't popup then definitely don't continue

mnicpt: @wsbrunson By removing native, it is saying SFVC is ok. We will catch that just prior to app switching. If in SFVC there, then we will fallback to web. Popup relates to WebView and want to keep that there.

PR #2395, opened 2024-05-20
Add disableDefaultStyle prop for Card Fields

mchoun:
### Description

Currently, the Card Fields are constrained by the allowed styling props and does not allow a height to be passed. Adding this prop will pass the flag to the Card Fields component in `scnw`, modify the prerender template to fit the merchant's parent container, and conditionally change the dimensions passed to the zoid component. 

We are making these changes to facilitate more custom styling of the Card Fields. 

Jira Ticket - DTPPCPSDK-2203

*Default styling*
image

*Disabled Default Styling*
image

image

`scnw` PR - #682

❤️ Thank you!

### DIFF
--- a/src/zoid/card-fields/component.jsx
+++ b/src/zoid/card-fields/component.jsx
@@ -127,7 +127,14 @@ type CardFieldsChildren = {|
 const url = () =>
   `${getPayPalDomain()}${__PAYPAL_CHECKOUT__.__URI__.__CARD_FIELD__}`;
 
+const DEFAULT_HEIGHT = "78px";
+
 const prerenderTemplate = ({ props, doc }) => {
+  if (props.parent.props.disableDefaultStyle) {
+    props.style.height = "100%";
+    console.log("it's true, baby!");
+  }
+  console.log("mervin", props);
   return (
     <CardPrerender nonce={props.nonce} height={props.style?.height} />
   ).render(dom({ doc }));
@@ -147,7 +154,7 @@ export const getCardFieldsComponent: () => CardFieldsComponent = memoize(
         url,
 
         dimensions: {
-          height: "30px",
+          height: DEFAULT_HEIGHT,
           width: "100%",
         },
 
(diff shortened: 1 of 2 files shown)


src/zoid/card-fields/component.jsx
```diff
@@ -158,7 +173,7 @@ export const getCardFieldsComponent: () => CardFieldsComponent = memoize(
         },
 
         autoResize: {
-          height: true,
+          height: false,
```
mchoun, line 176: @sebastianfdz and I were not change this value conditionally since `autoResize` does not accept a callback. We decided to default this to `false` instead of `true` for all renders. We tested the default integration and it is working as expected and there are no breaking changes.

src/zoid/card-fields/prerender.jsx
```diff
@@ -5,61 +5,67 @@ import { node, type ChildType } from "@krakenjs/jsx-pragmatic/src";
...
           nonce={nonce}
-          innerHTML={`
-                        html, body {
-                            padding: 0;
-                            margin: 0;
-                            width: 100%;
-                            height: 100%;
-                        }
+          innerHTML={
+            disableDefaultStyle
```
mchoun, line 25: We decided to not display the prerender template when the `disableDefaultStyle` flag is passed since we now rely on the container style. That style is what will be displayed while the card renders. 

If we did not do this, the existing pre-render style may not be a good experience.

siddy2181 (approved):
Looks great!

src/zoid/card-fields/component.jsx
```diff
@@ -587,6 +611,12 @@ export const getCardFieldsComponent: () => CardFieldsComponent = memoize(
           queryParam: true,
         },
 
+        disableDefaultStyle: {
+          type: "boolean",
+          required: false,
+          queryParam: true,
```
mchoun, line 617: Is the queryParam parameter necessary? Can this be removed?

siddy2181, line 617: Non blocker. Do we want to default this param to `false`?

mchoun:
Team decided to go down the path of adding support for additional css properties instead of this single flag that is doing too much.

PR #2401, opened 2024-05-25
Add shopperInsightsUsed xprop 👾

spencersablan:
### Description

**JIRA:** [DTPPCPSDK-2369](https://paypal.atlassian.net/browse/DTPPCPSDK-2369)

- 🙅‍♂️ _**Added new xprop**_ - Added `shopperInsightsUsed` as new xprop that returns a function so that we can call it in smart-payment-buttons

❤️ Thank you!

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -41,6 +41,7 @@ import {
   getExperimentation,
   getSDKAttribute,
   getJsSdkLibrary,
+  wasShopperInsightsUsed,
 } from "@paypal/sdk-client/src";
 import {
   rememberFunding,
@@ -950,6 +951,13 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
       sessionState: {
         type: "object",
         value: () => sessionState,
+      },
+
+      shopperInsightsUsed: {
+        type: "function",
+        value: () => {
+          return () => true;
+        },
       },
 
       stageHost: {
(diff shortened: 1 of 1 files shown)


src/zoid/buttons/component.jsx
```diff
@@ -952,6 +953,13 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
         value: () => sessionState,
       },
 
+      shopperInsightsUsed: {
```
wsbrunson, line 956:
```suggestion
      getShopperInsightsUsed: {
```

we should probably call this `get...` because its a function not a property

spencersablan: great callout here!

src/zoid/buttons/component.jsx
```diff
@@ -948,6 +949,13 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
         value: () => sessionState,
       },
 
+      getShopperInsightsUsed: {
+        type: "function",
+        value: () => {
+          return () => wasShopperInsightsUsed();
+        },
+      },
```
wsbrunson, line 957: this is totally a nit so ignore if you like it the way you wrote it.

```suggestion
        value: () => wasShopperInsightsUsed,
      },
```

wsbrunson (approved):

src/zoid/buttons/component.jsx
spencersablan: Good nit 🚀

PR #2404, opened 2024-06-05
Button autoResize by button container

imbrian:
### Description

This change sets `autoResize.element` on the Button component to the button container. I'm introducing a new class to prevent reliance on an existing class with other uses.

Unblocks DTPPCPSDK-1915 #2361 
Does not implement `disableMaxHeight`

Resolves autoResize issue introduced when adding the `disableMaxHeight` prop, which sets container heights within the iframe to 100%. This height prevented resize events, because zoid [observes](https://github.com/krakenjs/zoid/blob/main/docs/api/create.md#autoresize--height-boolean-width-boolean-element-string-) the `body` [by](https://github.com/krakenjs/zoid/blob/main/src/child/child.js#L272) [default](https://github.com/krakenjs/zoid/blob/main/src/parent/parent.js#L1149). A height of 100% prevented this resize from occurring, because it would always match the initial container height.

### Reproduction steps

The issue occurred when the window was resized after the second render.

Test environment with this PR + current mainline first/second render: `te-button-autoresize`

[Storybook TE with this change and disable max height changes in SCNW](https://github.paypal.com/pages/Core-SDK/button-redesign-storybook/?path=/story/edge-cases-button-disablemaxheight--container-with-fixed-height) (does not include the changes from #2361 to fully support `disableMaxHeight`, but this change has been tested with #2361 locally)

This adds a new class and sets button resize behavior based on this new class. Both first and second render need this version to resize properly.

This change has been tested against the current mainline branch for first/second render. It should be released ahead of the `disableMaxHeight` changes. The server-side `disableMaxHeight` changes can then be released, following by #2361 

❤️ Thank you!

### DIFF
--- a/src/constants/class.js
+++ b/src/constants/class.js
@@ -2,6 +2,8 @@
 
 export const CLASS = {
   CONTAINER: ("paypal-button-container": "paypal-button-container"),
+  AUTORESIZE_CONTAINER:
+    ("paypal-autoresize-container": "paypal-autoresize-container"),
   BUTTON_ROW: ("paypal-button-row": "paypal-button-row"),
   BUTTON: ("paypal-button": "paypal-button"),
   BUTTON_LABEL:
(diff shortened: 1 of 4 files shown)


nikrom17 (approved):
✅

wsbrunson (approved):

ravishekhar (approved):
💯

PR #2406, opened 2024-06-10
fix(venmo): required popup:true

diegoHdzZ:
### Description

Updated eligibility to always return popup `true`.

[https://paypal.atlassian.net/browse/DTVP-598](https://paypal.atlassian.net/browse/DTVP-598)

❤️ Thank you!


wsbrunson:
would this need to wait until venmo web has been fully released?

PR #2410, opened 2024-06-13
use query param for atomics waiting state

wsbrunson:
Checkout window is already setup to interact with a query param sent from the instrumentation library


mchoun (approved):
👍

PR #2412, opened 2024-06-14
Fix Venmo Eligibility for Venmo Web & Shipping Callbacks 👾

spencersablan:
### Description
**JIRA:** https://paypal.atlassian.net/browse/LI-50460

_**🪄 Fixes Venmo's disappearing act**_ - Venmo was doing this stupid magic trick where it would show up on first render, then _✨ d i s a p p e a r ✨_ on second render.  This was because the `experiment` prop existed on first render but did not include `enableVenmoWeb`. So when you passed a shipping callback, it would be eligible according to the first render, but then the second render `experiment` prop would have `enableVenmoWeb: false`. 
Now, it won't show Venmo at all if you pass a shipping callback _unless_ you have Venmo Web enabled.
For now, if you pass a shipping callback & Venmo Web is enabled, this will cause a new magic trick where the Venmo button _✨ a p p e a r s ✨_ on second render... but that's a little better UX than what it was previously.  This can be fixed by adding the `venmoWebEnabled` experiment to the first render.

_**🪵 Added log for eligibility**_ - I'm a little tired of eligibility bugs being a guessing game outside of my local environment.  I added a log for when a funding source is eligible according to the funding eligibility check, but then fails in the funding config eligibility check.  If I had this before looking into this issue,  I would've been able to triage this in a fraction of the time. Hoping this helps us all in the future. 

❤️ Thank you!


src/funding/funding.js
```diff
@@ -114,26 +115,30 @@ export function isFundingEligible(
...
+        onShippingChange ||
+        onShippingAddressChange ||
+        onShippingOptionsChange
+    ),
+    wallet,
+    displayOnly,
+  };
+
+  if (fundingConfig.eligible && !fundingConfig.eligible?.(eligibilityProps)) {
+    getLogger().info(`${source}_ineligible`, {
```
wsbrunson, line 137: beautiful

src/funding/venmo/config.jsx
```diff
@@ -22,15 +22,22 @@ export function getVenmoConfig(): FundingSourceConfig {
...
       if (experiment && experiment.enableVenmo === false) {
         return false;
       }
 
-      if (
-        experiment &&
-        experiment.venmoWebEnabled === false &&
-        shippingChange
-      ) {
+      /**
```
wsbrunson, line 33: nice comment

wsbrunson (approved):
great work!

src/funding/funding.js
spencersablan: just like you homie

nikrom17 (approved):

PR #2413, opened 2024-06-17
Feature/add disable prerender option

sebastianfdz:
### Description

Ticker reference: DTPPCPSDK-2410

**WHAT**
We are adding a mechanism for merchants to disable the prerender template for when `CardFields` are loading.

**WHY**
`CardFields` include a prerender template by default when they are loading. Merchants might not want it to render because it may not fit with their UX/UI (darkmode for example).

**HOW**
We will expose a new optional property `styleOptions` where we will be able to incrementally add options for merchants to fully customize the way they render our components. Inside `styleOptions` we will have a `disablePrerender` flag that when `true`, will not render the prerender template. 

❤️ Thank you!


src/zoid/card-fields/component.jsx
```diff
@@ -51,6 +51,9 @@ type CardFieldsProps = {|
   style?: {|
     height: number,
   |},
+  styleOptions?: {|
+    disablePrerender?: boolean,
```
elizabethmv, line 55: Should this be something like `disableMaxHeight` so that it is consistent with the buttons style options? I'm not sure though, cc @wsbrunson

wsbrunson: are you asking if it should be in the `style` object?

elizabethmv: Oh actually I got flags confused, disregard

elizabethmv (approved):

Issue #707, opened 2018-06-05
Bundle size checkout.lib.js


michelalbers:
### Description

I recently ran the bundle analyzer over my webpack app and was astonished by the paypal bundle size which is above 1 MB:

![bildschirmfoto 2018-06-05 um 18 00 29](https://user-images.githubusercontent.com/646691/40988034-5cf16a50-68ea-11e8-9b8a-9dbec4ce6836.png)

Is there a way to split up the sdk so we can import only the files we need? 

### Steps to reproduce

Build any module with paypal SDK and webpack.

### Affected browsers

all

### Edit

The suggested way to load the sdk via

    <script src="https://www.paypalobjects.com/api/checkout.js" data-version-4></script>

is 1.5 MB (!!!!) in size. Even minified its over 900 kB.

![dafuq](https://media.giphy.com/media/wxw2e19ZMsjio/giphy.gif)

trainerbill:
v5 is being developed to mitigate this.  will be using chunking to dynamically load pieces of the sdk that are used.

leeoniya:
@trainerbill,

until v5 ships, you guys should leave one of these "size" issues open rather than closing them all (#443, https://github.com/khoanguyen96/vue-paypal-checkout/issues/30, #707).

loading 992k [1] of script when previously 0k was required [2] is a bit absurd. for instance, we don't even ship 100k of our own scripts in total.

despite [3] being "deprecated", the recommended replacement is not a viable option, especially on mobile where it takes several seconds to load.

is there any movement on v5 since the last update?

thanks!

[1] https://www.paypalobjects.com/api/checkout.min.js
[2] `https://www.paypal.com/cgi-bin/webscr?cmd=_express-checkout&token={token}`
[3] https://developer.paypal.com/docs/classic/express-checkout/ht_ec-singleItemPayment-curl-etc/

trainerbill:
I can't give any timeframes on GA release, but it is an active project.  For now, it is recommended to load the script async on a previous page so it is cached.

https://github.com/paypal/paypal-checkout/blob/master/docs/performance.md

I will reopen for visibility since its a good title.

leeoniya:
> For now, it is recommended to load the script async on a previous page so it is cached.

for our simple (and common) auth -> capture use case there's no reason to invest time to switch to the new `checkout.js` api while it needs these crutches to produce a sub-par experience. hopefully the deprecated api's stick around until v5 is GA and weighs < 50k (pre gzip).

[...]

PR #2417, opened 2024-06-26
Update pre-render template to take the height of the input selector

mchoun:
### Description

Currently, the pre-render template will only look for a height property to be passed at the top level of the custom style object. Passing the height in this way introduced a bug where the CSS properties of the rendered Card Fields would not be applied. 

This PR is to allow the pre-render template to check for a height property to be passed to the input selector and to utilize that height.

scnw PR #707 introduced a bug where passing height at the top level would break the CSS of the rendered input field
Jira ticket - DTPPCPSDK-2442

**Top level height with input height**
Pre-render template at 100px and rendered input fields at 50px
```
      style: {
        height: "100px",
        input: {
          height: "50px",
          background: "#E79482",
        },
      },
```
image

**Only input height**
Pre-render template and rendered input fields at 50px
```
      style: {
        input: {
          height: "50px",
          background: "#E79482",
        },
      },
```
image

scnw PR #717

❤️ Thank you!

### DIFF
--- a/src/zoid/card-fields/component.jsx
+++ b/src/zoid/card-fields/component.jsx
@@ -128,9 +128,10 @@ const url = () =>
   `${getPayPalDomain()}${__PAYPAL_CHECKOUT__.__URI__.__CARD_FIELD__}`;
 
 const prerenderTemplate = ({ props, doc }) => {
-  return (
-    <CardPrerender nonce={props.nonce} height={props.style?.height} />
-  ).render(dom({ doc }));
+  const height = props.style?.height || props.style?.input?.height || null;
+  return (<CardPrerender nonce={props.nonce} height={height} />).render(
+    dom({ doc })
+  );
 };
 
 export type CardFieldsComponent = ZoidComponent<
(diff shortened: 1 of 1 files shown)


src/zoid/card-fields/component.jsx
```diff
@@ -128,9 +129,10 @@ const url = () =>
   `${getPayPalDomain()}${__PAYPAL_CHECKOUT__.__URI__.__CARD_FIELD__}`;
 
 const prerenderTemplate = ({ props, doc }) => {
-  return (
-    <CardPrerender nonce={props.nonce} height={props.style?.height} />
-  ).render(dom({ doc }));
+  const height = props.style?.height || props.style?.input?.height || null;
```
sebastianfdz, line 132: One nit pick is to use `??` rather than `||` because the it will only be triggered by `null` or `undefined` values. Merchant could potentially pass `height` as 0 (or any falsy value for that matter) and that would be a valid value but we would be nulling it and applying the default height.

mchoun: This is valid feedback and not a nitpick at all. I'll make the change!

siddy2181 (approved):
LGTM

sebastianfdz (approved):

PR #2420, opened 2024-07-23
Fix button styles when paypal button is not vaultable

nikrom17:
### Description

We had a styling bug where the gold color was being applied to the venmo and card buttons. This was due to the nomalizeButtonStyle function using the paypal funding source config to determine the default if no funding source was passed in as a prop. This was an issue for the smart stack use case when the paypal button was not eligible. Instead of using the default color for the funding source in the Button component, it was using the gold value set as a default fallback.

The paypal button was not eligible for vaulting when passing `displayOnly: ["vaultable"]`

Where we pull in the default button color based on the eligible funding source for the smart stack.
https://github.com/paypal/paypal-checkout-components/blob/main/src/ui/buttons/button.jsx#L110

https://paypal.atlassian.net/browse/LI-46995

Screenshot 2024-07-23 at 4 31 51 PM

❤️ Thank you!

### DIFF
--- a/src/ui/buttons/props.js
+++ b/src/ui/buttons/props.js
@@ -317,7 +317,7 @@ export type OnClick = (OnClickData, OnClickActions) => void;
 
 export type ButtonStyle = {|
   label: $Values<typeof BUTTON_LABEL> | void,
-  color: $Values<typeof BUTTON_COLOR>,
+  color?: $Values<typeof BUTTON_COLOR>,
   shape: $Values<typeof BUTTON_SHAPE>,
   tagline: boolean,
   layout: $Values<typeof BUTTON_LAYOUT>,
@@ -644,6 +644,7 @@ export function normalizeButtonStyle(
   }
 
   let {
+    color,
     label,
     layout = fundingSource
       ? BUTTON_LAYOUT.HORIZONTAL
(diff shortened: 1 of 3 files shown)


src/ui/buttons/props.js
```diff
@@ -664,9 +665,6 @@ export function normalizeButtonStyle(
     tagline = false;
   }
 
-  // if color is a falsy value, set it to the default color from the funding config
```
imbrian, line 667: only concern is that there's no test change required when there's a comment saying "we need this." sounds like we're just missing a test for this entirely though?

if so, we may want to add it, since it looks like this color logic is happening across two files

nikrom17: sounds good

I updated the tests, but they don't actually hit this specific use case. I tried adding `displayOnly: ['vaulatable']` and the skd would throw an error saying no valid funding sources were available.

I did add a comment in the buton.jsx file about the logic for choosing the default button color

ravishekhar: Looks good. I wonder if this issue could have been caught by unit tests. We should improve unit tests  with visual regressions in this module.

ravishekhar (approved):

src/ui/buttons/props.js
imbrian: The added tests at least account for taking away this comment.

If we can test it with VRT, we should be able to catch it with unit tests, but that's something we should create a ticket for in the new tech debt epic.

imbrian (approved):

src/ui/buttons/props.js
nikrom17: it definitely could have been caught with these tests but would require some updates to the integration test server to be able to render the use case. I could have spent more time trying to make that happen, but was also unsure of how the change would affect the other tests. 

I have a lot of thought on how to improve our testing suits and will add this use case as something we should be able to better capture.

PR #2421, opened 2024-07-24
feat: allows button message amount type to be string

danzhaaspaypal:
### Description

Allows button messages to support taking string values for the amount (ex: "100")

As a developer, I’d like the integration for passing the amount to buttons and messages to be the same.

❤️ Thank you!


danzhaaspaypal:
@wsbrunson could we have this reviewed please?

ravishekhar (approved):
lgtm

PR #2423, opened 2024-07-25
Add experiment prop to funding source validation 👾

spencersablan:
### Description
**JIRA:** [DTPPCPSDK-2495](https://paypal.atlassian.net/browse/DTPPCPSDK-2495)

_**🧍‍♂️ Fixes Venmo standalone with shipping callback**_ 
As of right now, the Venmo button is ineligible as a standalone button when you pass a shipping callback.  This is because the Venmo button is ineligible if (1) the `venmoWebEnabled` experiment is false or undefined _and_ (2) you pass a shipping callback. 
The problem was that the `venmoWebEnabled` experiment does not exist until second render and if you pass a funding source into your Buttons config, we validate the eligibility of that funding source on first render. 
To fix this we moved the `venmoWebEnabled` experiment to the first render, which allows us to have access to the `experiment` prop when we validate the eligibility of the funding soruce.

_**🫂 Fixes Venmo button's delayed arrival with a shipping callback**_
In [a previous PR](https://github.com/paypal/paypal-checkout-components/pull/2412), we had to temporarily settle for the Venmo button not showing on first render if you pass a shipping callback, then rendering if Venmo Web is enabled. But this still is not a great UX.  Since the `venmoWebEnabled` experiment is now moved to the first render, this delayed appearance is fixed. 😄

### Screenshots
**OLD:**
image

**NEW:** 
image

❤️ Thank you!


src/zoid/buttons/component.jsx
```diff
@@ -539,6 +536,7 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
               fundingSource,
               fundingEligibility,
               enableFunding,
+              experiment,
```
imbrian, line 539: do we have unit tests we could add to that would assert this doesn't get removed in the future? simple regression test. although I guess more value here would be tests comparing first and second render.

src/zoid/buttons/util.js
```diff
@@ -306,6 +307,7 @@ export function applePaySession(): ?ApplePaySessionConfigRequest {
 export function getButtonExperiments(): EligibilityExperiment {
   return {
     ...getVenmoExperiment(),
+    ...getFirstRenderExperiments(),
```
ravishekhar, line 310: Please validate if there are any other methods using `getButtonExperiments()` and may have any side affects to due this change.

ravishekhar (approved):

mnicpt (changes requested):
@spencersablan Please wait to merge as a product requirement change may not require this.

src/zoid/buttons/util.js
spencersablan: That I did 😄  Its only. used in zoid/buttons/component.jsx 🚀

src/zoid/buttons/component.jsx
spencersablan: Yeah, I just realized we have Venmo standalone button tests that didn't catch that this was broken in the first place! I'll update this if we decide to move forward with these changes.

Thanks B 😄

mnicpt (approved):

PR #2424, opened 2024-07-25
Fix button color logic for smart stack

nikrom17:
### Description

There was a recent [PR](https://github.com/paypal/paypal-checkout-components/pull/2420) to fix this [LI](https://paypal.atlassian.net/browse/LI-48101).

This PR introduced a new UI bug that affected the default colors of the smart stack. This PR fixes that issue

### Screenshots (if applicable)
With UI bug
Screenshot 2024-07-25 at 3 41 22 PM

With UI fix
Screenshot 2024-07-25 at 3 39 15 PM

### Groups who should review (if applicable)

❤️ Thank you!

PR #2426, opened 2024-08-21
Fix button color styles being applied incorrectly

nikrom17:
### Description

When the PayPal button is not eligible in a vaulting flow our logic for applying the button color was incorrect.

[LI-46995](https://paypal.atlassian.net/browse/LI-46995)

Relavant PRs
1. https://github.com/paypal/paypal-checkout-components/pull/2420
1. https://github.com/paypal/paypal-checkout-components/pull/2424

### Reproduction Steps
JS SDK script 
```html
<script src="https://www.paypal.com/sdk/js?client-id=B_AxqJZnX9Fg6HifHgzGWffqQhnRxsdiXR2eGsIS50nBpG8IPJXi7-bZA4zgvlDGaEs0QwxlTGbuGdRmo&merchant-id=NY9D8KUEC8W54&enable-funding=venmo"></script>
```

button configuraton
```js
paypal.Buttons({
    displayOnly: ["vaultable"],
    style: {
        color: ''
    },
  }).render('#paypal-container-id');`);
```

1. First relevant PR fixed the issue with the colors being applied incorrectly, but failed to account for the effects to the default colors on the smart stack (secondary colors)
image-20240808-185002

3. The second relevant PR attempted to fix the issue with the default colors but was unsuccessful and reintroduced the bug this LI pointed out
smart stack

5. This PR contains a solution for both of these issues 
Screenshot 2024-08-21 at 7 21 19 AM
Screenshot 2024-08-21 at 7 36 15 AM

❤️ Thank you!

### DIFF
--- a/src/ui/buttons/button.jsx
+++ b/src/ui/buttons/button.jsx
@@ -111,9 +111,19 @@ export function Button({
 
   // if no color option is passed in via style props
   if (color === "" || typeof color === "undefined") {
-    // if multiple buttons are being rendered (smart stack), we set default color as gold
     // if a single button is rendered, we set color to first option in the fundingSource config
-    color = multiple ? "gold" : colors[0];
+    color = colors[0];
+
+    // if multiple buttons are being rendered (smart stack), we set default color as gold > first
+    if (multiple) {
+      color = "gold";
+    }
+  }
+
+  // validate the first button rendered has a valid color
+  // this check is needed to validate the first button in a smart stack gets the correct color
+  if (i === 0 && !colors.includes(color)) {
+    color = colors[0];
   }
 
   // The secondary colors are used to render the smart stack (multiple buttons)
@@ -137,6 +147,8 @@ export function Button({
     logoColors[color] || logoColors[LOGO_COLOR.DEFAULT] || LOGO_COLOR.DEFAULT;
   const textColor =
     textColors[color] || textColors[TEXT_COLOR.DEFAULT] || TEXT_COLOR.DEFAULT;
+
+  console.log({ fundingSource, logoColor, textColor, color });
 
   const { Label, WalletLabel, Logo, showWalletMenu } = fundingConfig;
 
--- a/src/ui/buttons/button.jsx
+++ b/src/ui/buttons/button.jsx
@@ -148,8 +148,6 @@ export function Button({
   const textColor =
     textColors[color] || textColors[TEXT_COLOR.DEFAULT] || TEXT_COLOR.DEFAULT;
 
-  console.log({ fundingSource, logoColor, textColor, color });
-
   const { Label, WalletLabel, Logo, showWalletMenu } = fundingConfig;
 
   const clickHandler = (event, opts) => {


src/ui/buttons/button.jsx
```diff
@@ -138,6 +148,8 @@ export function Button({
   const textColor =
     textColors[color] || textColors[TEXT_COLOR.DEFAULT] || TEXT_COLOR.DEFAULT;
 
+  console.log({ fundingSource, logoColor, textColor, color });
```
imbrian, line 151: Did you mean to leave this in?

src/ui/buttons/button.jsx
```diff
@@ -111,9 +111,19 @@ export function Button({
...
+
+    // if multiple buttons are being rendered (smart stack), we set default color as gold > first
+    if (multiple) {
+      color = "gold";
+    }
+  }
+
+  // validate the first button rendered has a valid color
+  // this check is needed to validate the first button in a smart stack gets the correct color
+  if (i === 0 && !colors.includes(color)) {
```
imbrian, line 125: Oh nice. This `i` prop is unfortunately named, but used a lot here at this point.

src/ui/buttons/button.jsx
```diff
@@ -111,9 +111,19 @@ export function Button({
 
   // if no color option is passed in via style props
   if (color === "" || typeof color === "undefined") {
-    // if multiple buttons are being rendered (smart stack), we set default color as gold
     // if a single button is rendered, we set color to first option in the fundingSource config
-    color = multiple ? "gold" : colors[0];
+    color = colors[0];
+
+    // if multiple buttons are being rendered (smart stack), we set default color as gold > first
```
imbrian, line 117: Does this assumption break with any of the other buttons? I feel like it's something we could accidentally break with a different button in the future.

Is the order of colors in the config too intertwined with other logic to prevent switching around the order to put the default first? Possibly risky given the lack of test coverage around this.

imbrian (commented):
logic for handling this lgtm, but I'm still concerned about the lack of unit tests around this. It's definitely unit-testable without VRT, but if we don't have Jest/vitest set up to run for these `ui/` files, might be a separate ticket.

wsbrunson (approved):
you think its possible to make a function called `getButtonColors` that would return `{color, textColor, logoColor}` and then we can test that separately from the button render itself?

no need to do it with this PR, we can create a jira ticket and put it in the backlog

imbrian (approved):

src/ui/buttons/button.jsx
nikrom17: When `multiple === true` , we will use the `secondaryColor` in the funding config to decided the color of the button in the smart stack.

i.e. when you pass in `color: gold`, the PayPal button & PayLater button will actually be gold, but the Venmo button gets set to `blue`, Credit gets set to `darkblue` and Card gets set to `black`

I think a lot of the confusion lies around the `secondaryColors`. We can't simply set the default color to `gold` since not all buttons have `gold` as a valid color, but we need `gold `to be set as the default color to correctly look up the color to render in a button in the smart stack using the `secondaryColors['gold']`

The bug was caused by `color === 'gold'` on a button that does not have `gold` as an available color. If we can be sure we're not going to run into this scenario, we should be good to go. Then we can come back and refactor this logic once we have solid testing in place for validating the correct class is applied to the buttons.

src/ui/buttons/button.jsx
nikrom17: yes, it represents the position of the buttons in the smart stack. I would be for updating the prop name, maybe `buttonPosition` or `buttonIndex`

nikrom17:
Ticket for adding `getButtonColor` function: https://paypal.atlassian.net/browse/DTPPCPSDK-2678

ravishekhar (approved):

PR #2427, opened 2024-08-21
fix(hosted-buttons): fix an issue where shipping callbacks would not throw

rjstires:
The result of calling actions.reject was not being returned, causing the callback to not reject when the consumer passed an action.reject that rejects.

### Description
The existing onShippingChange and onShippingOptionsChange callbacks used by hosted buttons do not return the result of calling the client application supplied action.reject method. This allows execution to continue and the callback resolves as if it were successful.

This causes the need for unsightly and hard to test error handling scenarios in the client application.

(ticket incoming) Supporting NCPS shipping callbacks.

Passing `{ reject: message => Promise.reject(message) }` when calling either callback from the client application and the callback will not reject. See original test and changes for confirmation.

❤️ Thank you!

### DIFF
--- a/src/hosted-buttons/utils.test.js
+++ b/src/hosted-buttons/utils.test.js
@@ -636,10 +636,20 @@ describe("buildHostedButtonOnShippingAddressChange", () => {
         status: 500,
       })
     );
-    // $FlowIssue
-    await onShippingAddressChange(data, actions);
-    expect(actions.reject).toHaveBeenCalledWith(errors.ADDRESS_ERROR);
-    expect.assertions(1);
+
+    const localActions = {
+      // eslint-disable-next-line compat/compat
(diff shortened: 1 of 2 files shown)


src/hosted-buttons/utils.test.js
```diff
@@ -636,10 +636,20 @@ describe("buildHostedButtonOnShippingAddressChange", () => {
         status: 500,
       })
     );
-    // $FlowIssue
-    await onShippingAddressChange(data, actions);
```
rjstires, line 640: The test is titled, ""throws error when shipping-options api doesn't return 200" yet in the original test the callback does not throw. The test only verifies that the actions.reject method was called.

wsbrunson (approved):

imbrian (approved):

ravishekhar (approved):

PR #2440, opened 2024-10-01
Check formatting 👾

spencersablan:
### Description

_**✅ Check formatting in CI**_ - Noticed that we weren't verifying that PR's match our expected formatting before merging. Thought it might be good to do so.

_**👨‍🔧 Fix formatting**_ -  There was some formatting that was off.

❤️ Thank you!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,6 +29,9 @@ jobs:
         with:
           useLockFile: false
 
+      - name: ▶️ Check Formatting
+        run: npm run format:check
+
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -29,6 +29,9 @@ jobs:
         with:
           useLockFile: false
 
+      - name: ▶️ Check Formatting
```
wsbrunson, line 32: I'm surprised this isn't in the `lint` task, we should probably put it there

PR #2447, opened 2024-10-16
Defining ThreeDomainSecureClient component interface

mchoun:
### Description

This PR is to add the framework for the `ThreeDomainSecureClient` component. Two additional methods of `isEligible` and `show` have been added to this component. 

https://paypal.atlassian.net/browse/DTPPCPSDK-2658

image

❤️ Thank you!


src/three-domain-secure/component.jsx
```diff
@@ -1,30 +1,52 @@
...
+  }
+
+  logger.info("three domain secure v2 invoked").track({
+    [FPTI_KEY.TRANSITION]: "three_DS_auth_v2",
+  });
+
+  return sdkConfig;
 };
+export interface ThreeDomainSecureComponentInterface {
+  isEligible(): void;
```
imbrian, line 25: nit: do we want this interface to go ahead and reflect the future state by returning a boolean? Then could update line 43 to just always return `false` for now. The test for it would then be closer to the design too (and not test for a mocked `console.log`.

mchoun: This is a good call out. I modified the interface for both `isEligible()` and `show()` to better reflect the future state.

imbrian (approved):

src/zoid/card-fields/component.jsx
```diff
@@ -48,11 +48,11 @@ const CARD_FIELD_TYPE = {
...
-  financingCountryCode : string,
-  currencyCode : string,
-  billingCountryCode : string,
-  amount : string,
-  includeBuyerInstallments ? : boolean
+  financingCountryCode: string,
+  currencyCode: string,
+  billingCountryCode: string,
+  amount: string,
+  includeBuyerInstallments?: boolean,
```
siddy2181, line 55: nit: do we want to remove the lint updates out of this PR for clarity?

src/three-domain-secure/component.test.js
```diff
@@ -1,44 +1,59 @@
...
+    warn: vi.fn().mockReturnThis(),
+    error: vi.fn().mockReturnThis(),
     track: vi.fn().mockReturnThis(),
-    flush: vi.fn().mockReturnThis(),
-  })),
-}));
-vi.mock("../lib", () => ({
-  ValidationError: vi.fn(),
-}));
-describe("getThreeDomainSecure returns ThreeDomainSecureComponent", () => {
```
siddy2181, line 20: Do we still want to keep this test?

src/three-domain-secure/component.jsx
```diff
@@ -1,30 +1,54 @@
...
-    } else {
-      throw new ValidationError(
-        `script data attribute sdk-client-token is required but was not passed`
-      );
-    }
-  };
-
-  return ThreeDomainSecureAuth;
+type SdkConfig = {|
+  sdkToken: ?string,
```
siddy2181, line 10: nit: We can probably remove the `?` here. It is required to be a string

siddy2181 (commented):
Looks Great!!!

src/zoid/card-fields/component.jsx
mchoun: Good call. I can create a separate PR for the lint commits.

src/three-domain-secure/component.test.js
mchoun: I moved this test further down on Line 46 as a general initialization test. Instead of mocking the ValidationError I'm just asserting that it would throw an error. Is this OK or does it change the test? I wasn't entirely sure.

src/three-domain-secure/component.jsx
mchoun: Agreed!

mchoun: Removing this creates a flow error. I believe that it is possible for null or undefined to be returned by `getSDKToken` which is why it should be optional

src/zoid/card-fields/component.jsx
siddy2181: We can leave it in for this one. But also the comment, so it is easier to trace back that it is simply a lint update and not actual code changes.

src/three-domain-secure/component.test.js
siddy2181: I see, yeah that should work too. My idea with that test was to capture `validationError` specifically since it will be one of the trace events in the future. 
This is not a blocker for the PR to be merged

PR #2450, opened 2024-10-29
Add validation for SDK token when createSubscription is being passed

9zees:
### Description

When the Unbranded CardFields is used for Subscriptions, we require an SDK token data-attribute to be passed, similar to 3DS Component. This PR adds a validation for this token when a createSubscription callback has been passed. This validation should not be invoked when a createSubscription callback is not passed.

Acceptance Criteria:
1. A validation error is thrown when attempting to use Card Fields subscriptions without the SDK token.
2. A validation error is not thrown when attempting to use Card Fields without subscriptions when an SDK token is not provided.

### NOTE: I was intending on using Zoid's validate() method but not able to get it invoked despite of following the documentation. Leaving this here as proof that I did attempt to use this method.

There are 2 ways we can validate on Zoid.
**(1) Use validate() function in zoid.create(...)**
https://krakenjs.com/zoid/docs/api/create.html#validate--props---void

image

**(2) Use validate() function in individual props**
https://krakenjs.com/zoid/docs/api/prop-definitions.html#validate

image

I'm choosing the prop-level validation because we only want to check for createSubscription (prop-level) as opposed to running the validate method each time on all the props.

=================== End of Note ====================

https://paypal.atlassian.net/browse/DTPPCPSDK-2628

Use-Case 1: When we pass in SdkToken and createSubscription:
image

Use-Case 2: When we don't pass SdkToken but pass createSubscription:
image

Use-Case 3: When we don't pass SdkToken and don't pass createSubscription (but pass createOrder instead) :
image

❤️ Thank you!


Redrum1228 (approved):

imbrian (approved):

src/zoid/card-fields/component.jsx
```diff
@@ -35,7 +35,7 @@ import {
   type FundingEligibilityType,
 } from "@paypal/sdk-constants/src";
 
-import { getSessionID } from "../../lib";
+import { getSessionID, ValidationError } from "../../lib";
```
Redrum1228, line 38:
```suggestion
import { getSessionID, ValidationError } from "../../lib";
```

Redrum1228 (approved):

Redrum1228 (approved):

Redrum1228:
New to coding . What's the easiest way to use the program

PR #2451, opened 2024-10-31
add eligibility check and protect export for stage and local

mchoun:
### Description

This PR is to add the `isEligible()` method to the `ThreeDomainSecureClient`. This method will allow the merchant to make a request to determine if the any payer action is required for 3DS authentication. If a payer action is required, it will obtain the authentication URL and make it available to the `show()` method for the buyer to authenticate. 

JIRA Ticket - DTPPCPSDK-2660

❤️ Thank you!


src/lib/security.js
```diff
@@ -28,3 +29,13 @@ export function allowIframe(): boolean {
 export const protectedExport = (unprotectedExport) =>
   isPayPalDomain() ? unprotectedExport : undefined;
 /* eslint-enable no-confusing-arrow */
+
+// $FlowIssue
+export const localOrStageExport = (unprotectedExport) => {
```
ravishekhar, line 34: Can we rename it to `devEnvOnlyExport` 🤔 ?

src/three-domain-secure/component.jsx
```diff
@@ -23,29 +86,81 @@ const parseSdkConfig = ({ sdkConfig, logger }): SdkConfig => {
...
-    return new ZalgoPromise((resolve) => {
-      resolve(false);
-    });
+  async isEligible(merchantPayload: MerchantPayloadData): Promise<boolean> {
+    const data = parseMerchantPayload({ merchantPayload });
+    try {
+      // $FlowFixMe
+      const { status, links } = await this.request<requestData, responseBody>({
+        method: "POST",
+        url: `${this.sdkConfig.paypalApiDomain}/v2/payments/payment`,
```
ravishekhar, line 146: I feel we should keep all the endpoints in a single const file.

ravishekhar (approved):
Left a few trivial comments. Looks great otherwise.

src/three-domain-secure/component.jsx
mchoun: Done!

src/lib/security.js
mchoun: Much better name :)

Dlbandy04 (approved):

PR #2452, opened 2024-11-08
feature(3DS): Implement show method and return response with enriched nonce

siddy2181:
### Description

The ThreeDomainSecureClient component provides a method to the merchant to display the 3DS challenge to the buyer if the response from the `isEligible` method is true. After the buyer has completed the 3DS challenge, the `liabilityShift`, `authenticationState`, and a`nonce` with enriched 3DS data will be returned to the merchant.

Code Sample: 
```javascript
const submitButton = document.getElementById("submit-button");

submitButton.addEventListener("click", async () => {
  if (isThreeDomainSecureEligible) {
    const {
      liabilityShift, // "no", "unknown", "possible"
      authenticationState, // "succeeded", "cancelled", "errored"
      nonce, //Enriched nonce or the original nonce
    } = await threeDomainSecureComponent.show();
    
    if (authenticationState === "succeeded") {
      // Check the liability shift and decide on continuing the transaction
    } else {
      // Cancelled or errored, merchant needs to decide what they want to check on
    }
  }
})
```

This PR covers the following tickets

DTPPCPSDK-2662
DTPPCPSDK-2665
DTPPCPSDK-2805
DTPPCPSDK-2855

❤️ Thank you!


Dlbandy04:
[]()

__sdk__.js
```diff
@@ -94,6 +94,8 @@ module.exports = {
     entry: "./src/shopper-insights/interface",
   },
   "three-domain-secure": {
+    globals,
+    automatic: true,
```
mchoun, line 98: We don't need to `automatic: true` anymore, correct? I believe we were testing this when we were running into the zoid xprops issue.

src/lib/security.js
```diff
@@ -39,3 +39,6 @@ export const devEnvOnlyExport = (unprotectedExport) => {
     return undefined;
   }
 };
+// TODO: Remove after testing
+// $FlowIssue
+export const payPayDomainRegEx = /\.paypal\.(com|cn)(:\d+)?$/; // eslint-disable-line security/detect-unsafe-regex
```
mchoun, line 44: We should remove this and use the `getPayPalDomainRegex()` method that sdk-client provides.

src/three-domain-secure/component.jsx
```diff
@@ -115,47 +60,75 @@ const parseMerchantPayload = ({
...
+
     try {
       // $FlowFixMe
-      const { status, links } = await this.request<requestData, responseBody>({
+      const { status, links } = await this.restClient.request<
+        requestData,
+        responseBody
+      >({
         method: "POST",
-        url: `${this.sdkConfig.paypalApiDomain}/${PAYMENT_3DS_VERIFICATION}`,
```
mchoun, line 147: Are we OK with keeping this a constant from `src/constant/api.js`? It wasn't clear why this was changed.

src/three-domain-secure/utils.jsx
```diff
@@ -0,0 +1,105 @@
...
+            focus={focus}
+            event={event}
+            frame={frame}
+            prerenderFrame={prerenderFrame}
+            content={props.content}
+            nonce={props.nonce}
+          />
+        ).render(dom({ doc }));
+      },
+      domain: payPayDomainRegEx,
```
mchoun, line 52: We should remove this and use the getPayPalDomainRegex() method that sdk-client provides.

src/three-domain-secure/component.jsx
```diff
@@ -115,47 +60,75 @@ const parseMerchantPayload = ({
 };
 
 export interface ThreeDomainSecureComponentInterface {
-  isEligible(): Promise<boolean>;
-  show(): ZoidComponent<void>;
+  isEligible(payload: MerchantPayloadData): Promise<boolean>;
+  show(): ZalgoPromise<threeDSResponse>;
 }
+
+type Update3DSTokenResponse = {|
```
mchoun, line 67: We should move this to `./types.js`

src/three-domain-secure/component.jsx
```diff
@@ -115,47 +60,75 @@ const parseMerchantPayload = ({
...
     this.logger = logger;
-    this.request = request;
+    this.restClient = restClient;
+    this.graphQLClient = graphQLClient;
     this.sdkConfig = parseSdkConfig({ sdkConfig, logger });
   }
 
   async isEligible(merchantPayload: MerchantPayloadData): Promise<boolean> {
     const data = parseMerchantPayload({ merchantPayload });
+    this.fastlaneNonce = merchantPayload.nonce;
```
mchoun, line 103: There should be some validation on the nonce or the merchant payload. Ran into an issue where undefined was being passed to the GQL call. I believe this is going to be addressed by DTPPCPSDK-2861

__sdk__.js
mchoun: `automatic: true` will return this component for everyone that loads the JS SDK. The exported [setup()](https://github.com/paypal/paypal-checkout-components/pull/2452/files#diff-c310050612136b5f5121d74da169666e0bf8d687f51c7275de09e3a4bafef6deR24) method is what will automatically create the zoid component when the component is loaded. 

This will probably be fine for now since this component is still protected but we should remove this before we allow this in Production. If we don't, `paypal.ThreeDomainSecureClient` will be available to everyone even if they don't pass `three-domain-secure` into the components query param.

wsbrunson: we should probably remove it now because even though its hidden, it will show up in all bundles served. I'm sure its super small but might as well not ship javascript if we don't have to

siddy2181: That makes sense. I've removed that flag. Thanks both for clarifying :)

wsbrunson (approved):

src/three-domain-secure/component.test.js
```diff
@@ -113,24 +138,65 @@ describe("three domain secure component - isEligible method", () => {
...
       threeDomainSecureClient.isEligible(defaultMerchantPayload)
     ).rejects.toThrow(new Error("Error with API"));
-    expect(mockRequest).toHaveBeenCalled();
+    expect(mockRestClient.request).toHaveBeenCalled();
   });
 });
 
-describe("three domain descure component - show method", () => {
-  test.todo("should return a zoid component", () => {
+describe.todo("three domain descure component - show method", () => {
```
imbrian, line 154: this is still `todo`. is this intentional?

imbrian, line 154: this is still `todo`. is this intentional?

wsbrunson: oh wow i didn't know you could do that

wsbrunson (approved):

imbrian (approved):

PR #2461, opened 2025-01-10
update resume flow return value parameter names

ravishekhar:
### Description

As per latest discussions the return URL would be as below
```
https://example.com/redirectUrl#onApprove?token=ABCDED&PayerID=SDFFF23232
```
This PR updates the return url parsing logic as per above contract. 

❤️ Thank you!

### DIFF
--- a/src/lib/appSwitchResume.js
+++ b/src/lib/appSwitchResume.js
@@ -1,5 +1,6 @@
 /* @flow */
 import { FUNDING } from "@paypal/sdk-constants/src";
+import { parseQuery } from "@krakenjs/belter/src";
 
 import { APP_SWITCH_RETURN_HASH } from "../constants";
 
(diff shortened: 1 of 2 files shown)


src/lib/appSwitchResume.js
```diff
@@ -16,46 +17,47 @@ export type AppSwitchResumeParams = {|
...
-      // $FlowIgnore[incompatible-type]
-      checkoutState: urlHash,
-    };
-    return params;
-  }
-  return null;
+
+  const {
+    token,
+    PayerID,
```
wsbrunson, line 34: oh they are doing `PayerID` and not `payerID`?

wsbrunson: I'll have to do a new PR too

PR #2464, opened 2025-01-21
feat: update node version to 18

mchoun:
### Description

Vitest is breaking due to old Node 16.18 version. This is to update node version and certain scripts to play nicely with 18. 

We also had to bump the maxSize of the bundle by 1kb despite no changes other than the node version. Validated that there were no changes in the bundle. 

❤️ Thank you!


wsbrunson (approved):

ravishekhar (approved):
Thanks for fixing this. Lets mark an action item to revisit this increase in bundle size.

PR #2466, opened 2025-01-28
refactor: update component name, match success state response

siddy2181:
- The ThreeDS Zoid component does not have any fastlane related logic, so a generic name is more suitable.
- No breaking change
- Fixes the response state from ~`success`~ to `succeeded` to match the integration guide pattern

❤️ Thank you!


mchoun (approved):
Looks good!

imbrian (approved):

PR #2467, opened 2025-01-29
redirect method in button props

ravishekhar:
### Description
Add a new redirect method to button props which can be used by Buttons to redirect

❤️ Thank you!


wsbrunson (approved):
nice

wsbrunson (approved):

PR #2471, opened 2025-02-12
add timing data to xprops

wsbrunson:
add sdk init time from `sdk-client` as well as the `duration` property from the first sdk script loaded on the page.

We will use this information in the buttons iframe to mark sdk init end and also report on the duration of the sdk load time on the merchant's page

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -43,6 +43,7 @@ import {
   getJsSdkLibrary,
   wasShopperInsightsUsed,
   isPayPalTrustedUrl,
+  getSDKInitTime,
 } from "@paypal/sdk-client/src";
 import {
   rememberFunding,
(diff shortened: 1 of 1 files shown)


src/zoid/buttons/component.jsx
```diff
@@ -1067,6 +1068,47 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+
+          const isCached = (performanceEntry) => {
+            if (
+              !performanceEntry ||
+              typeof performanceEntry.duration === "undefined"
+            ) {
+              return "unknown";
+            }
+
+            return performanceEntry.duration === 0 ? "yes" : "no";
```
imbrian, line 1090: Is this `yes/no` standard for existing performance queries that we have?

wsbrunson: its what we did in nextgen. normally I would have done `true/false` the presence of `unknown` made me go a different direction. 

I'm not sure of any precedent for this type of thing, I'm open to suggestions

ravishekhar (approved):
lgtm

PR #2484, opened 2025-04-01
Add PayPal App Switch Overlay

nikrom17:
### Description

We need a new overlay for the new flow PayPal App Switch

https://paypal.atlassian.net/browse/DTPPCPSDK-2897

https://github.com/user-attachments/assets/a02d655e-7d8c-4a77-9edf-0dafec31b4dd

❤️ Thank you!


src/ui/buttons/props.js
```diff
@@ -501,6 +501,15 @@ export type ButtonExtensions = {|
   resume: () => void,
 |};
 
+type ShowPayPalAppSwitchOverlay = {|
```
ravishekhar, line 504: we also need to accept the universal link to render the continue link to app.

ravishekhar (commented):
looks good apart from the one comment 👍

src/ui/buttons/props.js
nikrom17: I've added that logic in the focus function in `setupOverlays`

I can update the focus function to take the args `props.redirect` & `url.href` and call `props.redirect(url.href)` from inside the overlay component

nikrom17: https://github.paypal.com/Checkout-R/smartcomponentnodeweb/pull/930/files#diff-5706d68d378768a97412d3ee5d456d81d91fee49a759dd73ce9d6fef2da324e3R322

wsbrunson (approved):

test/integration/tests/button/paypalAppSwitchOverlay.js
```diff
@@ -0,0 +1,215 @@
...
+                if (getElementRecursive(".paypal-checkout-sandbox")) {
+                  done(
+                    new Error(
+                      "Expected overlay to be removed from dom after close was called"
+                    )
+                  );
+                }
+              } catch {
+                // an error will be thrown if the overlay is not found, which means overlay was removed successfully
+                done();
```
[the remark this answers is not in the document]
nikrom17: `getElementRecursive(".paypal-checkout-sandbox")` will return an error if it does not find what it is looking for. For this test, I am expecting no overlay to be found so I want the done condition to be called in the catch statement.

src/ui/buttons/props.js
nikrom17: @ravishekhar , I attempted a solution that took the url as an argument and ran into issues running the karma tests with the redirect call now being a part of `overlay.jsx`

With the Karma framework, there is no way to mock `location.href ` or `location.redirect`. When the redirect is called, the karma test runner is interrupted and the tests fail.

ravishekhar (approved):
Based on offline discussions we will be taking up the changes to refactor the postMessage passing between SDK and Buttons as a followup PR.

PR #2485, opened 2025-04-07
Feature/disable max height new fix

delbarrow:
### Description

This feature adds new css responsive styling based on the height of the button when disable max height is true.
The current responsive behavior of disableMaxHeight is not ideal since the height of the logo and label are based off the width of the button instead of the height. 
To achieve this a new function, generateDisableMaxHeightStyles, was created which generates the new styling for disable max height based on height breakpoints set in, BUTTON_DISABLE_MAX_HEIGHT_STYLE. This new styling is appended below the previous styling.

[https://paypal.atlassian.net/browse/DTPPCPSDK-2986]

Before:
Before
XXXL Before

After:
After
XXXL After

Edge Cases to consider:
Specific styling for ApplePay, EPS, and MyBank


src/ui/buttons/styles/styleUtils.js
```diff
@@ -0,0 +1,206 @@
...
+import type { Experiment } from "../../../types";
+import { BUTTON_DISABLE_MAX_HEIGHT_STYLE, BUTTON_SIZE_STYLE } from "../config";
+import { BUTTON_SIZE, BUTTON_DISABLE_HEIGHT_SIZE } from "../../../constants";
+
+const BUTTON_MIN_ASPECT_RATIO = 2.2;
+const MIN_SPLIT_BUTTON_WIDTH = 300;
+
+const WALLET_BUTTON_PERC = 60;
+
+function getLabelHeight({
```
nikrom17, line 15: Let's add unit tests for all these helper functions as well.

src/ui/buttons/styles/styleUtils.test.js
```diff
@@ -427,6 +617,83 @@ describe("test responsive style variables for disable max height", () => {
   );
 });
 
+describe("test rebrand responsive style variables when shouldResizeLable == true for disable max height", () => {
```
nikrom17, line 620: Let's also add a similar describe block but for `getResponsiveStyleVariables`

nikrom17:
Screenshot 2025-04-11 at 4 25 35 PM

We need to think about this edge cause when we do the full redesign. The min width of `150px` is not big enough to fit the 3XL button. The legacy logo fits just fine so we should be good to merge this in and address it in this [ticket](https://paypal.atlassian.net/browse/DTPPCPSDK-2987)

PR #2494, opened 2025-05-06
set hasShippingCallback in checkout web url if shipping callbacks are used

ravishekhar:
### Description
* pass `hasShippingCallback` param to checkout web if a shipping callbacks are used by merchant 

❤️ Thank you!


src/zoid/checkout/component.jsx
```diff
@@ -283,6 +283,19 @@ export function getCheckoutComponent(): CheckoutComponent {
           required: false,
         },
 
+        hasShippingCallback: {
```
nikrom17, line 286: Would it be possible to use the value returned from props?
https://github.com/paypal/paypal-checkout-components/blob/b61109fa2d90bee48dc9215d580620c2b7562e4a/src/ui/buttons/props.js#L1040C5-L1040C24

ravishekhar: Yes we can use, but then we would need SPB to pass this prop during initialization of the `Checkout({....})` zoid component. I was trying keep this change in checkout-components as it is just a derived property.

imbrian (approved):

PR #2520, opened 2025-07-29
Revert "chore(release): 5.0.373 🎉"

jtanya17:
Reverts paypal/paypal-checkout-components#2518

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@paypal/checkout-components",
-  "version": "5.0.373",
+  "version": "5.0.372",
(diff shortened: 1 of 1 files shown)


nikrom17:
We can leave in this PR and just publish `374` with the changes from this revert [PR](https://github.com/paypal/paypal-checkout-components/pull/2519)

If we undo the version change, we will run into issues running the release script and trying to publish the same version to NPM.

Once something is published to NPM its best to just push an update vs trying to undo or delete the version

jtanya17:
We won't be needing to revert this change anymore.

PR #2521, opened 2025-07-29
Adding funding source to button's iframe title to make each title unique

gusveloso:
### Description
Making each button title attribute unique using funding source information

[LI-96757: [Adyen] Accessibility issues reported by merchant Acer](https://paypal.atlassian.net/browse/LI-96757)
We're making each button iframe title attribute unique in order to stay compliant to accessibility guidelines.

- Run the project locally
- Inspect any button
- On the outer iframe element, you will see the title as "Paypal-<funding-source>

Screenshot 2025-07-29 at 18 34 38 1

❤️ Thank you!

### DIFF
--- a/src/zoid/buttons/component.jsx
+++ b/src/zoid/buttons/component.jsx
@@ -219,13 +219,13 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
       ).render(dom({ doc }));
     },
 
-    attributes: {
+    attributes: ({ props }) => ({
       iframe: {
         allowpaymentrequest: "allowpaymentrequest",
         scrolling: "no",
-        title: FUNDING_BRAND_LABEL.PAYPAL,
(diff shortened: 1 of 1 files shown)


src/zoid/buttons/component.jsx
```diff
@@ -219,12 +219,19 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+      let fundingSource = "";
+      if (props.fundingSource) {
+        fundingSource = props.fundingSource;
+      }
+
+      return {
+        iframe: {
+          allowpaymentrequest: "allowpaymentrequest",
+          scrolling: "no",
+          title: `${FUNDING_BRAND_LABEL.PAYPAL}-${fundingSource}`,
```
delbarrow, line 232: Looks great, my only comment would be if the funding source is undefined, like in the smart stack, then the title would be "PayPal-". Maybe we want to add the "-" to the funding source on line 225 instead so that it is only added if the funding source is defined.

delbarrow (approved):

PR #2545, opened 2025-08-27
fix(venmo): Makes the prescense of poupBridge required for webview eligibility

mynormh:

❤️ Thank you!

### DIFF
--- a/src/funding/venmo/config.jsx
+++ b/src/funding/venmo/config.jsx
@@ -48,8 +48,9 @@ export function getVenmoConfig(): FundingSourceConfig {
 
     requires: ({ experiment, platform }) => {
       const isNonNativeSupported =
-        experiment?.venmoEnableWebOnNonNativeBrowser === true ||
-        (__WEB__ && window.popupBridge);
+        experiment?.venmoEnableWebOnNonNativeBrowser === true &&
+        __WEB__ &&
+        window.popupBridge;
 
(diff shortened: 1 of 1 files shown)


mnicpt (approved):

src/zoid/buttons/component.jsx
```diff
@@ -1266,13 +1272,23 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
...
+          return isSupportedNativeBrowser();
+        },
         queryParam: true,
       },
 
       supportsPopups: {
         type: "boolean",
-        value: () => userAgentSupportsPopups(),
+        value: ({ props }) => {
+          if (props?.fundingSource === FUNDING.VENMO) {
```
ravishekhar, line 1287: This method will be called during button render. I am bit concerned if the fundingSource will be available during render?

src/zoid/buttons/component.jsx
```diff
@@ -1266,13 +1275,28 @@ export const getButtonsComponent: () => ButtonsComponent = memoize(() => {
 
       supportedNativeBrowser: {
         type: "boolean",
-        value: isSupportedNativeBrowser,
+        value: ({ props }) => {
+          if (props.fundingSource === FUNDING.VENMO) {
```
ravishekhar, line 1279: will we ever hit this condition as the props are passed down to buttons iframe during render 🤔 ?

PR #2546, opened 2025-09-03
[Feature]/DTXOCOEXP-1400 Add `verificationMethod` Parameter

jtanya17:
### Description

Screenshot 2025-09-23 at 8 47 15 AM

/DTXOCOEXP-1400

❤️ Thank you!


src/three-domain-secure/component.jsx
```diff
@@ -34,21 +34,34 @@ const parseSdkConfig = ({ sdkConfig, logger }): SdkConfig => {
...
+    transactionContext,
+  } = merchantPayload;
+
+  let verificationMethod = "SCA_WHEN_REQUIRED";
+
+  if (threeDSRequested !== undefined) {
+    verificationMethod = threeDSRequested ? "SCA_ALWAYS" : "SCA_WHEN_REQUIRED";
+  } else if (threeDSTriggerMode) {
+    verificationMethod = threeDSTriggerMode;
+  }
```
AleGastelum, line 57: Checking `threeDSRequested` before `threeDSTriggerMode` is making the `should prioritize threeDSTriggerMode over threeDSRequested` test fail. Changing the order switches the priority to `threeDSTriggerMode` and makes the test pass. Would something like this work with the rest of the implementatino?

```suggestion
  if (threeDSTriggerMode) {
    verificationMethod = threeDSTriggerMode;
  } else if (threeDSRequested !== undefined) {
    verificationMethod = threeDSRequested ? "SCA_ALWAYS" : "SCA_WHEN_REQUIRED";
  } 
```

src/three-domain-secure/component.jsx
```diff
@@ -147,12 +159,14 @@ export class ThreeDomainSecureComponent {
         payerActionUrl: this.authenticationURL,
         onSuccess: async (res) => {
           const { reference_id, liability_shift, success } = res;
+          // $FlowFixMe incompatible type payload
+          this.logger.info("helios_response", JSON.stringify(res));
           let enrichedNonce;
           // Helios returns a boolen parameter: "success"
           // It will be true for all cases where liability is shifted to merchant
-          // and false for downstream failures and errors
+          // and false for downstream failures and errors where liability_shift= NO.
```
AleGastelum, line 167: Could the `liability_shift` value be `unknown` for a `false` success response?

src/three-domain-secure/component.jsx
siddy2181: Actually that is what @jtanya17  had originally. I'll revert my last commit.

delbarrow (approved):

src/three-domain-secure/component.jsx
siddy2181: Adding the explanation here:
There are 2 aspects to and error here, it
1.) Error occured while sending the response to SDK -> No `liability_shift`
2.) Error occured when Helios received a message from their downstream, `NO | UNKNOWN`
In either case, the guidance to merchant is the same. They can call the get API and fetch the status and not rely on the response SDK gives.

AleGastelum: Thanks for the explanation and for the change in the comment. That makes it more clear!

PR #2624, opened 2026-04-17
feature(SavedPaymentMethods): Adding loading template for prerender

seraphD:
### Description

Adding a reusable loading template for prerender

[jira](https://paypal.atlassian.net/browse/DTBILLMOD-1619)

https://github.com/user-attachments/assets/8248dce9-626e-45ef-a676-443f267ccd44

❤️ Thank you!


src/ui/saved-payment-methods/template.jsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/**
+ * Server-only render (commonjs2 / web: false) — mirrors smart buttons `Buttons()`.
+ * Non-interactive loading shell until the iframe client app hydrates.
+ */
+export function SavedPaymentMethods(
+  props: SavedPaymentMethodsServerRenderProps
+): ElementNode {
+  return SavedPaymentMethodsLoadingBody({
```
kolman, line 25: This component has no functionality, it only renders another component. Consider removing `SavedPaymentMethodsLoadingBody` and moving the code here.

src/ui/saved-payment-methods/loading-body.jsx
```diff
@@ -0,0 +1,66 @@
...
+  const shouldRenderMessageContainer = Boolean(styleConfig.layout.message);
+
+  return (
+    <Fragment>
+      <style nonce={nonce}>
+        {`
+            body {
+              margin: 0;
+            }
+            ${styles}
```
kolman, line 35: Styles are strongly coupled to the HTML structure rendered here. Because we don't have to share the styles across projects (as before), there's no need to have them apart in `getStyles` function. Consider moving the styles here, it's easier to understand when we can see CSS and HTML in one place.

src/zoid/saved-payment-methods/util.js
```diff
@@ -34,6 +62,184 @@ const STYLE_LAYOUT_FIELD_TYPES: {|
   iconSize: "string",
 };
 
+export const getStyleConfig = (style) => ({
```
kolman, line 65: This is a flow-typed file. Please add type annotations.

src/zoid/saved-payment-methods/util.js
```diff
@@ -34,6 +62,184 @@ const STYLE_LAYOUT_FIELD_TYPES: {|
...
+    ...(style.layout || {}),
+  },
+});
+
+export const getStyles = (styleConfig) =>
+  `
+  * {
+    box-sizing: border-box;
+  }
+    .saved-payment-methods-container {
```
kolman, line 85: This component will be always rendered inside of an `iframe` with `paypal.com` origin. We don't have to prefix every CSS class to make it safe. If you still want to make the relation obvious, I recommend using a shorter, abbreviated prefix, like `.spm-`.

Note that above we're using `*` selector. We don't need to use the prefix, this is a de-facto global stylesheet for our iframes.

src/zoid/saved-payment-methods/util.js
```diff
@@ -34,6 +62,184 @@ const STYLE_LAYOUT_FIELD_TYPES: {|
...
+
+  /* Preact mounts into this node; avoid an extra flex item / box alongside logo + label. */
+  #saved-payment-methods-app-root {
+    display: contents;
+  }
+
+  /* If both skeleton and button ever coexist, collapse the skeleton so it takes no space. */
+  #saved-payment-methods-app-root:has(.saved-payment-methods-tag)
+    > .saved-payment-methods-tag-loading {
+    display: none !important;
```
kolman, line 142: Why `!important`? Tag specificity should work here. Also, why so many other CSS properties after setting `display: none`? It should be enough to turn it off.

src/zoid/saved-payment-methods/util.js
```diff
@@ -34,6 +62,184 @@ const STYLE_LAYOUT_FIELD_TYPES: {|
...
+    object-fit: contain;
+  }
+
+  /* Preact mounts into this node; avoid an extra flex item / box alongside logo + label. */
+  #saved-payment-methods-app-root {
+    display: contents;
+  }
+
+  /* If both skeleton and button ever coexist, collapse the skeleton so it takes no space. */
+  #saved-payment-methods-app-root:has(.saved-payment-methods-tag)
```
kolman, line 140: Using `:has()` may not be safe, as it is relatively new selector and we need to support older browsers.

src/zoid/saved-payment-methods/prerender.jsx
```diff
@@ -5,6 +5,7 @@ import { node, type ChildType } from "@krakenjs/jsx-pragmatic/src";
 import type { ZoidProps } from "@krakenjs/zoid/src";
 
 import { type SavedPaymentMethodsProps } from "./props";
+import { SavedPaymentMethodsLoadingBody } from "../../ui/saved-payment-methods/loading-body";
```
kolman, line 8: Prerender should be using `src/ui/saved-payment-methods/template.jsx` to make sure it's exactly the same component as the one used in SCNW.

kolman (changes requested):

seraphD:
已阅读你的邮件。我会尽快查看并回复。谢谢！有事联系18368380219--张立辉

src/ui/saved-payment-methods/template.jsx
seraphD: fixed

src/ui/saved-payment-methods/loading-body.jsx
seraphD: fixed moved to getStyleconfig function to the template and now using plain string now

src/zoid/saved-payment-methods/util.js
seraphD: moved to template and added type annotations

src/zoid/saved-payment-methods/util.js
seraphD: replaced with spm prefix

src/zoid/saved-payment-methods/util.js
seraphD: this class is not needed so removed

src/zoid/saved-payment-methods/util.js
seraphD: it's more like a safe net removing it

src/zoid/saved-payment-methods/prerender.jsx
seraphD: now all using template now

src/ui/saved-payment-methods/template.jsx
```diff
@@ -0,0 +1,254 @@
...
+            .spm-tag-edit-button svg path {
+              fill: ${styleConfig.root.primaryColor};
+            }
+
+            .spm-message {
+              height: 31px; /* Based on the error message in browser console */
+              margin-top: 6px;
+              ${
+                styleConfig.layout.logo
+                  ? /* TODO: Remove calc */ `
```
dwilliams27: Should fix TODO

src/ui/saved-payment-methods/template.jsx
```diff
@@ -0,0 +1,254 @@
...
+              width: ${styleConfig.layout.iconSize};
+              border-radius: 4px;
+            }
+
+            .spm-tag-edit-button svg path {
+              fill: ${styleConfig.root.primaryColor};
+            }
+
+            .spm-message {
+              height: 31px; /* Based on the error message in browser console */
```
dwilliams27, line 201: Hmm, why is this value based on error message in console?

src/ui/saved-payment-methods/template.jsx
```diff
@@ -0,0 +1,254 @@
...
+
+import { node, Fragment, type ElementNode } from "@krakenjs/jsx-pragmatic/src";
+
+import { validateSavedPaymentMethodsStyle } from "../../zoid/saved-payment-methods/util";
+
+import { PPRebrandLogo } from "@paypal/sdk-logos/src";
+
+export type SavedPaymentMethodsServerRenderProps = {|
+  nonce?: string,
+  style?: Object,
```
dwilliams27: Can we type this out more fully than just Object?

dwilliams27 (commented):
Few small comments, but overall looking good

src/ui/saved-payment-methods/template.jsx
kolman: This is what I was getting during the development. It is not documented though 🤷 
Image

src/ui/saved-payment-methods/template.jsx
seraphD: fixed

src/zoid/saved-payment-methods/component.jsx
```diff
@@ -98,7 +98,7 @@ export const getSavedPaymentMethodsComponent: () => SavedPaymentMethodsComponent
       autoResize: {
         width: false,
         height: true,
-        element: ".saved-payment-methods-container",
+        element: ".spm-container",
```
kolman: This can force us to release this library at the exact same time as our internal server impl, as currently the container div with this class is rendered there. We want to change it to use the template from this library, but until it happens, these two classes would be out of sync. Maybe it's a good idea to keep (just this) class name as it is (including in template.jsx)

src/ui/saved-payment-methods/template.jsx
seraphD: fixed

src/zoid/saved-payment-methods/component.jsx
seraphD: fixed, but does that mean we will need to create a follow up PR to fix this?

kolman (approved):
Looks good to me now. Great job!

src/zoid/saved-payment-methods/component.jsx
kolman: If we keep this class name as is, no further PR is necessary. Yes the prefix is not consistent, but I'm fine with that, given that this is the top-level component container. When we start to use this component internally for server-render, it will be possible to change it here because there will be just one source of truth.

gregjopa (approved):

gregjopa:
@seraphD please fix the linting errors causing the build to fail

seraphD:
> @seraphD please fix the linting errors causing the build to fail

fixed please check again

gregjopa (approved):

ravishekhar (approved):

PR #2630, opened 2026-04-28
feat: add variationName prop support for PayPal monogram marks

Vishala230531:
This PR introduces support for the [variationName] prop in PayPal Marks components, enabling developers to request specific PayPal mark variations (wordmark vs monogram) through the SDK API.

[https://paypal.atlassian.net/browse/DTPPCPSDK-6244](url)

 Features
Variation Support: Choose between wordmark (default PayPal logo) and monogram (PP mark)
Backward Compatible: Existing code continues to work unchanged (defaults to wordmark)

🏗️ Implementation Details
Variation Logic:

[variationName: "wordmark"] or undefined → Renders standard PayPal logo
[variationName: "monogram"]→ Renders PayPal monogram mark
Variation logic only applies to [FUNDING.PAYPAL], other funding sources unaffected

Screenshot 2026-04-28 at 3 06 23 PM

### DIFF
--- a/src/constants/button.js
+++ b/src/constants/button.js
@@ -119,8 +119,8 @@ export const MESSAGE_ALIGN = {
   RIGHT: ("right": "right"),
 };
 
-// Mark variation options for funding sources
-export const MARK_VARIATIONS = {
+// PayPal mark variation options
+export const PAYPAL_MARK_VARIATIONS = {
   WORDMARK: ("wordmark": "wordmark"),
   MONOGRAM: ("monogram": "monogram"),
 };
--- a/src/constants/button.js
+++ b/src/constants/button.js
@@ -119,8 +119,8 @@ export const MESSAGE_ALIGN = {
   RIGHT: ("right": "right"),
 };
 
-// PayPal mark variation options
-export const PAYPAL_MARK_VARIATIONS = {
+// Mark variation options (generic for future funding source support)
+export const MARK_VARIATIONS = {
   WORDMARK: ("wordmark": "wordmark"),
   MONOGRAM: ("monogram": "monogram"),
 };
--- a/src/constants/button.js
+++ b/src/constants/button.js
@@ -118,3 +118,9 @@ export const MESSAGE_ALIGN = {
   LEFT: ("left": "left"),
   RIGHT: ("right": "right"),
 };
+
+// Mark variation options for funding sources
+export const MARK_VARIATIONS = {
+  WORDMARK: ("wordmark": "wordmark"),
+  MONOGRAM: ("monogram": "monogram"),
+};
--- a/src/funding/paypal/monogramMark.jsx
+++ b/src/funding/paypal/monogramMark.jsx
@@ -1,10 +1,10 @@
 /* @flow */
 /** @jsx node */
 
-import { node, type ChildType } from "@krakenjs/jsx-pragmatic/src";
+import { node, type ComponentNode } from "@krakenjs/jsx-pragmatic/src";
 import { PPRebrandLogoExternalImage } from "@paypal/sdk-logos/src";
 import { LOGO_COLOR } from "@paypal/sdk-logos/src/constants";
 
-export function PayPalMonogramMark(): ChildType {
+export function PayPalMonogramMark(): ComponentNode<{| logoColor: string |}> {
   return <PPRebrandLogoExternalImage logoColor={LOGO_COLOR.BLUE} />;
 }
--- a/src/funding/paypal/monogramMark.jsx
+++ b/src/funding/paypal/monogramMark.jsx
@@ -1,0 +1,10 @@
+/* @flow */
+/** @jsx node */
+
+import { node, type ChildType } from "@krakenjs/jsx-pragmatic/src";
+import { PPRebrandLogoExternalImage } from "@paypal/sdk-logos/src";
+import { LOGO_COLOR } from "@paypal/sdk-logos/src/constants";
+
+export function PayPalMonogramMark(): ChildType {
+  return <PPRebrandLogoExternalImage logoColor={LOGO_COLOR.BLUE} />;
+}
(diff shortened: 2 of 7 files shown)


src/marks/component.jsx
```diff
@@ -52,6 +52,7 @@ type MarksInstance = {|
 
 type MarksProps = {|
   fundingSource?: ?$Values<typeof FUNDING>,
+  variationName?: string,
```
nikrom17: `variationName` is typed as `string` here, but only `"wordmark"` and `"monogram"` are valid values. The rest of the codebase uses `$Values<typeof CONSTANT>` for this pattern (e.g., `fundingSource` uses `$Values<typeof FUNDING>`). Using `string` means a typo like `variationName: "monogrm"` won't produce a Flow error.

```js
variationName?: ?$Values<typeof MARK_VARIATIONS>,
```

Same applies to the `MarkOptions` and `MarksElementOptions` types in `template.jsx` and `templateRebrand.jsx`.

src/marks/template.jsx
```diff
@@ -16,13 +16,15 @@ import { CLASS } from "../constants";
...
   fundingSource: $Values<typeof FUNDING>,
+  variationName?: string,
   fundingEligibility: FundingEligibilityType,
   experiment: Experiment,
   env: $Values<typeof ENV>,
 |};
 
 function Mark({
   fundingSource,
+  variationName, // eslint-disable-line no-unused-vars
```
nikrom17: This prop is destructured and immediately suppressed with `eslint-disable-line no-unused-vars`. Since the non-rebrand `Mark` doesn't use `variationName`, it would be cleaner to not pass it through this file at all. You could remove `variationName` from `Mark`, `MarksElement`, and both type definitions in this file. It's only needed in `templateRebrand.jsx`.

src/marks/templateRebrand.jsx
```diff
@@ -68,11 +72,36 @@ function Mark({
...
+
+  // Helper function to render monogram
+  const renderMonogram = () => (
+    <div class="paypal-mark-rebrand paypal-mark-rebrand-white">
+      <PayPalMonogramMark />
+    </div>
+  );
+
+  // Handle PayPal variations
+  if (fundingSource === FUNDING.PAYPAL) {
```
nikrom17: This branching has a subtle gap: `variationName === undefined` uses strict equality, so `null` (common in JSX conditional props like `variationName={condition ? "monogram" : null}`) skips both branches and falls through to the default `renderWordmark()`. The behavior happens to be correct, but it's accidental.

You could simplify the whole block by flipping the logic. Instead of checking for wordmark + undefined, just check for monogram and let everything else default to wordmark:

```jsx
if (fundingSource === FUNDING.PAYPAL && variationName === MARK_VARIATIONS.MONOGRAM) {
  return renderMonogram();
}

return renderWordmark();
```

This handles undefined, null, "wordmark", and any invalid value in one path, and there's no gap between branches.

src/constants/button.js
```diff
@@ -118,3 +118,9 @@ export const MESSAGE_ALIGN = {
   LEFT: ("left": "left"),
   RIGHT: ("right": "right"),
 };
+
+// Mark variation options for funding sources
+export const MARK_VARIATIONS = {
```
nikrom17, line 123: Since variations only apply to `FUNDING.PAYPAL`, the naming should make that clear. A few renames to consider:

- `MARK_VARIATIONS` → `PAYPAL_MARK_VARIATIONS`
- `variationName` prop → `paypalMarkVariation` (in `component.jsx`, `template.jsx`, `templateRebrand.jsx` types and destructuring)
- `renderMonogram` → `renderPayPalMonogram` (in `templateRebrand.jsx`)

This makes it obvious at the call site that the prop won't affect Venmo, Credit, or other funding sources.

src/funding/paypal/monogramMark.jsx
```diff
@@ -0,0 +1,10 @@
+/* @flow */
+/** @jsx node */
+
+import { node, type ChildType } from "@krakenjs/jsx-pragmatic/src";
+import { PPRebrandLogoExternalImage } from "@paypal/sdk-logos/src";
+import { LOGO_COLOR } from "@paypal/sdk-logos/src/constants";
+
+export function PayPalMonogramMark(): ChildType {
```
nikrom17: No tests were added for this new component or the variation branching logic in `templateRebrand.jsx`. A few scenarios worth covering:

- Monogram renders when `variationName: "monogram"` with `FUNDING.PAYPAL`
- Wordmark renders when `variationName` is undefined or `"wordmark"`
- Non-PayPal funding sources ignore `variationName` entirely
- Fallback to wordmark for unrecognized `variationName` values

nikrom17 (changes requested):

src/constants/button.js
```diff
@@ -118,3 +118,9 @@ export const MESSAGE_ALIGN = {
   LEFT: ("left": "left"),
   RIGHT: ("right": "right"),
 };
+
+// PayPal mark variation options
+export const PAYPAL_MARK_VARIATIONS = {
```
nikrom17: Sorry for the back and forth on naming. Venmo alternative mark wasn't included in the Asset Refinements page, so we're not adding it now. But let's keep the naming generic (`MARK_VARIATIONS`, `markVariation`) instead of PayPal-specific, in case we need to support variations for other funding sources later.

nikrom17 (changes requested):

src/marks/templateRebrand.jsx
nikrom17, line 192: The PayPal wordmark logo needs `padding-top: 2px` to align properly within the mark. Instead of applying it inline, add a CSS rule in the style block below (around the existing `.paypal-mark-rebrand .paypal-logo` rules):

```css
.paypal-mark-rebrand .paypal-logo-paypal-rebrand {
    padding-top: 2px; 
    margin-right: ${toPx(rebrandHeight / 5)};
}
```

This targets only the PayPal logo specifically (via `paypal-logo-paypal-rebrand` class) without affecting other funding source logos in the marks.

nikrom17 (changes requested):

nikrom17 (approved):

gregjopa (approved):

PR #2657, opened 2026-08-25
Resize Venmo QR modal for QR enhancement experiment

nityasp:
This PR introduces a smaller modal size (512×382) for buyers bucketed into the venmoDesktopQRCodeEnhancements experiment, while keeping the existing legacy size (720×612).

1. Adjusted `#qrModal` CSS dimensions for the new Venmo QR design.
2. Gated size behind the experiment - Added a new prop `isRedesign` which conditionally renders `#qrModal` at 512×382 (redesign) vs 720×612 (legacy). Added experiment to `QRCodeProps` type.
3. Added experiment as an explicit zoid prop definition with `allowDelegate: true`

Here is the related v5 [PR](https://github.com/OnePayPal/smartcomponentnodeweb/pull/1491) for this.


jtanya17 (approved):

tmanderson (approved):
Looks good. Pointed this out in the v5 PR, but wanted to call out [this work](https://github.com/OnePayPal/smartcomponentnodeweb/compare/develop...feature/DTPPCPSDK-6328-venmo-qr-code-modal-close-button#diff-69a4c898165a4ebd5771da5291578bcf9c2079730fd2c554b2965bbf44739e04R206) I'd done earlier to pass the experiment flags down to the iframe displays.

Might need to chat a bit more on that.

PR #2660, opened 2026-09-08
observe window name

imbrian:
### Description

Observes changes to `window.name` from within Checkout component when component is a child. Ignores changes from zoid components, as zoid will mutate this itself as part of the popup component lifecycle.

❤️ Thank you!

### DIFF
--- a/src/zoid/checkout/component.jsx
+++ b/src/zoid/checkout/component.jsx
@@ -25,6 +25,7 @@ import {
   noop,
   supportsPopups,
   inlineMemoize,
+  stringifyError,
 } from "@krakenjs/belter/src";
 import { FUNDING } from "@paypal/sdk-constants/src";
 import {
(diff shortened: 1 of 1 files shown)


ravishekhar (approved):

src/zoid/checkout/component.jsx
```diff
@@ -42,6 +43,50 @@ import { DEFAULT_POPUP_SIZE } from "./config";
...
+// See buildChildWindowName in @krakenjs/zoid.
+const ZOID_WINDOW_NAME_PATTERN = /^__zoid__.+__$/;
+
+function spyOnWindowNameAssignment(): void {
+  try {
+    let currentWindowName = window.name;
+
+    Object.defineProperty(window, "name", {
+      configurable: true,
+      get(): string {
```
dtjones404, line 57: Technically I don't think we're doing anything to guarantee that the new value is a string.  If not I think we'll throw later on when we call currentWindowName.slice

imbrian: nice catch, fixed

dtjones404: LGTM.  Nit: should we use 'unknown' for the type annotation instead of 'string'?

dtjones404 (approved):

